PR #10, opened 2017-07-18
chapters/appendix-I: Bump license list to v2.6 (via new script)

### LINKED ISSUE #6: Update SPDX License List in Appendix 1
The following table contains the full names and short identifiers for the SPDX License List, v2.5 which was released July 2016.  For the full and most up-to-date version of the SPDX License List as well as other related information, please see http://spdx.org/licenses/

Should we upgrade to v2.6 in the next fix release

wking:
Add `pull-license-list.py` to automatically pull the current license and exception JSON into Markdown tables.  Using JavaScript might have been more idiomatic than the Python script, but it's faster for me to write the Python; someone who is more familiar with JavaScript can translate the script and incorperate it into the build tooling later if we decide against carrying the auto-generated tables in this repository.

Two things made the auto-generation more difficult than it needed to be:

* Newlines in exception names (spdx/license-list-data#10).
* [edit: addressed [below][1]] No charset declaration in JSON served from spdx.org:

        $ curl -sI https://spdx.org/licenses/licenses.json | grep -i content-type
        Content-Type: application/json

    it should have been `Content-Type: application/json; charset=utf-8`.  I don't know if there is a repository with the Nginx config used for spdx.org, so I haven't filed an issue for this last one.

Fixes #6.

[1]: https://github.com/spdx/spdx-spec/pull/10#issuecomment-343275781

### DIFF
--- a/bin/pull-license-list.py
+++ b/bin/pull-license-list.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+#
+# Automatically update the license- and exception-list Markdown based
+# on the currently-live JSON.
+#
+# usage: ./bin/pull-license.py
+#
+# Script licensed under SPDX-License-Identifier: MIT
+
+import codecs
+import itertools
+import json
+import os.path
+import re
+import urllib.request
+
+
+VERSION_REGEXP = re.compile(
+    pattern='(.* SPDX License List, v)([^ ]*)( which was released )([^.]*)(\..*)',
+    flags=re.DOTALL)
+
+
+def get_json(url):
+    with urllib.request.urlopen(url=url) as response:
+        charset = response.headers.get_content_charset('UTF-8')
+        reader = codecs.getreader(charset)
+        return json.load(reader(response))
+
+
+def format_table(headers, rows):
+    widths = [
+        max(len(row[i]) for row in rows + [headers])
+        for i, _ in enumerate(headers)
+    ]
+    template = '| {} |\n'.format(' | '.join(
+        ('{' + '{}:{{{}}}'.format(2*i, 2*i + 1) + '}')
+        for i, _ in enumerate(headers)
+    ))
+    yield template.format(*itertools.chain(*zip(headers, widths)))
+    yield '|{}|\n'.format('|'.join('-' * (width + 2) for width in widths))
+    for row in rows:
+        yield template.format(*itertools.chain(*zip(row, widths)))
+    yield '\n'
+
+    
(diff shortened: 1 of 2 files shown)


goneall:
I don't know of any nginx configs available, but I'll went ahead and filed a ticket with the Linux foundation helpdesk - request ID #43220.

emsearcy:
Unless I'm misunderstanding something, [this SO article](https://stackoverflow.com/a/26206930) makes the case that adding a charset to application/json is pointless because all RFC-compliant clients must ignore the charset on this particular type.

Is it *only* /licenses/licenses.json that you are requesting this be added for?  Doing a quick check on the server shows many ASCII .json files under licenses, and I'm not aware of an nginx or httpd option to dynamically pick the correct charset.

```
# find licenses -name '*.json' | xargs file | grep UTF-8 | wc -l
297
# find licenses -name '*.json' | xargs file | grep ASCII | wc -l
1460
```

goneall:
> Is it only /licenses/licenses.json that you are requesting this be added for? Doing a quick check on the server shows many ASCII .json files under licenses, and I'm not aware of an nginx or httpd option to dynamically pick the correct charset.

All .json files in the /licenses directory.

All of the JSON files should be encoded using UTF-8.  I believe the ASCII text found was actually the word "ASCII" in the data portion of the files.

@wking Based on the article SO article referenced above, perhaps we do not need the charset declaration and can just assume it is UTF-8.

BTW - if you find any encoding issues, it is likely due to the source text in the license-list or license-list-XML repositories being incorrect.   Everything should be UTF-8, but I have found several cases where the source license text was not in fact UTF-8.

wking:
On Thu, Nov 09, 2017 at 06:10:47PM +0000, Eric Searcy wrote:
> Unless I'm misunderstanding something, [this SO
> article](https://stackoverflow.com/a/26206930) makes the case that
> adding a charset to application/json is pointless because all
> RFC-compliant clients must ignore the charset on this particular
> type.

Fair.  It looks like Python has added direct-byte-handling for this in
3.6 [1,2], so I'll change my mind and say it's appropriate to continue
to not set charset for application/json.

[1]: https://docs.python.org/3/whatsnew/3.6.html#json
[2]: https://bugs.python.org/issue17909

wking:
Updated to use Python 3.6+ with 1e38bd9 → 79e2162.

iamwillbar:
This PR is superseded by #151.

Issue #16, opened 2017-08-01
Remove + from valid idstring values and consistently link to appendix IV


wking:
The 2.1 spec is not particularly [DRY][1] on `idstring` values.  There are a number of [local][2] [definitions][3] that match up with `1*(ALPHA / DIGIT / "-" / ".")`, a [definition that includes `+`][4] (perhaps from before it was a [Licence Expression operator][5]?) and a “[defined in Appendix][6]” (without specifying which appendix).  I think we should extend our use of ABNF to include more than just appendix IV.  We'd define `idstring` (or just `id`?) in the first place we needed it ([here][4]?), and then later sections would link that earlier definition and consume it's ABNF rule.

I'm happy to work up a PR for this if it sounds useful.

[1]: https://en.wikipedia.org/wiki/Don't_repeat_yourself
[2]: https://spdx.org/spdx-specification-21-web-version#h.2bn6wsx
[3]: https://spdx.org/spdx-specification-21-web-version#h.32hioqz
[4]: https://spdx.org/spdx-specification-21-web-version#h.h430e9ypa0j9
[5]: https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60
[6]: https://spdx.org/spdx-specification-21-web-version#h.7i829o8ausxy

kestewart:
Talked on call and agreed a good target for 2.2.  Pull request welcomed.

wking:
> Pull request welcomed.

I started looking into this, but decided that I'd rather follow [the XML spec][1] and have an early section for these common constructs.  However, if I add a chapter 2 for common constructs, I'll have to renumber the following chapters, update their section numbers, and update any internal references.  I'm really missing LaTeX or some other system with auto-numbered sections and references ;).  I'll grind through the renumbering, but I'd like to wait until #8 lands first.

[1]: https://www.w3.org/TR/REC-xml/#sec-common-syn

PR #17, opened 2017-08-01
chapters/3: Lots of Markdown and copy edits (e.g. quotes -> backticks)

wking:
I initially went in to clean up some confused curly quotes (e.g. XML examples that used `”` where they clearly intended `"`.  But there were enough Markdown issues that I ended up taking a pass through the whole file fixing everything I noticed.  I've also tried to:

* Add backticks around literals, instead of using curly quotes or not marking them up at all.
* Add cross-links where appropriate, since that makes it easier to dig into terms you don't recognize if you start in the middle of the spec.
* Add links to external projects, for folks who are unfamiliar with the names we're dropping.
* Add links to RFCs, where we discuss concepts that are documented in RFCs.
* Add “in class” entries to take everything back to `spdx:Package`, so the nesting is explict.  This lead to the removal of `packageVerificationCode` in §3.9.7 and `Checksum` in §3.10.7.
* Remove redundant wording (“Here, ”, “This field”, etc., as well as longer phrases that didn't seem to be adding anything).
* Capitalize the first word after “Data Format:”, for consistency with the other entries.
* Standardize around “Data Format:” (the old text was split between that and “DataFormat:”).
* Standardize around `…` for elided text.

I imagine I missed a few things, and I'd also like to make some larger cleanups (e.g. using ABNF to formally define compound constructs, #16), but this commit seemed like a good start.

### DIFF
--- a/chapters/3-package-information.md
+++ b/chapters/3-package-information.md
@@ -4,10 +4,10 @@ One instance of the Package Information is required per package being described.
 
 Cardinality: Optional, one or many.
 
-In tag:value format, the order in which package and files occur is syntactically significant.
+In `tag:value` format, the order in which package and files occur is syntactically significant.
 
-A new Package Information section is denoted by the Package Name field.
-All Package Information fields must be grouped together before the start of a Files section, if file(s) are present.
+A new Package Information section is denoted by the [Package Name](#3.1) field.
+All Package Information fields must be grouped together before the start of a [Files section](4-file-information.md), if file(s) are present.
 All files contained in a package must immediately follow the applicable Package Information.
 A new Package Information section (via Package Name) denotes the start of another package.
 Sub-packages should not be nested inside a Package Information section, but should be separate and should use a Relationship to clarify.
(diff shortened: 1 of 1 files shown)


wking:
I've pushed c89cef2 → 1d75963 with updates from the 2018-01-23 meeting:

* punt on `verificationcode` refactoring, although I am adding an explicit `read(…)` function so `SHA1(…)` can always operate on bytes.
* Drop the `packageVerificationCode` / `PackageVerificationCode`, etc. unifications, since @goneall says they're intentional (with the RDF subject lowercase and predicate uppercase).
* Fix some closing backtick and Markdown link typos in §3.14.4 and 3.15.4.
* Stick with `<…>` for substitutions.
* Punt on words with spaces in substitutions
* Use paragraphs and list-ish entries for `NOASSERTION` and friends.

Besides those meeting changes, I've also added a few more missing cross-links, including fixing a “defined in Appendix.” typo.

wking:
I've pushed 1d75963 → 79f6bcc, which makes a few changes requested by @kestewart:

* Punts every line touch in §3.9.4 and §3.9.5 (PackageVerificationCode), even curly → straight quote changes.
* Rewords the `<type>` line in §3.21.4 to reference the enumerated types in Appendix VI. I've removed the `idstring` reference from §3.21.4, so there's no longer a need to reference [the `idstring` ABNF in Appendix IV][1].  And I will be very happy if we land the Appendix VI removal in-flight with #54, although that's currently milestoned 3.0.

[1]: https://github.com/spdx/spdx-spec/blame/df26160783956ed87d78889dd1d4a53022e713eb/chapters/appendix-IV-SPDX-license-expressions.md#L11

Issue #22, opened 2017-08-29
Permit comments to be added to SPDX documents


kestewart:
Comments in SPDX documents (depends on file formats).
In RDFa/XML - there is a specific term defined.
In tag:value -  # - at start of line - need to be added to document.  Entire line comments.
No middle of comment line in SPDX document.

Comments take the form of '#', as the first non-blank character, and continue to the end of line (marked by characters U+000D or U+000A) or end of file if there is no end of line after the comment marker. Comments are treated as white space.  
    ⇒  Alexios recommends aligning with Turtle

What about SPDX-License-Identifier: (  ) in source code.  

"#"  ok if disallowed character?   Need to check specification.

wking:
On Tue, Aug 29, 2017 at 05:00:40PM +0000, Kate Stewart wrote:
> In RDFa/XML…

RDFa and RDF/XML seem to be two separate things [1].

> … - there is a specific term defined.

Do you mean rdfs:comment and such (e.g. [2]).  Or CreatorComment [3]?

> In tag:value - # - at start of line - need to be added to document.

Things like CreatorComment are already specified in both our
XML-serialized RDF and our tag:value format (see §2.10.5 and 2.10.6
[3]).  Are you looking for a higher-level comment analogous to XML's
 [4]?  I expect that's the best equivalent for a # line in a
tag:value file.

[1]: https://www.w3.org/wiki/RDFa_vs_RDFXML
[2]: https://github.com/spdx/spdx-spec/blame/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/2-document-creation-information.md#L157
[3]: https://github.com/spdx/spdx-spec/blob/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/2-document-creation-information.md#2.10
[4]: https://www.w3.org/TR/REC-xml/#sec-comments

goneall:
> Do you mean rdfs:comment and such (e.g. [2]).  Or CreatorComment [3]?
rdfs:comment

kestewart:
@wking - it was looking for the equivalent of higher-level comment to XML's.   # seems like the best option.

wking:
Do we actually define the tag/value syntax anywhere?  This sounds like
another thing that could go in a common constructs chapter [1].

[1]: https://github.com/spdx/spdx-spec/issues/16#issuecomment-325764699

kestewart:
[...]

PR #37, opened 2017-09-13
appendix-IV: Refactor the license expression appendix

wking:
This is a large diff, but I aimed for restructuring/polishing without changing the end result.  I did make a few minor, intentional changes:

* Extended `license-id` to include appendix I.3 (deprecated licenses).  We don't want folks using these in license expressions (because they're deprecated), but they are valid (or we would have removed them instead of just deprecating them).  That means that in some cases the nature of a string is unclear.  For example `GPL-1.0+` could be the depreacted `license-id`, or it could be a `simple-expression` using the non-deprecated `GPL-1.0` `license-id` and the `+` operator.  I don't think that's a problem though, because I can't think of a case where the ambiguity would matter.

* I've allowed `+` for `license-ref` (it used to be only for `license-id`).  There could be external licenses which offer a choice between only-this-version and or-later grants, and allowing `+` for `license-ref` makes it easier to support those licenses as they transition into the SPDX License List.  This isn't a big deal, but it avoids needing separate `license-ref`s for the only-this-version and or-later grants if you need both.

* I've added explicit whitespace handling, vs. the previous version which just discussed it in the text.  That way the ABNF is the sole source of normative syntax information.

* I've added `enclosed-license-expression`, so consumers like the tag:value format can suggest/require it.  This allows for more precision in consumers (e.g. appendix V should be updated to require `enclosed-license-expression`), but I've left those other sections alone for this commit.  Ideally the tag:value line would be moved to a separate section that defined the tag-value format, but [we don't have such a section yet][1].

Also, it seems odd that there's no way to define an external license exception (`LicenseExceptionRef`?).  But I haven't touched that in this commit.

The RDF/XML bindings seem to have some holes around the `+` and `WITH` operators.  I've left some FIXMEs where I think we need adjustments.

[1]: https://github.com/spdx/spdx-spec/issues/22#issuecomment-327253273

### DIFF
--- a/chapters/appendix-IV-SPDX-license-expressions.md
+++ b/chapters/appendix-IV-SPDX-license-expressions.md
@@ -1,142 +1,159 @@
-# SPDX License Expressions
+# Appendix IV: SPDX License Expressions
 
-## Overview
+## IV.1 Overview <a name="IV.1"></a>
 
-Often a single license can be used to represent the licensing terms of a source code or binary file, but there are situations where a single license identifier is not sufficient. A common example is when software is offered under a choice of one or more licenses (e.g., GPL-2.0 OR BSD-3-Clause). Another example is when a set of licenses is needed to represent a binary program constructed by compiling and linking two (or more) different source files each governed by different licenses (e.g., LGPL-2.1 AND BSD-3-Clause).
+Often a single license can be used to represent the licensing terms of a source code or binary file, but there are situations where a single license identifier is not sufficient. A common example is when software is offered under a choice of one or more licenses (e.g., `GPL-2.0 OR BSD-3-Clause`). Another example is when a set of licenses is needed to represent a binary program constructed by compiling and linking two (or more) different source files each governed by different licenses (e.g., `LGPL-2.1 AND BSD-3-Clause`).
 
-SPDX License Expressions provides a way for one to construct expressions that more accurately represent the licensing terms typically found in open source software source code. A license expression could be a single license identifier found on the SPDX License List; a user defined license reference denoted by the LicenseRef-`[idString]`; a license identifier combined with an SPDX exception; or some combination of license identifiers, license references and exceptions constructed using a small set of defined operators (e.g., `AND`, `OR`, `WITH` and `+`). We provide the definition of what constitutes a valid an SPDX License Expression in this section.
+SPDX License Expressions allow accurate representations of the licensing terms typically found in open source software. A license expression could be a single license identifier found on the [SPDX License List](appendix-I-SPDX-license-list.md); a user defined license reference denoted by the `LicenseRef-[idString]`; a license identifier combined with an SPDX exception; or some combination of license identifiers, license references and exceptions constructed using a small set of defined operators (`AND`, `OR`, `WITH` and `+`).  We provide the definition of what constitutes a valid an SPDX License Expression in this section.
 
-The exact syntax of license expressions is described below in [ABNF](http://tools.ietf.org/html/rfc5234).
+## IV.2 Syntax <a name="IV.2"></a>
 
-idstring = 1*(ALPHA / DIGIT / "-" / "." )
+The exact syntax of license expressions is described below in [ABNF][rfc5234].  `ALPHA`, `DIGIT`, and `WSP` are from the [ABNF core rules][rfc5234-aB].
 
-license-id = \<short form license identifier in [Appendix I.1](./appendix-I-SPDX-license-list.md)>
+```
+idstring                       = 1*(ALPHA / DIGIT / "-" / "." )
+license-id                     = <short form license identifier>
+license-exception-id           = <short form license exception identifier>
+license-ref                    = ["DocumentRef-" idstring ":"] "LicenseRef-" idstring
+license                        = license-id / license-ref
+space                          = 1*WSP
+simple-expression              = license ["+"] [space "WITH" space license-exception-id]
+binary-expression-operator     = "AND" / "OR"
+license-expression             = simple-expression
+                               / license-expression space binary-expression-operator space license-expression
+                               / "(" license-expression ")"
+                               / space license-expression
+                               / license-expression space
+enclosed-license-expression    = [space] simple-expression [space]
+                               / [space] "(" [space] license-expression [space] ")" [space]
+```
 
-license-exception-id = \<short form license exception identifier in [Appendix I.2](appendix-II-license-matching-guidelines-and-templates.md)>
+Where:
 
-license-ref = ["DocumentRef-"1\*(idstring)":"]"LicenseRef-"1*(idstring)
+* Short form license identifiers are drawn from sections [I.1](appendix-I-SPDX-license-list.md#I.1) and [I.3](appendix-I-SPDX-license-list.md#I.3).
+* Short form license exception identifiers are drawn from [section I.2](appendix-I-SPDX-license-list.md#I.2).
 
-simple-expression = license-id / license-id"+" / license-ref
+License expressions are case sensitive, and license expressions must use the canonical case for all components.
+However, the SPDX Licence List commits to [never contain identifiers which differ only in case but have different semantics][list-case-commitment].
+Similarly, SPDX License Expressions will never contain operators or other local components which differ only in case but have different semantics.
+Tools are encouraged, but not required, to take advantage of those commitments and support parsing license expressions which alter the canonical casing of any expression components.
 
-compound-expression = 1*1(simple-expression /
+When a particular rule is not referenced, “SPDX License Expressions” means the more general `license-expression`.
+Consumers that need the more limited `enclosed-license-expression` (which may be easier to parse depending on the surrounding context) SHOULD say so explicitly.
 
+Tag:value authors SHOULD use `enclosed-license-expression` instead of the more general `license-expression`.
 
-simple-expression "WITH" license-exception-id /
+### IV.2.1 The `+` Operator <a name="IV.2.1"></a>
 
-  compound-expression "AND" compound-expression /
+The unary `+` suffix operator represents the given version of the license or any later version.
 
-  compound-expression "OR" compound-expression ) /
+For example, a grant like:
 
-  "(" compound-expression ")" )
+> This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 2 of the License, or (at your option) any later version.
 
-license-expression = 1*1(simple-expression / compound-expression)
+could be be represented as `GPL-2.0+`, although that form has been deprecated in favor of `GPL-2.0-or-later`.
 
-In the following sections we describe in more detail `<license-expression>` construct, a licensing expression string that enables a more accurate representation of the licensing terms of modern day software.
+The set of alternatives may depend on the specific license text.  For example, `AGPL-1.0+` (and its successor `AGPL-1.0-or-later`) invokes the “any later version” condition from §9 of the [`AGPL-1.0`][AGPL-1.0], allowing you to redistribute and/or modify it under the terms of the [AGPL-2.0][] (a later version published by Affero) or under the terms of the [`GPL-3.0-or-later`][GPL-3.0-or-later] (specifically allowed via the AGPL-1.0's §9).
 
-A valid `<license-expression>` string consists of either:
+### IV.2.2 Exception `WITH` Operator <a name="IV.2.2"></a>
 
-(i) a simple license expression, such as a single license identifier; or
+Sometimes a set of license terms apply except under special circumstances.  In this case, use the binary `WITH` operator to construct a new simple expression to represent the exception situation.  A valid `simple-expression` is where the left operand is a `license` value (with an optional `+` suffix) and the right operand is a `license-exception-id` that represents the exception terms.
 
-(ii) a more complex expression constructed by combining smaller valid expressions using Boolean license operators.
+For example, when the Bison exception is applied to `GPL-2.0-or-later`, the `simple-expression` would be:
 
-There MUST NOT be whitespace between a license-id and any following `+`. This supports easy parsing and backwards compatibility. There MUST be whitespace on either side of the operator "WITH". There MUST be whitespace and/or parentheses on either side of the operators `AND` and `OR`.
+    GPL-2.0-or-later WITH Bison-exception-2.2
 
-## Simple License Expressions <a name="simple-expr"></a>
+The current set of valid exceptions can be found in [Appendix I, section 2](appendix-I-SPDX-license-list.md#I.2).  For the most up to date set of exceptions please see [spdx.org/licenses](https://spdx.org/licenses).  If the applicable exception is not found on the SPDX License Exception List, then use a single `license-ref` to represent the entire license terms (including the exception).
 
-A simple `<license-expression>` is composed one of the following:
+### IV.2.3 Conjunctive `AND` Operator <a name="IV.2.3"></a>
 
-* An SPDX License List Short Form Identifier. For example: GPL-2.0
-* An SPDX License List Short Form Identifier with a unary"+" operator suffix to represent the current version of the license or any later version. For example: GPL-2.0+
-* A SPDX user defined license reference: ["DocumentRef-"1\*(idstring)":"]"LicenseRef-"1*(idstring)
+If required to simultaneously comply with two or more licenses, use the conjunctive binary `AND` operator to construct a new license expression, where both the left and right operands are valid `license-expression` values.
 
-Some examples:
+For example, when one is required to comply with both the `LGPL-2.1` or `MIT` licenses, a valid `license-expression` would be:
 
-    LicenseRef-23
-
-    LicenseRef-MIT-Style-1
-
-    DocumentRef-spdx-tool-1.2:LicenseRef-MIT-Style-2
-
-## Composite License Expressions <a name="composite-expr"></a>
-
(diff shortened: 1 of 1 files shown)


wking:
Cross linking:

* An old [spdx Bugzilla issue][1] with ABNF discussion.
* jslicense/spdx-expression-parse.js#20 with ABNF discussion.

[1]: https://bugs.linuxfoundation.org/show_bug.cgi?id=1334

sschuberth:
FYI, I already had implemented `LicenseExpressionParser: Allow "+" as part of license-refs` as part of https://github.com/spdx/tools/pull/66, but back then we couldn't reach a consensus about that on the mailing list IIRC, so I closed my PR unmerged.

wking:
Some previous discussion in [1,2,3,4].  [4] looks like #16.  But I
haven't gone through those threads carefully enough to post a summary
yet.

[1]: https://lists.spdx.org/pipermail/spdx-tech/2015-October/002855.html
     Subject: Is "+" a valid character of a LicenseRef idstring?
     Date: Wed Oct 28 09:28:20 UTC 2015
[2]: https://lists.spdx.org/pipermail/spdx-tech/2015-November/002874.html
     Subject: Is "+" a valid character of a LicenseRef idstring?
     Date: Mon Nov 2 09:56:47 UTC 2015
[3]: https://lists.spdx.org/pipermail/spdx-tech/2015-December/002966.html
     Subject: [Bug 1333] New: SPDX SPEC fails to discuss whitespace
     Date: Tue Dec 8 16:54:54 UTC 2015
[4]: https://lists.spdx.org/pipermail/spdx-tech/2015-December/002986.html
     Subject: [Bug 1326] Inconsistency whether "+" is a valid
       character in LicenseRef ID strings
     Date: Tue Dec 22 18:17:39 UTC 2015

sschuberth:
@goneall, I also couldn't find the email thread again, but AFAIR my PR would need to take the SPDX version into account, which would have made the already quite ugly code even more ugly, so I decided to not further work on this.

goneall:
After reading through the history, it looks like we decided to deprecate the + in the license ID's which would allow for it to be used in license-ref's for version 2.1 or later.

I'll think about the code a bit more - I agree with @sschuberth that it would be quite messy to check for the version.  That being said, we should update the tools to handle this case.

chapters/appendix-IV-SPDX-license-expressions.md
```diff
@@ -1,175 +1,174 @@
...
-represents a license choice between either LGPL-2.1 and the expression BSD-3-Clause AND MIT because the AND operator takes precedence over (is applied before) the OR operator.
+represents a license choice between either `LGPL-2.1` and the expression `BSD-3-Clause AND MIT` because the `AND` operator takes precedence over (is applied before) the `OR` operator.
 
-When required to express an order of precedence that is different from the default order a \<license-expression> can be encapsulated in pairs of parentheses: ( ), to indicate that the operators found inside the parentheses takes precedence over operators outside. This is also similar to the use of parentheses in an algebraic expression e.g., (5+7)/2.
+## IV.3 License Expressions in XML <a name="IV.3"></a>
 
-For instance, the following expression:
+### IV.3.1 The `+` Operator <a name="IV.3.1"></a>
+
+FIXME: How are you supposed to cover this in XML?
```
goneall: The following text can be added:
The `+` operator can be expressed in XML via a `<spdx:OrLaterOperator>` element, with a single `spdx:member` property for the license preceeding the `+` operator.

     <spdx:OrLaterOperator>
          <spdx:member rdf:resource="http://spdx.org/licenses/GPL-2.0"/>
     </spdx:OrLaterOperator>

chapters/appendix-IV-SPDX-license-expressions.md
```diff
@@ -1,175 +1,174 @@
...
 
-### 5) License Expressions in RDF <a name="rdf-expr"></a>
+* [`rdfs:comment`][rdfs:comment] element describing the nature of the exception.
+* [`rdfs:seeAlso`][rdfs:seeAlso] element (optional) referencing external sources of information on the exception.
+* `spdx:example`: Text describing examples of this exception.
+* `spdx:name`: The full human readable name of the item.
+* `spdx:licenseExceptionId`: The identifier of an exception in the SPDX License List to which the exception applies.
 
-A conjunctive license can be expressed in RDF via a `<spdx:ConjunctiveLicenseSet>` element, with an spdx:member property for each element in the conjunctive license. Two or more members are required.
+    FIXME: I don't think this element makes sense.  What you want is a way to represent a `simple-expression` based on `WITH`.  This subsection seems to be focused on defining exceptions, but it does not cover *consuming* those license exceptions.  And “and exception … to which the exception applies” is really not clear; is it the left or right `WITH` operand?  What happens for licenses/exceptions that are not in the SPDX License List?
```
goneall: There is another SPDX class that needs to be described.  The With operator:

The WITH operator can be expressed in XML via a `<spdx:WithExceptionOperator>` element, with an `spdx:member` property for the license and an `spdx:licenseException` property for the exception.

     <spdx:WithExceptionOperator>
          <spdx:member rdf:resource="http://spdx.org/licenses/GPL-2.0"/>
          <spdx:licenseException rdf:resource="https://spdx.org/licenses/FLTK-exception">
      </spdx:WithExceptionOperator>

chapters/appendix-IV-SPDX-license-expressions.md
```diff
@@ -1,175 +1,174 @@
...
+idstring                    = 1*(ALPHA / DIGIT / "-" / "." )
+license-id                  = <short form license identifier>
+license-exception-id        = <short form license exception identifier>
+license-ref                 = ["DocumentRef-" idstring ":"] "LicenseRef-" idstring
+license                     = license-id / license-ref
+space                       = 1*WSP
+simple-expression           = license ["+"] [space "WITH" space license-exception-id]
+binary-compound-operator    = "AND" / "OR"
+license-expression          = simple-expression
+                            / license-expression space binary-license-operator space license-expression
```
goneall: binary-license-operator -> binary-compound-operator

chapters/appendix-IV-SPDX-license-expressions.md
```diff
@@ -1,175 +1,174 @@
...
-license-ref = ["DocumentRef-"1\*(idstring)":"]"LicenseRef-"1*(idstring)
+* Short form license identifiers are drawn from sections [I.1](appendix-I-SPDX-license-list.md#I.1) and [I.3](appendix-I-SPDX-license-list.md#I.3).
+* Short form license exception identifiers are drawn from [section I.2](appendix-I-SPDX-license-list.md#I.2).
 
-simple-expression = license-id / license-id”+” / license-ref
+When a particular rule is not referenced, “SPDX License Expressions” means the more general `license-expression`.
+Consumers that need the more limited `enclosed-license-expression` (which may be easier to parse depending on the surrounding context) SHOULD say so explicitly.
 
-compound-expression = 1*1(simple-expression /
+Tag:value authors SHOULD use `enclosed-license-expression` instead of the more general `license-expression`.
```
goneall: SHOULD -> MUST - this is a rule we agreed to to enable machine parsing

chapters/appendix-IV-SPDX-license-expressions.md
```diff
@@ -1,175 +1,174 @@
...
+        <rdf:type rdf:resource="http://spdx.org/rdf/terms#LicenseException"/>
+    </rdf:Description>
+
+### IV.3.3 Conjunctive `AND` Operator <a name="IV.3.3"></a>
+
+A [conjunctive license](#IV.2.3) can be expressed in XML via a `<spdx:ConjunctiveLicenseSet>` element, with an `spdx:member` property for each element in the conjunctive license.  Two or more members are required.
 
     <spdx:ConjunctiveLicenseSet>
         <spdx:member rdf:resource="http://spdx.org/licenses/GPL-2.0"/>
         <spdx:ExtractedLicensingInfo rdf:about="http://example.org#LicenseRef-EternalSurrender">
```
goneall: I think this is a missing <spdx:member> which should enclose the ExtractedLicensingInfo

chapters/appendix-IV-SPDX-license-expressions.md
wking: > binary-license-operator -> binary-compound-operator

I'll go with `binary-expression-operator` because the arguments are `license-expression`s.

chapters/appendix-IV-SPDX-license-expressions.md
wking: > SHOULD -> MUST - this is a rule we agreed to to enable machine parsing

But 2.1 [uses][1]:

> For the Tag:value format, any license expression that consists of more than one license identifier and/or LicenseRef, should be encapsulated by parentheses

I think making this stricter without a major version bump is a bit dicey, especially with [wild exceptions that don't set those enclosing parens][2].

On the other hand, [the tag:value format is currently poorly spec'ed][3], so I guess we can always tell broken docs and tools “sorry, but you should have known you were making assumptions about the eventual format once it was spec'ed out”.

[1]: https://github.com/spdx/spdx-spec/blame/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/appendix-IV-SPDX-license-expressions.md#L62
[2]: https://github.com/spdx/spdx-spec/issues/45#issuecomment-336303074
[3]: https://github.com/spdx/spdx-spec/issues/22#issuecomment-327253273

goneall: @wking Good point - I would like to "correct" the 2.1 spec.  

This was highly debated at the time, but I believe there was a consensus to require the parenthesis.  I do know that my parser would break without the parenthesis and I believe the syntax is ambiguous making it unreliable for any implementation of a parser for tag/value  @kestewart - let me know if you agree we can change this to "must".

wking: > I do know that my parser would break without the parenthesis and I believe the syntax is ambiguous making it unreliable for any implementation of a parser for tag/value…

Parsers can be patched, but the data they're parsing may be historical (like the wild example I link above).  I'm interested in having *new* parsers be able to parse as many *existing* declarations as is reasonably possible.  I'm less worried about breaking existing parsers.

chapters/appendix-IV-SPDX-license-expressions.md
wking: Addressed in cf8decf → 9c8d189.

chapters/appendix-IV-SPDX-license-expressions.md
wking: Addressed in cf8decf → 9c8d189.

chapters/appendix-IV-SPDX-license-expressions.md
wking: Addressed in cf8decf → 9c8d189.

chapters/appendix-IV-SPDX-license-expressions.md
wking: > The WITH operator can be expressed in XML via a `<spdx:WithExceptionOperator>` element…

That makes sense to me.  But I expect we want to drop the `<spdx:LicenseException>` element, because as far as I can tell we provide no other way for authors to define exceptions.  They can only use exceptions from [our list][1], or they can [define a `LicenseID-*` that includes both the license and the exception][2].  But I see nowhere outside of this appendix that suggests they can define their own exceptions.

[1]: https://github.com/spdx/spdx-spec/blob/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/appendix-I-SPDX-license-list.md#i2-exceptions-list-
[2]: https://github.com/spdx/spdx-spec/blob/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/6-other-licensing-information-detected.md#61-license-identifier-

wking:
I've pushed cf8decf → 9c8d189 addressing some of @goneall's points and including a few more copy-edits.  Outstanding issues are:

* Whether we want to keep `<spdx:LicenseException>` now that we're adding `<spdx:WithExceptionOperator>` ([thread][1]).

* Whether we want SHOULD or MUST for tag:values using `enclosed-license-expression` ([thread][2]).

[1]: https://github.com/spdx/spdx-spec/pull/37#discussion_r145836880
[2]: https://github.com/spdx/spdx-spec/pull/37#discussion_r145838125

chapters/appendix-IV-SPDX-license-expressions.md
goneall: @wking This was discussed and resolved during 2.1 development that tag/value files need parenthesis if the expression is complex.  I consider it a defect in the spec that this wasn't worded as a must.  The reference to parsers is just pointing out the implications of not documenting what was agreed to.

Since you disagree, why don't you keep it as should and I'll create my own pull request after yours is complete to change it to must.  That way it will not appear that you are requesting a change that you disagree with.  We can then pick up the debate on the new pull request.

chapters/appendix-IV-SPDX-license-expressions.md
goneall: The reference for the ```<spdx:LicenseException>``` is still of value for those wanting to parse the details out of the exception.  We may want to move this to a different section, however, so we don't confuse the reader as to whether they can define license exceptions.  Feel free to remove it for now and we can add it back in when we document the other fields.

wking:
I've pushed 9c8d189 → 922031a dropping `<spdx:LicenseException>` and adding `<spdx:WithExceptionOperator>` now that @goneall [gave that a green light][1].  We may come back to external extention definitions in follow-up work.

We've also [agreed to punt][2] the tag:value `enclosed-license-expression` SHOULD/MUST to follow-up work.

I'm not aware of any outstanding change requests for this PR.

[1]: https://github.com/spdx/spdx-spec/pull/37#discussion_r146060446
[2]: https://github.com/spdx/spdx-spec/pull/37#discussion_r146059763

goneall:
@wking thanks for all the updates.  I like the current state of the files.

@kestewart if you get a chance - could you review as well?  I would like to merge this in so that I can re-raise the must/should parenthesis discussions.

pombredanne:
> I've allowed + for license-ref (it used to be only for license-id). There could be external licenses which offer a choice between only-this-version and or-later grants, and allowing + for license-ref makes it easier to support those licenses as they transition into the SPDX License List. This isn't a big deal, but it avoids needing separate license-refs for the only-this-version and or-later grants if you need both.

This does not make sense to me at all. A ref is a ref. This is not an ID. This is easy to implement but just has no clear basis. And it opens a can of worms that make things complicated otherwise. Should you have one text for each of a + and base ref version or only one?

chapters/appendix-IV-SPDX-license-expressions.md
pombredanne: IMHO In most cases parens are not needed and just make the expression heavier and ugly looking for no good reason. I think that rather than mandating them, we should relax their use in upcoming spec versions. 

Also in the common case it never makes sense to have multi-line expressions .... so we should IMHO state that expression MUST be a single line. In doing so you make things cleaner and simpler.

wking:
On Tue, Nov 14, 2017 at 08:29:58AM +0000, Philippe Ombredanne wrote:
> > I've allowed + for license-ref (it used to be only for license-id)…
>
> This does not make sense to me at all. A ref is a ref. This is not
> an ID.

I see very little space between license identifiers defined on the
license list (e.g. ‘GPL-3.0’) and a license identifier defined by a
third party (e.g. ‘LicenseRef-MIT-Style-1’) besides whether you have
commit rights to the SPDX license list.  In both cases, there is a
license, and somebody is giving it a handful of characters as an
identifier.  I don't see a reason to support operators for identifiers
coined by the SPDX but then to not support those same operators for
identifiers coined by someone else.

> And it opens a can of worms that make things complicated
> otherwise. Should you have one text for each of a + and base ref
> version or only one?

Only one.  This is somewhat up in the air with [1]'s hard-coded
approach which removes the + operator entirely (or maybe just in some
cases?).  But regardless of how that shakes out, the motivation here
is to treat SPDX-created IDs and third-party-created IDs with the same
logic.  So however many copies of the GPL 2.0 the SPDX wants to keep
to support its various versioned grants, third parties should be able
to keep that many as well, and use the same operators to craft grands
based on their LicenseRef identifiers.

[1]: https://wiki.spdx.org/view/Legal_Team/Minutes/2017-11-09#Agenda

chapters/appendix-IV-SPDX-license-expressions.md
wking: > IMHO In most cases parens are not needed and just make the expression heavier and ugly looking for no good reason.

“We want an unambiguous way to identify the end of this value” is a good reason.  Requiring (or at least strongly recommending) enclosing parens is one way to do that.  Defining clear folding-whitespace rules (#44, #45) is another way.  Forbidding newlines in values is a third way.  All of those approaches have pros and cons, especially in the context of (backwards) compatibility.

And debating this particular issue was not a priority for this PR.  Just tell me what I need to say here to get this landed, and we can come back and kick this particular point around in follow-up work (e.g. #44 and #45).

wking:
I've pushed 922031a → eafa739 which restores semantic docs for `+` (which I'd accidentally dropped with 922031a) and adds a new explanation of the `AGPL-1.0+` based on [this reasoning][1].

[1]: https://github.com/spdx/license-list-XML/issues/542#issuecomment-354306839

iamwillbar:
@goneall and @kestewart, what are your thoughts on this PR for 2.2 vs 3.0?

goneall:
@iamwillbar @kestewart I'm thinking 3.0.  There are a large number of issues mixed into this one PR.  Resolving all of these will likely take a few tech calls.

wking:
I'm happy to split this up if you want to lay out how you want it divided :)

goneall:
@wking Thanks for the offer to split it up.  I think the controversial items to be excluded would be the `+` operator changes for licenseRef's and the parenthesis "MUST" requirement.

The fixes to the RDF would be nice to get into the 2.2 release.

@kestewart Can you take a look and let us know if you agree with that split.

kestewart:

goneall:
This PR is rather stale - closing.  If we want to refactor, we should refactor the current version in the development/v3.0 branch.

PR #39, opened 2017-09-15
Adding SPDX models.  

goneall:
The .vpp file is for Visual Paradigm - a visual modeling tool.

The output graphic can be included in the documentation.

I'm adding this to the repository to keep track of the source modeling document and keep it in sync with the releases.


wking:
On Fri, Sep 15, 2017 at 09:50:12AM -0700, goneall wrote:
> The .vpp file is for Visual Paradigm - a visual modeling tool.

According to Wikipedia, VP-UML has a free community edition [1].  But
I'd rather draw this graphic with a tool that can be run from a
Makefile or other automatic build tooling.  Does GitBook support
anything like that (in which case we could skip the local compiled
form entirely).  Otherwise I'd rather use something like Graphviz [2]
(licensed under the EPL-1.0 [3]), similar to [4,5,6].

[1]: https://en.wikipedia.org/wiki/Visual_Paradigm_for_UML
[2]: http://graphviz.org/
[3]: http://graphviz.org/License.php
[4]: https://github.com/opencontainers/image-spec/blob/v1.0.0/Makefile#L105-L106
[5]: https://github.com/opencontainers/image-spec/blob/v1.0.0/img/media-types.dot
[6]: https://github.com/opencontainers/image-spec/blob/v1.0.0/img/media-types.png

goneall:
I just checked and the tool output to  .XML, .XMI and BPNM-2.0 formats.  Since these are standard, I suggest storing the file in one of these format.  I'm not familiar with any of these standards, but XMI looks like a reasonable choice.  I tested export and import and it seems work find with the tool.

Attached is the file.  @wking let me know if this format looks good to you and I'll update the files accordingly.
[spdx-model.xmi.txt](https://github.com/spdx/spdx-spec/files/1307142/spdx-model.xmi.txt)

wking:
On Fri, Sep 15, 2017 at 05:26:19PM +0000, goneall wrote:
> I just checked and the tool output to .XML, .XMI and BPNM-2.0
> formats.  Since these are standard, I suggest storing the file in
> one of these format.  I'm not familiar with any of these standards,
> but XMI looks like a reasonable choice.  I tested export and import
> and it seems work find with the tool.

XMI looks fine, and it's an open standard.  And ArgoUML seems to
support it, but that project seems dead upstream [1].  It's also
packaged in Gentoo [2], but not Debian [3].

Graphviz, on the other hand, has a most recent commit 16 hours ago [4]
;).  And maybe there are other active FLOSS tools for XMI, I don't
know.

> https://github.com/spdx/spdx-spec/files/1307142/spdx-model.xmi.txt

Genric XMI caveats aside, this looks fine to me.  I'm somewhat
concerned that about half the file (by line count) is part of a
‘<xmi:Extension extender="Visual Paradigm">’ element, but from the
tags I'm guessing that's positioning information which would be
non-critical.  So I'm in favor of XMI over their proprietary file
format, but would still rather have the image source be written in a
format that had active FLOSS tool support.

But I'm also ok just leaving it up to you and others with XMI-handling
tools installed to maintain this content, and we can revisit
translation to another format if/when you leave ;).

[1]: http://argouml.tigris.org/source/browse/argouml/trunk/src/
     Last touched two years ago.
[2]: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/argouml?id=811f1687ed1acf679523759e11b164e6d78210ab
[3]: https://packages.debian.org/search?searchon=contents&keywords=argouml&mode=exactfilename&suite=stable&arch=any
[4]: https://github.com/ellson/graphviz

goneall:
I replaced the .vpp file with a .xmi file.

PR #40, opened 2017-09-27
Slight improvement of language of sect 3

jonasob:
I'm proposing a slight wording change of section 3 to make it clearer the Package information only applies if the SPDX file is grouped by package.

### DIFF
--- a/chapters/3-package-information.md
+++ b/chapters/3-package-information.md
@@ -1,6 +1,6 @@
 # 3 Package Information
 
-One instance of the Package Information is required per package being described. A package can contain sub-packages, but the information in this section is a reference to the entire contents of the package listed. Starting with SPDX 2.0, it is not necessary to have a package wrapping a set of files.
+If you organise your SPDX information by grouping into packages, then one instance of the Package Information is required per package being described. A package can contain sub-packages, but the information in this section is a reference to the entire contents of the package listed. Starting with SPDX 2.0, it is not necessary to have a package wrapping a set of files.
 
 Cardinality: Optional, one or many.
 


chapters/3-package-information.md
```diff
@@ -1,6 +1,6 @@
 # 3 Package Information
 
-One instance of the Package Information is required per package being described. A package can contain sub-packages, but the information in this section is a reference to the entire contents of the package listed. Starting with SPDX 2.0, it is not necessary to have a package wrapping a set of files.
+If you organise your SPDX information by grouping into packages, then one instance of the Package Information is required per package being described. A package can contain sub-packages, but the information in this section is a reference to the entire contents of the package listed. Starting with SPDX 2.0, it is not necessary to have a package wrapping a set of files.
```
wking: Your new wording seems tautological (“if you use Package Information, then Package Information is required”).  I think the difficulty comes from defining cardinality alongside the child element, when it's really a parent property.  I'd rather see each parent clearly define their allowed children, with potential backlinks from children to possible parents.  For an example of this in another spec, see HTML's [`html`][2] and [`p`][3] elements, which have a [normative “Content model”][4] and a [convenience “Contexts in which this element can be used”][5].  Once we shift those around, we could have the root element clearly declare that it could contain zero or more package entries, zero or more file entries, etc.  And the docs in this section would only have to address a single package.

And we currently use [very similar wording for files][1] that we should probably adjust at the same time.

nit: Not sure if the spec has an official en-US vs. en-GB position, but “organize” is currently more popular:

```
$ git describe
v2.1-43-gcfa1b9d
$ git grep -i organis | wc -l
0
$ git grep -i organiz | wc -l
24
```

[1]: https://github.com/spdx/spdx-spec/blame/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/4-file-information.md#L3
[2]: https://www.w3.org/TR/html5/semantics.html#the-html-element
[3]: https://www.w3.org/TR/html5/grouping-content.html#the-p-element
[4]: https://www.w3.org/TR/html5/dom.html#element-dfn-content-model
[5]: https://www.w3.org/TR/html5/dom.html#element-dfn-contexts

jonasob: That makes sense to me as well; it's a somewhat different problem than what I was seeking to address though (which was just a matter of wording). So perhaps a sensible course of action is we move your comment to an issue to be worked on?

wking: > So perhaps a sensible course of action is we move your comment to an issue to be worked on?

Done with #41.

tsteenbe:
Merged in https://github.com/spdx/spdx-spec/commit/df8f54c7af0f56ec607b275e791e3bc06945b892

PR #54, opened 2017-12-13
chapters/3: Explicit external-reference ABNF

### LINKED ISSUE #53: Consider adopting package-url for external package references
https://github.com/package-url is an emerging grass root effort to design a simple and mostly universal package id and locator. This would be a much improved alternative to the partially defined https://github.com/spdx/spdx-spec/blob/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/appendix-VI-external-repository-identifiers.md

wking:
This section has never been particularly tight.  For example, the old:

    `<type>` is an [idstring] that is defined in Appendix.

was probably intended to reference the old Appendix VII (removed by this pull request).  This pull request ties external-references strictly to generic URIs (see discussion in #53).  That breaks compatibility with the old form, but since the old form was unclear, I think that's ok.

The `cpe` entries are already URIs in their own right, although they aren't [registered with IANA][2].  You can resolve them [here][3] (like [this][4]).  There are also other currently-unregistered schemes for referencing packages, e.g. [package URLs][5].  But managing all of that complexity isn't something that SPDX should handle.  Instead, folks interested in providing stable packaging and security references should work on registering their URIs with IANA (or on establishing them in the SPDX ecosystem despite their not being registered).

Fixes #53.

[2]: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
[3]: https://nvd.nist.gov/products/cpe/search
[4]: https://nvd.nist.gov/vuln/search/results?adv_search=true&cpe=cpe%3A2.3%3Aa%3Apivotal_software%3Aspring_framework%3A4.1.0%3A%2A%3A%2A%3A%2A%3A%2A%3A%2A%3A%2A%3A%2A
[5]: https://github.com/package-url/purl-spec


chapters/3-package-information.md
```diff
@@ -946,56 +946,41 @@ Example:
...
+**3.21.4** Data Format: `<category> <URI>`
 
-where:
-
-`<category>` is “SECURITY” | “PACKAGE-MANAGER” | “OTHER”
-
-`<type>` is an [idstring] that is defined in Appendix.
+In Tag:value format, values must match the following [ABNF][rfc5234].
+`URI` is from pRFC 3986][rfc3986-s3].
+`space` is from [Appendi IV](appendix-IV-SPDX-license-expressions.md#overview).
```
wking: `space` is getting added in #37, but if there is interest in getting this PR in first, I can reroll to require a single U+0020 space for the time being.

chapters/3-package-information.md
```diff
@@ -946,56 +946,41 @@ Example:
...
-**3.21.4** Data Format: `<category> <type> <locator>`
+**3.21.4** Data Format: `<category> <URI>`
 
-where:
-
-`<category>` is “SECURITY” | “PACKAGE-MANAGER” | “OTHER”
-
-`<type>` is an [idstring] that is defined in Appendix.
+In Tag:value format, values must match the following [ABNF][rfc5234].
+`URI` is from pRFC 3986][rfc3986-s3].
```
wking: Typo `pRFC` -> `RFC`.

wking:
In 0fc92df → bce5990, I've dropped the 2018-01-04 legal meeting and bumped the legal time from 10am to 9am Pacific to catch up with [this][1].  In the event that the time ends up staying at 10am Pacific, I'll go back and file another adjustment later.

[1]: https://lists.spdx.org/pipermail/spdx-legal/2017-December/002434.html

wking:
I've pushed 849b91a → 6e8b3bc, addressing Yev's concerns from #58 (except for `PACKAGE-MANAGER` → `PACKAGE_MANAGER`).  The `rdf:resource` relationship is now clearer, and I've dropped `OTHER` in favor of allowing authors to supply their own extension categories (which I recommend they describe in the SPDX file and submit to us for future standardization).  I've also added semantic docs for the `SECURITY` and `PACKAGE-MANAGER` values.

wking:
Rebased onto master (around #8) with 9c2b5b7 → 5ecc05d.

The rebase also adds a `CONTRIBUTORS.md` update for the shifted CC-BY-3.0 reference (although the old reference was stale anyway, #73).

chapters/3-package-information.md
```diff
@@ -946,56 +946,59 @@ Example:
 
 **3.21.3** Cardinality: Optional (one or many)
 
-**3.21.4** Data Format: `<category> <type> <locator>`
+**3.21.4** Data Format: `<category> <URI>` <a name="3.21.4"></a>
```
yevster: Even _if_ we had agreed on removing "type", it should first to remain as a deprecated field for one version of the spec. But I don't believed we had agreed on deprecating it. Have we?

wking: > The term URI here can be confusing in an RDF/linked data context, because almost everything there is a URI.

How about `<external-URI>`?  Or `<reference-URI>`?  Or just `<reference>` and leave the URI-ness to the subsequent text?

> Even if we had agreed on removing "type", it should first to remain as a deprecated field for one version of the spec. 

That would be possible.  The 2.1 wording [was][1]:

    <type> is an [idstring] that is defined in Appendix.

The Appendix reference was fixed in 02aaac402b, but that hasn't been released yet.  So I don't think a breaking change with such an underspecified field is a big deal, but if we want to keep support through 3.0 (or even 4.0), I'd be ok with:

```markdown
Data Format: `<category> [<type>] <URI-or-whatever>`
```

> But I don't believed we had agreed on deprecating it. Have we?

This PR is just my recommendations.  Review and approval depends on whether we end up agreeing or not, but agreement doesn't have to be established before the PR is filed ;).

[1]: https://spdx.org/spdx-specification-21-web-version#h.7i829o8ausxy

kestewart:

PR #68, opened 2018-01-03
chapters/9: Fix reference to annotations (section 7 → 8)

wking:
Also:

* Add an anchor to the section 8 header.
* Backtick the code excerpt `annotationType_review` in the line I touch.


tsteenbe:
Thank you for pointing this . Saw the same issue myself and fixed in 097c7a6

PR #70, opened 2018-01-03
chapters/8: Allow rdf:about for referencing external SPDX IDs

### LINKED ISSUE #67: Proposal for digitally signing SPDX 2.0 documents
This was https://bugs.linuxfoundation.org/show_bug.cgi?id=1189
Moving it here so we don't loose track of it, but this is likely to be handled by other projects outside SPDX (OpenChain, SParts (w/ Hyperledger), etc.)

**Gary O'Neall 2014-03-27 23:01:44 UTC**
Based on the discussion at LinuxCon:
Problem statement - Today, there is no way to validate whether SPDX document(s) which have been reviewed have been modified after the review (either the file described by the SPDX document(s) or the metadata in the SPDX document(s)).

Proposal to have a documented best practice for creating a separate file outside of the SPDX documents being reviewed. This document would contain the file names and sha1 checksums for all SPDX documents which have been reviewed (NOTE: This should include any externally referenced SPDX documents).  Additional reviewer comments/annotations would also be included in this separate file.  The resultant file could be digitally signed.

**Jeremiah C. Foster 2014-03-28 08:48:56 UTC**
+1

As an example (only as an example, I'm not saying one has to copy this) here is a URL to a Debian "description" file for the Tomcat package; http://ftp.de.debian.org/debian/pool/main/t/tomcat6/tomcat6_6.0.35-6+deb7u1.dsc

It has the properties described in the orginal bug description for 1189.

**Kate Stewart 2015-03-03 19:01:01 UTC**
This will be addressed after 2.0 is available.

wking:
`rdf:about` seems to be documented [here][1].  There are also a lot of examples using it [here][2], but I couldn't find clear a specification for `rdf:about` there.  What I'm looking for is a way to reference an external subject, the annotation tag is the predicate, and the contents of the annotation element are the object.  Ping @goneall for some RDF/XML assistance.

Also:

* Fix an “annotationg” → “annotating” typo.
* Remove a doubled `<annotation>` element.
* Fix curly quotes in the example I touch.

Fixes #67, by covering the XML case (the tag/value case is already covered since 2.0).

[1]: https://www.w3.org/TR/rdfa-primer/#alternative-for-setting-the-context-about
[2]: https://www.w3.org/TR/rdf-syntax-grammar/

### DIFF
--- a/chapters/8-annotations.md
+++ b/chapters/8-annotations.md
@@ -114,15 +114,19 @@ Example:
 
 **8.4.6** RDF:
 
-For RDF, the annotations are a property of the SPDX element it is annotationg.
+For RDF, the annotations may be a property of the SPDX element it is annotating.
 
-    <SpdxElement rdf:about=”#SPDXRef-45”>
-        <annotation>
-            <Annotation>
-                ...
-            </Annotation>
-        </annotation>
-    </SpdxElement rdf:about=”#SPDXRef-45”>
+    <SpdxElement rdf:about="#SPDXRef-45">
+        <Annotation>
+            ...
+        </Annotation>
+    </SpdxElement rdf:about="#SPDXRef-45">
+
+Alternatively, they may contain an `rdf:about` attribute with an SPDXID.
+
+    <Annotation rfd:about="#SPDXRef-45">
+        ...
+    </Annotation>
 
 ## 8.5 Annotation Comment <a name="8.5"></a>
 


goneall:
> rdf:about seems to be documented here. 

The above referenced link is for RDFa (not RDF XML specifically).  The second link provided is a more consistent with the RDF/XML used as examples in the SPDX document.

The `about`  refers to the element the attribute is attached to.

To solve issue #67 with an annotation, you would use the same format, but just use the URI for the external in the element about.  For example, if you wanted to provide an annotation to an element with SPDX ID `SPDXRef-125` in an external SPDX document with an SPDX document namespace `http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82` and this annotation was not in the same document as the document, you would use the following:
```
   <SpdxElement rdf:about=”http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82#SPDXRef-125”>
        <annotation>
            <Annotation>
                ...
            </Annotation>
        </annotation>
    </SpdxElement>
```

Since we are now referencing an external document in the SPDX document containing the annotation, we should also include the external document reference.  For example:

```
<externalDocumentRef>
       <ExternalDocumentRef>
             <spdx:externalDocumentId>DocumentRef-spdx-tool-1.2</spdx:externalDocumentId>
                  <spdxDocument rdf:about=”http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82.” />
                  <checksum>
                     <Checksum>
                         <algorithm rdf:resource="checksumAlgorithm_sha1"/>
                         <checksumValue>d6a770ba38583ed4bb4525bd96e50461655d2758
                         </checksumValue>
                     </Checksum>
                   </checksum>
        </ExternalDocumentRef>
 </externalDocumentRef>
```

wking:
On Thu, Jan 04, 2018 at 10:40:27AM -0800, goneall wrote:
>    <SpdxElement rdf:about=”http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82#SPDXRef-125”>
>         <annotation>
>             <Annotation>

Why do you have both <annotation> and <Annotation>?  I thought that
was a typo in the current spec.

And why do you need a wrapping SpdxElement instead of:

  <Annotation rdf:about="http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82#SPDXRef-125">
    …
  </Annotation>

?

goneall:
> Why do you have both <annotation> and <Annotation>?  

The outer `<annotation>` (lower case)  is the name of the property.  The inner `<Annotation>` is the name of the class (or type).

wking:
On Fri, Jan 05, 2018 at 01:00:05AM +0000, goneall wrote:
> The outer `<annotation>` (lower case) is the name of the property.
> The inner `<Annotation>` is the name of the class (or type).

Are there docs for that?  I don't see <annotation> mentioned in the
spec outside of examples.

goneall:
There is an [SPDX RDF OWL](https://spdx.org/owl-ontology-21-specification-0) representation and [an HTML representation](https://spdx.org/owl-ontology-21-specification-0) of the RDF (based on the OWL document).  See the [W3C OWL documentation](https://www.w3.org/OWL/) for info on the OWL format.

wking:
On Thu, Jan 04, 2018 at 09:07:16PM -0800, goneall wrote:
> There is an [SPDX RDF
> OWL](https://spdx.org/owl-ontology-21-specification-0)
> representation and [an HTML
> representation](https://spdx.org/owl-ontology-21-specification-0) of
> the RDF (based on the OWL document).

If the normative spec for the RDF/XML is the OWL, then do we need to
cover it in the Markdown?  It seems odd to have partial docs in the
Markdown.

goneall:
I don't know if I would add the OWL to the markdown since most RDF users would import the OWL document to a tool.  We should cross-reference this, however.

wking:
On Fri, Jan 05, 2018 at 07:33:15PM +0000, goneall wrote:
> I don't know if I would add the OWL to the markdown since most RDF
> users would import the OWL document to a tool.

Then drop the RDF notes from the Markdown and say “Documentation for
RDF representations can be found (here|link-to-OWL)”?

chapters/8-annotations.md
```diff
@@ -114,15 +114,19 @@ Example:
...
+    <SpdxElement rdf:about="#SPDXRef-45">
+        <Annotation>
+            ...
+        </Annotation>
+    </SpdxElement rdf:about="#SPDXRef-45">
+
+Alternatively, they may contain an `rdf:about` attribute with an SPDXID.
+
+    <Annotation rfd:about="#SPDXRef-45">
+        ...
```
goneall: I believe the `rdf:about` in this case resolves to the URI for the `Annotation`.  The intent is that the `Annotation` is an `annotation` property of the `SpdxElement`.  You could assign a unique ID for the `Annotation` and refer to that ID.  The example would be:
```
<SpdxElement rdf:about="#SPDXRef-45">
        <annotation rdf:about="#AnnotationId-12"/>
         ...
</SpdxElement rdf:about="#SPDXRef-45">
<Annotation rdf:about="#AnnotationId-12>
    <rdfs:comment>Comment</rdfs:comment>
     ...
</Annotation>
```

goneall:
Discussed on tech call 10 Dec 2019.  We are planning to restructure the examples in 3.0 to resolve this issue.

Opened an issue for 3.0 to track: #162 

Closing this PR

PR #71, opened 2018-01-05
chapters/3: Optional PackageDownloadLocation

wking:
This is WIP, because I still need to fill in some of the Bazaar section.

There's no reason to require `PackageDownloadLocation` if we also allow `NOASSERTION`.  This commit deprecates `NOASSERTION`, and encourages authors to leave the property unset in those cases.

It also allows multiple entries, which can be useful if one download location is temporarily down or if clients are expected to prefer different authorities or protocols.

It also punts the version-control URI specification to a separate appendix.  Ideally VCSes would be registering their own schemes (and there is [a provisional registration for Git][1]).  We can't drop our VCS URI spec completely without backwards-compat concerns, but by shifting this information to an appendix we:

* Provide a more direct place for external consumers to link if they just want to lean on our VCS URI spec (e.g. [package-url][2]).

* Reduce noise for folks who are using IANA-registered schemes.

* Make it easier to deprecate and eventually drop our VCS URI spec as the VCSes register their own schemes with the IANA (hopefully Git) or fall out of common usage (possibly Subversion and Bazaar).

Related “let the IANA handle scheme registration” thoughts in #54 and #53.

While formalizing the VCS URI spec, I've dropped the:

    git+git@git.myproject.org:MyProject

example.  Git [supports the SCP-style syntax][3]:

    [user@]host.xz:path/to/repo.git/

but there's no easy way to formulate that as a URI.  The example I'm removing has its first colon way out towards the end, which, with the [URI ABNF][4] would make all of `git+git@git.myproject.org` the scheme (or a relative URI, since some of those are non-scheme chars).

Removing the SCP style does not reduce generality, because that same resource can be represented as:

    git+ssh://git.myproject.org/MyProject

or:

    git+ssh://git@git.myproject.org/MyProject

depending on how you read the SCP-style example.

I've also changed the examples from myproject.org to example.net, because that's what the IANA has [reserved for examples][5].

[1]: https://www.iana.org/assignments/uri-schemes/prov/git
[2]: https://github.com/package-url/purl-spec/blame/b6fbbf5cea6dae09c0d8ccd12b575efb0e2de8c2/README.rst#L734
[3]: https://git-scm.com/docs/git-fetch#_git_urls_a_id_urls_a
[4]: https://tools.ietf.org/html/rfc3986#section-3
[5]: https://tools.ietf.org/html/rfc2606#section-3

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -16,4 +16,5 @@
 * [Appendix IV: SPDX License Expressions](./chapters/appendix-IV-SPDX-license-expressions.md)
 * [Appendix V: Using SPDX short identifiers in Source Files](./chapters/appendix-V-using-SPDX-short-identifiers-in-source-files.md)
 * [Appendix VI: External Repository Identifiers](./chapters/appendix-VI-external-repository-identifiers.md)
-* [Appendix VII: Creative Commons Attribution License 3.0 Unported](./chapters/appendix-VII-creative-commons-attribution-license-3.0-unported.md)
+* [Appendix VII: Version-Control URIs](./chapters/appendix-VII-version-control-uris.md)
+* [Appendix VIII: Creative Commons Attribution License 3.0 Unported](./chapters/appendix-VIII-creative-commons-attribution-license-3.0-unported.md)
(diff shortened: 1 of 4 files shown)


wking:
Fixed up the Bazaar stuff with a3e663d → c620e81, so I've removed the WIP tag.

nishakm:
@tsteenbe @kestewart I can update this PR if that's OK.

kestewart:
Thanks Nisha,  its a 3.0 change, but definitely a direction we're going in.

goneall:
This is resolved in 3.0

PR #74, opened 2018-01-24
chapters/3: Simplify the verification-code algorithm

wking:
Spun of from #17.  Review that PR first.

Instead of using a newline-delimited file to hold and sort hashes, just store the hashes in an array.  This avoids the need to inject newlines, sort, and then remove the newlines.  Most languages will have a native array structure, so there's no need to use the newline approach as a custom array structure.

Also, use the more intuitive (to me, anyway):

```
  if file is in "excludes" {
    continue  /* exclude SPDX analysis file(s), etc. */
  }
  operate on the file that's not excluded
```

instead of the old:

```
  if file is an 'excludes' file, skip it
    operate on the file that's not excluded
  }
```

### DIFF
--- a/chapters/3-package-information.md
+++ b/chapters/3-package-information.md
@@ -4,10 +4,10 @@ One instance of the Package Information is required per package being described.
 
 Cardinality: Optional, one or many.
 
-In tag:value format, the order in which package and files occur is syntactically significant.
+In `tag:value` format, the order in which package and files occur is syntactically significant.
 
-A new Package Information section is denoted by the Package Name field.
-All Package Information fields must be grouped together before the start of a Files section, if file(s) are present.
+A new Package Information section is denoted by the [Package Name](#3.1) field.
+All Package Information fields must be grouped together before the start of a [Files section](4-file-information.md), if file(s) are present.
 All files contained in a package must immediately follow the applicable Package Information.
 A new Package Information section (via Package Name) denotes the start of another package.
 Sub-packages should not be nested inside a Package Information section, but should be separate and should use a Relationship to clarify.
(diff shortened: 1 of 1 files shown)


sschuberth:
This PR does a lot more reformatting that just improving the pseudo-code for generating the Package Verification code, but I like the changes to the pseudo-code. Once this is merged I wonder whether it makes sense to also add some working examples in a few different different languages, like

Shell:
https://github.com/sschuberth/spdx/blob/3642a042f406774f16cb53d4e204a76cb45d128e/scripts/spdx_package_verification_code.sh#L1-L8

´´´ruby
$package_verification_code = []
scan_results['files'].each do |file|
    next if file['type'] == 'directory'
    sha1 = file['sha1']
    $package_verification_code << sha1
end
$package_verification_code.sort!
$package_verification_code = Digest::SHA1.hexdigest($package_verification_code.join)
´´´

wking:
> This PR does a lot more reformatting that just improving the pseudo-code for generating the Package Verification code…

No it doesn't.  The rest of the changes GitHub is showing are from #17, which is still open (and which my opening comment here says to review first).  The changes unique to this PR are just [these][1].

> Once this is merged I wonder whether it makes sense to also add some working examples in a few different different languages

That's fine with me, but the algorithm is reasonably straightforward on its own, so I'm fine without them too.  If you want them, I think you should make them functions that take an `excludes` array/map, so you can implement that portion of the algorithm as well.

[1]: https://github.com/spdx/spdx-spec/pull/74/commits/d12e74a6df74f25ec4b8f2fd8166596408120ad8

sschuberth:
>  The rest of the changes GitHub is showing are from #17

That's what I mean. Looking solely at the "File changed" tab shows too much. (Unmerged) PRs should be independent for clarity, IMO.

wking:
> Looking solely at the "File changed" tab shows too much. (Unmerged) PRs should be independent for clarity, IMO.

The changes overlap, so orthogonal PRs are not possible.  My understanding was that most of #17 was going into the next release while this algorithm change was going into a later release.  So I don't see a point in rebasing the tip commit onto master now, if I'monly going to end up rebasing it back onto #17 after that lands.

pombredanne:
IMHO verification codes should be optional.

sschuberth:
> IMHO verification codes should be optional.

That discussion should be held outside of this purely cosmetic change, IMO.

kestewart:
Discussed on the call, and we'll be closing this.   Please submit changes in smaller separated PRs that are easier to review and comment on.

PR #87, opened 2018-06-12
Fix incorrect RDF for the annotation type

goneall:


yash-nisar:
@goneall Looks good ! You may proceed with the merging ! :D

Issue #55, opened 2017-12-19
Support multiple SPDX file format like project HayStack 


tsteenbe:
[Project Haystack](https://project-haystack.org/doc/Intro#data) supports multiple formats

• Formats: JSON, Zinc (typed CSV), Trio (YAML)
• RDF/linked data (Microdata, RDFa, JSON-LD, Turtle, SPARQL)

Idea - could we use similar concept to add support for multiple formats in SPDX?

goneall:
Working on a prototype which will support YAML and JSON.

One issue I'm working through is how to flatten the graph of relationships.  If I do a pure JSON or YAML serialization of the objects, I get an infinite loop with bi-directional relationships (the graph of relationships is not restricted to a directed graph for SPDX).

I have a few ideas on how to resolve, some involve minor changes to the spec, other approaches do not.  I'll do a bit more investigation and prototyping before making any proposals.

pombredanne:
Having a DAG is the sane way to go. RDF is nuts there IMHO.

goneall:
> Having a DAG is the sane way to go.

We debated this during the development of SPDX 2.0 at the SPDX leadership conference and concluded on allowing cycles.  Wasn't my favorite approach due to the complications in the tools, but it was the conclusion of the group and is now implemented.

> RDF is nuts there IMHO.

Just a note - the use of graphs with cycles is completely independent of whether we support RDF.  The nice thing is we don't require the use of RDF to use SPDX, you're welcome to use the tag/value format.

goneall:
I prototyped a YAML translator for the SPDX tools.  The source is located in the [yamlsupport](https://github.com/spdx/tools/tree/yamlsupport) branch.

Attached is the YAML format for the example SPDX document.  Note that the example includes a few fields which are deprecated (e.g. reviewer, file dependencies).
[SPDXRdfExample-v2.1.yaml.txt](https://github.com/spdx/spdx-spec/files/1584613/SPDXRdfExample-v2.1.yaml.txt)

goneall:
In the prototype above, I wrapped all of the SPDX elements (files, packages, snippets) in the type name using a simple type (e.g. package, snippet, file).

This greatly simplifies the serialization and deserialization of the YAML files.  It also closely matches the terms used in the RDF spec.

An alternative would be to not wrap the elements and pre-pend the property names with the name of the type (e.g. packageName, fileName, fileChecksum).

[...]

PR #120, opened 2019-04-30
Add examples in non-official formats

swinslow:
As discussed on 2019-04-23 tech team call, this commit adds sample SPDX files in JSON, XML and YAML format, along with a corresponding tag-value format version. These are intended for further discussion and iterating before finalizing any of the new formats.

The versions of the files used in this commit are taken from the comment at https://github.com/spdx/spdx-spec/issues/55#issuecomment-413063219, with the only changes being (1) converting CRLF to LF, and (2) changing the version number in the filenames to 2.0, since these are in fact SPDX v2.0 files.

cc @goneall

Signed-off-by: Steve Winslow <swinslow@gmail.com>

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,5 @@
+## Caution
+
+The examples in this directory (except for the [Tag-Value file](SPDXTagExample-v2.0.spdx)) are NOT official SPDX formats.
+
+The only official formats currently recognized in the SPDX spec are RDF/XML and Tag-Value. The examples in this directory are intended to be working documents, for use by the SPDX team in evaluating additional formats. The schemas seen in these examples are unstable and should NOT be relied upon as final.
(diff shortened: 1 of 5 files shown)


goneall:
@tsteenbe Can you take a look at the travis build? Seems to be stalled asking for a passphrase.

goneall:
@xavierfigueroav  We can can use these examples to compare compatibility between the Java and Python libraries.  Feel free to create pull requests against the examples where you would like to make changes.

swinslow:
Hi @goneall, do you think we can go ahead and merge this PR, since it's in a separate branch that is just intended for discussion / iterating on the proposed new formats? 

I know the CI build is failing, but I'm guessing that doesn't really matter for the separate branch since it's just meant to be for scratch work and discussions.

PR #127, opened 2019-06-03
Added spdx-template

nitish81299:


zvr:
this looks like a Pandoc template for generating a more complete PDF from the Markdown source.

The template is a little bloated, as it has lots of stuff that is not needed in this use case. I'd propose we get it stripped to essentials.

PR #141, opened 2019-09-11
Create appendix-VIII-SPDX-Lite.md

hfukuchi:
A new proposal of the SPDX Lite that is the subset of the SPDX specification. 

[SPDX-Lite-overview-20190905.pdf](https://github.com/spdx/spdx-spec/files/3599056/SPDX-Lite-overview-20190905.pdf)


kestewart:
Missing ssh key for deploy,  which is why travis build can not complete.  
Going ahead and merging at this point.

Issue #140, opened 2019-09-05
Allow for Docker and OCI registry PackageDownloadLocation


nishakm:
Currently the spec for 2.1 says the PackageDownloadLocation should be a URL. Most images are downloaded from a container image registry which does not have the typical URL form of
`<transport>://<host_name>[/<path_to_repository>]`. The list of URLs for different package repositories seems to be listed in the "Appendix VI: External Repository Identifiers". This seems to be a good place to list container registry identifier. Perhaps a link from the PackageDownloadLocation to the External Repository Identifiers would be helpful to communicate that the plain URL is not the only format available for this tag.

nishakm:
I can submit a PR for this as discussed with @kestewart and @swinslow

nishakm:
@goneall says that the PackageDownloadLocation should be a URL only. So this will only add Docker and other container registries to Appendix VI

PR #142, opened 2019-09-17
chapters/appendix-VI: Add container registries

nishakm:
A container image can be considered as a package. These packages
are consumed from container registries. Dockerhub is a public
registry that many single use container images are downloaded.
Google Container Registries are used to download container images
for Kubernetes. This change adds a category type called
CONTAINER-IMAGE and adds to it two types: 'Dockerhub' and
'Google Container Registry'.

The most popular local client used to download these images is
Docker. The docker client defaults to Dockerhub. As a result, the
repo doesn't have a 'domain' name associated with it. If anyone
were to receive a tarball of this container image, they will just
find the repo and tag name.

For other registries though, docker will add the domain name to
the container image. For example, if you used docker to work with
images from GCR, you will find the domain name as part of the repo
name.

Other public container registries can be added to this category

This change resolves issue #140 

Signed-off-by: Nisha K <nishak@vmware.com>

### DIFF
--- a/chapters/appendix-VI-external-repository-identifiers.md
+++ b/chapters/appendix-VI-external-repository-identifiers.md
@@ -118,6 +118,54 @@ External Reference Site: [http://bower.io/](http://bower.io/)
 Documentation: [http://bower.io/docs/api/#install](http://bower.io/docs/api/#install)
 
 ***
+When <category> = `CONTAINER-IMAGE`:
+***
+
+## `<type>` Dockerhub <a name="dockerhub"></a>
+
+### `<locator>` Information
+
+Locator Format:
+
+    repo:tag
+    ^[^:]+:[^:]+$
+
+    repo@digest_type:digest
+    ^[^@]+@[^:]+$
+
+Contextual Example:
+
+    debian:buster
+
+    debian@sha256:2f04d3d33b6027bb74ecc81397abe780649ec89f1a2af18d7022737d0482cefe
+
+External Reference Site: [https://hub.docker.com](https://hub.docker.com)
+
+Documentation: [https://docs.docker.com/](https://docs.docker.com/)
+
+## `<type>` Google Container Registry <a name="gcr"></a>
+
+### `<locator>` Information
+
+Locator Format:
+
+    domain/repo:tag
+    ^[^\/]+\/^[^:]+:[^:]+$
+
+    domain/repo@digest_type:digest
+    ^[^\/]+\/^[^@]+@[^:]+$
+
+Contextual Example:
+
+    gcr.io/distroless/static:latest
+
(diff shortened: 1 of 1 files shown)


swinslow:
Looks good to me! Thank you @nishakm!

nishakm:
@swinslow Do you think this can be included in the `v2.2` spec?

swinslow:
@nishakm I'm in favor of it -- though I defer to @goneall @kestewart as the tech team leads!

goneall:
@tsteenbe Can you take a look at the travis failure on this?  Looks like something is broken in the decryption process.

goneall:
Unless @kestewart objects, I'll go ahead and label this for the 2.2 release.

iamwillbar:
The intent here is good, but I worry we'll chase our tails with the never ending variety of package managers. Should we just have an external reference type of PURL which can be any PURL reference? This will focus the SPDX Specification more and it will encourage contribution of types to the PURL Specification which will give consistency across more tools and scenarios.

swinslow:
@iamwillbar Personally, for 2.2 I'd lean towards including this as-is since there's a ready use case from @nishakm. And then perhaps for 3.0 moving towards a more unified approach using PURL or something comparable.

iamwillbar:
Both of @nishakm's scenarios are supported today by PURL so I think this would be a pretty minor change to her proposal:

```
pkg:docker/debian@sha256:2f04d3d33b6027bb74ecc81397abe780649ec89f1a2af18d7022737d0482cefe
pkg:docker/gcr.io/distroless/static@sha256:9b60270ec0991bc4f14bda475e8cae75594d8197d0ae58576ace84694aa75d7a
```

I'm happy to create a PR demonstrating the change if you'd like?

iamwillbar:
Threw together a quick example in #152.

nishakm:
@iamwillbar I'm totally OK with using PURL, and I would actually like to see it in the spec. This particular PR was meant to take care of an immediate use case because tern creates SPDX documents for container images following the 2.1 spec. Currently, the 2.1 (or the 2.2 spec AFAIK) doesn't have any reference to PURL.

iamwillbar:
@nishakm agreed, but to be fair 2.1 doesn't have CONTAINER-IMAGE either 😄. I think by doing #152 for 2.2 we can a) solve your use case and b) do it in a more general way that will make the specification more stable over time. For example, this PR implements two specific registries whereas PURL supports any registry so the specification wouldn't have to be updated for each registry.

If we included #152 in 2.2 then Tern could generate SPDX documents with PURL instead of CONTAINER-IMAGE and the end result should be the same but it's more flexible.

kestewart:
Agree on call can be handled by #152

kestewart:
Have handled in https://github.com/spdx/spdx-spec/pull/152

PR #145, opened 2019-10-10
Add a relationship for distributed components 

goneall:
Adding this relationship will support the use case where an auditor, analyst, package supplier or packager would like to communicate specific components that are distributed to clients.  For example, JavaScript files which are intended to run on the client machine.

The receiver of the SPDX document would then be able to use this information to determine license obligations for the distributed components.
 
### DIFF
--- a/chapters/7-relationships-between-SPDX-elements.md
+++ b/chapters/7-relationships-between-SPDX-elements.md
@@ -35,6 +35,7 @@
 | AMENDS                 | Is to be used when (current) SPDXRef-DOCUMENT amends the SPDX information in SPDXRef-B.               | (Current) SPDX document A version 2 contains a correction to a previous version of the SPDX document A version 1. Note the reserved identifier SPDXRef-DOCUMENT for the current document is required. |
 | PREREQUISITE_FOR       | Is to be used when SPDXRef-A is a prerequisite for SPDXRef-B | A library `bar.dll` is aprerequisite or dependency for APPLICATION `foo.exe`|
 | HAS_PREREQUISITE       | Is to be used when SPDXRef-A has as a prerequisite SPDXRef-B | An APPLICATION `foo.exe` has prerequisite or dependency of `bar.dll` |
+| DISTRIBUTED_COMPONENT  | Is to be used when SPDXRef-A is an included component of SPDXRef-B and is distributed or conveyed     | A server based APPLICATION `server.war` includes a JavaScript file which is distributed to the client browser `React.js` |
 | OTHER                  | Is to be used for a relationship which has not been defined in the formal SPDX specification. A description of the relationship should be included in the Relationship comments field. | |
 
 **7.1.2** Intent: Here, this field is a reasonable estimation of the relation between two identified elements (i.e. files or packages, or documents), from a developer perspective.


goneall:
I am adding this as a proposed based on the following thread in the OpenChain project:

https://lists.linuxfoundation.org/pipermail/openchain/2019-October/002740.html

swinslow:
Hi @goneall, this is an interesting thought and I can picture ways that it would be useful. 

The part I'm struggling with is that I'm not sure whether "X is a distributed component" is properly described as intrinsic to the Package, File, etc. -- or if it's more about how the Package gets used or distributed by a particular user.

Here's an example of what I'm trying to express: Let's say we've got an open source package X, that contains (1) server software; (2) JavaScript client software; (3) tests; and (4) documentation. Then:

* from the perspective of the **open source project**, all of it is a "distribution component" -- all of it is being distributed.
* from the perspective of a **commercial entity selling the project as a product**, probably all of it is also a "distribution component"; perhaps not the tests, if they don't ship the tests to their customers.
* from the perspective of a **SaaS company hosting the software**, probably just the JavaScript client software and (maybe) the documentation is a "distribution component".

I guess what I'm trying to say is, unlike most (or maybe all?) of the other relationships, this one wouldn't be intrinsic to the software itself. It would necessarily need the context of "who is distributing what to whom" for it to have relevant meaning.

Perhaps that's all fine -- and perhaps it would be useful to be able to express that in SPDX. But I think that's why I'm not immediately a +1 on this.  :)

goneall:
> The part I'm struggling with is that I'm not sure whether "X is a distributed component" is properly described as intrinsic to the Package, File, etc.

I also struggled with the question of is this a property of the package, a relationship, or how it is used.

The scenario I am thinking about is where the server software includes packages which are designed to be distributed based on the design of the server side software.   This is known to the developer of the server side software, but may not be known to the developer of the package being distributed.  In this scenario, "being distributed" not an intrinsic aspect of the package which is being distributed since that software may run on a server or a client.  It is also not an aspect of how the user is using the system - the user is using the server side software and may not even be aware that the distributed package is even being used let alone be involved in the decision as to whether the distributed package is being distributed to the client.  It seems like in this case it would be a relationship of the distributed package to the server side package.

> I guess what I'm trying to say is, unlike most (or maybe all?) of the other relationships, this one wouldn't be intrinsic to the software itself. It would necessarily need the context of "who is distributing what to whom" for it to have relevant meaning.

I think this can be solved by a better definition of the relationship.  We should define the relationship in a way that it is intrinsic to the relationship between the SPDX elements.  What I'm trying to capture is "if this software is used in an SaaS model, then these packages will be distributed".  Maybe we define distributed to be "intended to be distributed to a separate machine from the server software".

In general, I think the description and examples I propose could be improved to make this scenario easier to understand.  I struggled with the balance of keeping the description concise and making it descriptive enough.

BTW - the reason I am interested in including this is that for all of the audits I do for SaaS, I identify the client side components which are distributed.  Having this relationship would allow me to include this information in a standard SPDX file.

leonschwartz:
Just to add to Gary's points above -- one of the most time-consuming things during a typical review of third party software usage (for the purposes of determining compliance) is collecting the usage information that is needed to perform the review.

Currently, if Package X is being used, I'd have to ask the developers to tell me about its dependencies and whether those dependencies are modified, linked, distributed, etc.  Putting this relationship information in the SPDX file would allow me to get this information without having to ask developers for it (and would save the trouble of multiple people asking for the same information at different times).

I'm sure that there are many different cases to think through to make sure that we get this right, but I do think it can be quite useful.

iamwillbar:
Agree that this is a very common and useful scenario, it would be good to work through some example use cases for this and make sure that the model holds. How would you feel about moving this to the 3.0 work stream where we might have some more options?

swinslow:
I'm on board with moving this to consideration for 3.0. Given the more extensive changes we're discussing there, I could see this fitting well into some aspect of that, whereas some of the considerations I described above would make me a bit more hesitant for 2.2.

kestewart:

PR #147, opened 2019-10-18
Merging the ner formats into the v2.2 branch

goneall:
Per the tech call on 15 Oct 2019 - we will merge in the current examples and make pull requests against the examples when we resolve specific issues.


goneall:
Oops - merged the wrong direction - closing

PR #150, opened 2019-10-24
Update Charter to reflect the broader uses of SPDX

iamwillbar:
The Charter as it stands today is very specific to licensing and intellectual property concerns which unnecessarily scopes the usefulness of SPDX for potential adopters.

### DIFF
--- a/chapters/1-rationale.md
+++ b/chapters/1-rationale.md
@@ -2,7 +2,7 @@
 
 ## 1.1 Charter <a name="1.1"></a>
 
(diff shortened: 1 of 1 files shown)


kestewart:
Thanks William,   Really like the revised wording,  it is more inclusive and reflective of the project's directions,  so  +1 on substance of change.  

Looks like the PR has a bit of cruft (whitespace/EOL issue) in it though at the end.   Can that be fixed?

iamwillbar:
> Looks like the PR has a bit of cruft (whitespace/EOL issue) in it though at the end. Can that be fixed?

My `git`-fu is failing me and I'm having trouble removing that cruft, but I'll keep trying.

PR #152, opened 2019-10-24
chapters/appendix-VI: Add PURL

iamwillbar:
This is an alternate implementation for #142 which would hopefully meet @nishakm's requirements with the benefit of being more flexible as it allows many package types and wouldn't require the specification to be updated for each registry.


kestewart:
Agreed on call ok to apply.

PR #165, opened 2019-12-14
Correct annotation comment purpose

goneall:
The purpose for the annotation comments states that it is optional - the cardinality for this field is required.


kestewart:
Agree.  This inconsistency needs to be fixed.   Reviewed in meeting.  Applied.

PR #166, opened 2019-12-17
Fix RDF example for ExternalDocumentRef

goneall:
The RDF is defined to have 2 properties (`spdxDocument` and `checksum`).  The example has the DocumentRef ID as a property.  It should be the ID for the `externalDocumentRef` property.


kestewart:
Thomas reviewed in meeting.  No objections raised.   Merging.

PR #172, opened 2020-01-10
RFC: chapters: in-toto add in-toto chapter

SantiagoTorres:
This is a WIP proposal. I littered it with FIXMES for now, as I'm not completely sure the semantic that I've proposed works as expected...

More than happy to have some feedback!


kestewart:
From call,  
- need top level to reflect its use rather than in-toto keyword.
- concept is complimentary with SPDX, so no problem with approach and including.
- need to add para a the top to start to introduce this section.
- we need to see some examples to illustrate.
- rather than name, just refer to SPDX-ID - that has identification.

- possibly this could be handled as an annotation type?   first approach was trying to do this. 
Keeping them grouped together has a "provenence annotations" is direction from today's meeting.

From discussion,  move to 3.0.

PR #176, opened 2020-01-15
Update Specification Version from 2.1.0 to 2.2.0

iamwillbar:
Note that this does point at non-existent 2.2 RDF images and ontology which should be resolved before publishing.


iamwillbar:
This actually wasn't merged, I deleted the branch. I'll submit a new PR for this shortly.

PR #178, opened 2020-01-16
Update Specification Version from 2.1.0 to 2.2.0

iamwillbar:
Note that this does point at non-existent 2.2 RDF images and ontology which should be resolved before publishing.


goneall:
I added issue #181 to track the missing images and ontology.  Those will be updated after 2.2 is stable and before the release.

PR #183, opened 2020-01-21
Replace uses of obsolete GPL license ids

### LINKED ISSUE #89: Update references to GNU-family licenses in spec
In connection with #6 to update the License List Appendix to the latest version, any references to the GNU family of licenses (GPL, LGPL, AGPL, GFDL) that are used in the spec should be updated to the new ID format for these licenses.

Old format: GPL-2.0; GPL-2.0+
New format: GPL-2.0-only; GPL-2.0-or-later

iamwillbar:
This resolves #89, but it does not deprecate the `+` suffix notation. I'd like to do that as a separate PR.


chapters/appendix-V-using-SPDX-short-identifiers-in-source-files.md
```diff
@@ -45,9 +45,10 @@ Multiple licenses can be represented using a SPDX license expression as defined
 
 Examples:
 
-    SPDX-License-Identifier: (GPL-2.0 OR MIT)
-    SPDX-License-Identifier: (LGPL-2.1 AND BSD-2-CLAUSE)
-    SPDX-License-Identifier: (GPL-2.0+ WITH Bison-exception-2.2)
+    SPDX-License-Identifier: (GPL-2.0-only OR MIT)
+    SPDX-License-Identifier: (LGPL-2.1-only AND BSD-2-CLAUSE)
```
swinslow: Both here and in the original, it looks like the "BSD-2-CLAUSE" was incorrectly capitalized. The correct version should be "BSD-2-Clause". 

I think that capitalization might not matter for license identifier syntax, but it would be great if we can match the capitalization that's on the license list for consistency's sake.

chapters/appendix-V-using-SPDX-short-identifiers-in-source-files.md
```diff
@@ -45,9 +45,10 @@ Multiple licenses can be represented using a SPDX license expression as defined
...
+    SPDX-License-Identifier: (GPL-2.0-or-later[p;04ejr4\,er4\5j\ .] WITH Bison-exception-2.2)
```
swinslow: Did something get added incorrectly here following the `-or-later` addition? There seems to be an odd bracketed bit that I'm not sure what it's doing there...

swinslow:
Thanks @iamwillbar! I just left a review with a couple of nit edits.

Regarding the `+` syntax, I expect that should stay in place for now. There are a handful of other licenses with multiple versions where `+` can be used. The `-only` / `-or-later` bit is due to a request from the license steward in the particular case of the *GPL family of licenses.

chapters/appendix-V-using-SPDX-short-identifiers-in-source-files.md
iamwillbar: Me either, I blame my cat! I will get this fixed up.

chapters/appendix-V-using-SPDX-short-identifiers-in-source-files.md
iamwillbar: Agreed.

chapters/appendix-V-using-SPDX-short-identifiers-in-source-files.md
iamwillbar: Fixed.

iamwillbar:
@swinslow I've resolved your feedback. If there's no further feedback from anyone by end of week I'll merge it into the `development/v2.2` branch.

PR #184, opened 2020-01-22
Fixed build error due to SPDX Lite not being in mkdocs

iamwillbar:


kestewart:
Thanks for adding in this fix.

PR #185, opened 2020-01-22
Added recommended extensions for new formats

### LINKED ISSUE #170: Defined file extensions for the various SPDX formats
To make it easier for tooling to easily figure out how to parse an SPDX file I propose to define file extensions for the various SPDX formats.

Proposed File Extensions
- .spdx - Use for SPDX doc in tag-value format
- .spdx.json - Use for SPDX doc in JSON format 
- .spdx.rdf - Use for SPDX doc in RDF format
- .spdx.yml - Use for SPDX doc in YAML format
- .spdx.xml  - Use for SPDX doc in XML format

iamwillbar:
Resolves #170.

### DIFF
--- a/chapters/1-rationale.md
+++ b/chapters/1-rationale.md
@@ -76,7 +76,15 @@ In an SPDX document, Relationship elements can be used to indicate relationships
 
 **1.7.8** Tags and RDF properties are case sensitive.
 
-**1.7.9** Should be easy to recognize in a file system without opening the file. A suggested naming convention is to use \*.spdx (for `tag:value` format) and \*-spdx.rdf for RDF format.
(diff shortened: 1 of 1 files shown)


swinslow:
@iamwillbar Personally I like this and give it a thumbs-up. 

I do note though, the comment from @goneall in the issue thread [here](https://github.com/spdx/spdx-spec/issues/170#issuecomment-572301789) which suggested `.spdx.rdf.xml` instead of just `.spdx.rdf`.

Was the thinking to stick with `.spdx.rdf`, since that's what the spec currently suggests for RDF files?

iamwillbar:
@swinslow for 2.2 I thought we wouldn't want to change the recommendation, but I could be convinced either way. For 3.0 we could definitely change the recommendation.

PR #203, opened 2020-02-13
Upgrade to latest mkdocs

iamwillbar:


kestewart:
Discussed on call,  all good with moving to python 3 and the new version of mkdocs.

PR #204, opened 2020-02-18
Update Appendix V title to reference License List

kestewart:
With the inclusion of other tags to be included in source files (Appendix IX), make
it clear that these tags are based on license list.


swinslow:
Thanks @kestewart! A quick thought -- would it be useful to note that LicenseRef- strings can be used for licenses not on the license list, since that's how REUSE is noting this should be used? That occurred to me just now because that would be a bit different from saying that these could be used just for identifiers on the License List.

If so, I can add a quick sentence that should reflect this correctly.

kestewart:
I'll go ahead and apply this if you're ok,  and if you could submit a separate sentence, that's a good idea.

PR #205, opened 2020-02-18
Add JSON-LD context file

goneall:

Add a JSON-LD context file defining linked data for properties and property types.

This context file can be quite useful for tools developers needing the types for JSON properties.

This file includes types and property URI's for all of the properties defined in the ontology OWL file.

It also includes properties unique to the JSON format:
- plural form for arrays of properties
- SPDX identifiers
- SPDX external document reference ID's
- "Document" property

Note that some of the plural forms do have issues as they are machine generated.

This file is generated by a utility which converts the RDF OWL document:  https://github.com/goneall/Spdx-Tools-Java/blob/master/src/main/java/org/spdx/tools/RdfSchemaToJsonContext.java


ontology/spdx-ontology.context.json
```diff
@@ -0,0 +1,422 @@
...
+    },
+    "relatedSpdxElement" : {
+      "@id" : "spdx:relatedSpdxElement",
+      "@type" : "spdx:SpdxElement"
+    },
+    "licenseInfoFromFiles" : {
+      "@id" : "spdx:licenseInfoFromFiles",
+      "@type" : "spdx:SimpleLicensingInfo"
+    },
+    "licenseInfoFromFiless" : {
```
swinslow: @goneall Looks like there's an accidental double plural here ("Filess"). I note that there's a single-plural "licenseInfoFromFiles" just above it, can these be merged?

goneall: This is an artifact of the automatic adding of plurals for the JSON format.  This one happens to already be plural.

I'll need to add a special check for this in the tool that generates both the JSON format and the files.

goneall:
> is the idea here to provide a definition for the equivalent of everything that's in the OWL definitions? Or only those things that are actually used in the JSON format definition? 

This is primarily intended for the JSON format, but it is generated from the OWL document.  

>  Partly I'm thinking of things in here like "licenseException" and "isFsfLibre" -- I gather those might make sense in the OWL / RDF world, but I'm not sure they're applicable for JSON where license expressions are just strings.

There is a JSON format for the SPDX listed license that do use these terms.  Even though they are not in the SPDX document JSON form, the terms would still be useful for users of the listed licenses.

We could split these into 2 separate JSON context files, but that would create a lot more work for the tooling.

swinslow:
Got it, thanks @goneall! I'm good with this then after the `licenseInfoFromFiless` tag is updated. Thanks for pulling all of this together!

goneall:
@swinslow @tsteenbe  I just fixed the double ss and added types for data properties (e.g. strings, integers).

If you could do a quick review and merge if OK.

goneall:
@tsteenbe Can you see if you can get this merged?  I've tried 3 times to rebase / merge / merge then rebase/ etc and I keep getting merge conflicts.  If not, I'll submit a new PR on a new branch.  It is only one file.

goneall:
Replaced with PR #211

Issue #28, opened 2017-09-05
Add optional AttributionText field to File & Package level. 


kestewart:
Agreement from those on the call,   yes it should be at File & Package level.
This will be an optional field. 
Useful for generating notice files, etc.

Consider does it make sense for snippets? 

Adding “Attributions” to package field to store required attributions (per Oliver)
Alexios: property named FileAttributionText which will hold the text that has to be reproduced.
It can be considered as a combination of information found in properties like FileCopyrightText, LicenseInfoInFile, FileContributor, but it might not simply the sum of these values.
 
The relative part in the spec could be something like:
 
4.xx   File Attribution Text
4.xx.1 Purpose: This field provides a place for the SPDX data creator to record all attributions found in the file that are required to be communicated. These typically include copyright statement(s), license text, and a disclaimer.
4.xx.2 Intent: The intent is to provide the recipient of the SPDX file with all the legally required attributions in the file, therefore complying with the license obligations.
4.xx.3 Cardinality: Optional, one.
4.xx.4 Data Format: free form text that can (and usually will) span multiple lines
4.xx.5 Tag: "FileAttributionText:"
In Tag:value format, the multiple lines are delimited by "<text>" and "</text>".
Example:

```
FileAttributionText: <text>
#   Copyright (C) 2004 Free Software Foundation, Inc.
#   Written by Scott James Remnant, 2004
#
# This file is free software; the Free Software Foundation gives
# unlimited permission to copy and/or distribute it, with or without
# modifications, as long as this notice is preserved.
</text>
4.xx.6 RDF: property fileAttributionText in class spdx:File
Example:
    	<File rdf:about="...">
           	<fileAttributionText>
#   Copyright (C) 2004 Free Software Foundation, Inc.
#   Written by Scott James Remnant, 2004
#
# This file is free software; the Free Software Foundation gives
# unlimited permission to copy and/or distribute it, with or without
# modifications, as long as this notice is preserved.
           	</fileAttributionText>
    	</File>
```

wking:
On Tue, Sep 05, 2017 at 09:31:30AM -0700, Kate Stewart wrote:
> 4.xx.1 Purpose: This field provides a place for the SPDX data
> creator to record all attributions found in the file that are
> required to be communicated.

This can depend on the context.  For example, comparing BSD-3-Clause
and BSD-4-Clause, both require the copyright notice to be preserved in
source copies, but only the BSD-4-Clause requires a:

  This product includes software developed by …

[...]

PR #208, opened 2020-02-19
Attribution text field

kestewart:
Add in the Attribution Text field to Package and File level,  per discussion in issue https://github.com/spdx/spdx-spec/issues/28.   This makes it easier for tools to satisfy terms found in some licenses (like BSD-4-Clause, etc.).

### DIFF
--- a/chapters/3-package-information.md
+++ b/chapters/3-package-information.md
@@ -1050,33 +1050,33 @@ Example:
     
 ## 3.23 Package Attribution Text <a name="3.23"></a>
 
-**4.15.1** Purpose: This field provides a place for the SPDX data creator to record all attributions at the package level that are required to be communicated. These typically include copyright statement(s), license text, and a disclaimer.
+**3.23.1** Purpose: This field provides a place for the SPDX data creator to record, at the package level, acknowledgements that may be required to be communicated in some contexts. This is not meant to include the package's actual complete license text (see `PackageLicenseConcluded`, `PackageLicenseDeclared` and `PackageLicenseInfoFromFiles`), and may or may not include copyright notices (see also `PackageCopyrightText`). The SPDX data creator may use this field to record other acknowledgements, such as particular clauses from license texts, which may be necessary or desirable to reproduce.
 
-**4.15.2** Intent: The intent is to provide the recipient of the SPDX file with all the legally required attributions at a package level, therefore complying with the license obligations.
+**3.23.2** Intent: The intent is to provide the recipient of the SPDX file with acknowledgement content at a package level, to assist redistributors of the package with reproducing those acknowledgements. This field does not necessarily indicate where, or in which contexts, the acknowledgements need to be reproduced (such as end-user documentation, advertising materials, etc.) and the SPDX data creator may or may not explain elsewhere how they intend for this field to be used.
 
-**4.15.3** Cardinality: Optional, one or many.
+**3.23.3** Cardinality: Optional, one or many.
 
-**4.15.4** Data Format: free form text that can (and usually will) span multiple lines.  
+**3.23.4** Data Format: free form text that can span multiple lines.
 
-**4.15.5** Tag: `AttributionText:`
+**3.23.5** Tag: `PackageAttributionText:`
 
 In `tag:value` format multiple lines are delimited by `<text> .. </text>`.
 
 Example: 
 
-    AttributionText: <text>
-All advertising materials mentioning features or use of this software must display the
-following acknowledgement:  This product includes software developed by the AT&T.
+    PackageAttributionText: <text>
+    All advertising materials mentioning features or use of this software must display the
+    following acknowledgement:  This product includes software developed by the AT&T.
     </text>
 
-**4.15.6** RDF: property 'attributionText' in class 'spdx:Package'
+**3.23.6** RDF: property 'attributionText' in class 'spdx:Package'
 
 Example:
 
     <Package rdf:about="...">
            	<attributionText>
-All advertising materials mentioning features or use of this software must display the
-following acknowledgement:  This product includes software developed by the AT&T.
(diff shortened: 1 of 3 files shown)


chapters/3-package-information.md
```diff
@@ -1047,6 +1047,38 @@ Example:
...
         ...
     </spdx:package>
+    
+## 3.23 Package Attribution Text <a name="3.23"></a>
+
+**4.15.1** Purpose: This field provides a place for the SPDX data creator to record all attributions at the package level that are required to be communicated. These typically include copyright statement(s), license text, and a disclaimer.
+
+**4.15.2** Intent: The intent is to provide the recipient of the SPDX file with all the legally required attributions at a package level, therefore complying with the license obligations.
+
+**4.15.3** Cardinality: Optional, one or many.
```
goneall: I think the many cardinality is fine, but we could define this to be optional 1 only and require the text field to contain all necessary attribution text in a single field.

swinslow:
I wasn't involved in the tech team when it looks like this was discussed in 2017, so I can't speak to what was discussed then.

That said, I have concerns about including something where the purpose is to say "data in this field is legally required, to comply with license obligations." I think that SPDX (at least currently) doesn't get into that:

> 1.5 What is not covered in the specification?
> 
> . . .
> 
> 1.5.4 Legal interpretation of the licenses or any compliance actions that have been or may need to be taken.

I'm not sure how to add a field with the purpose described here, in a way that doesn't essentially say "the license interpretation requires that you do X to comply."

If someone wants to include the text from the notice, though, is there a reason that the FileNotice section isn't sufficient? Could we add a corresponding PackageNotice section instead, with the same semantics but at the Package level?

chapters/3-package-information.md
kestewart: It was originally 1,  but I could see where tools might be picking up fragments from various contributors under different licenses in same file which might be awkward to collapse together, so figured multiple didn't hurt anything.

kestewart:
Good point @swinslow.   I was going along with what was in the issue- but you're right in that we probably shouldn't state it this way.   Would the following address your concerns?
 
**4.15.2** Intent: The intent is to provide the recipient of the SPDX file with any attributions needed to comply with the stated license obligations.

swinslow:
Hi @kestewart, I'm not trying to be picky, I promise :) but I think that still says essentially the same thing? "needed to comply with the stated license obligations" seems like it is still saying "the license requires you to do X with this."

Perhaps part of the problem is that I'm not clear what exactly is supposed to go in this field. The "Purpose" description in the Package instance says that "These typically include copyright statement(s), license text, and a disclaimer" but the example there seems to be more of an "advertising clause" notification.

Should this perhaps be moved to 3.0 rather than 2.2? That way we can work through the specifics of what this should include and also whether it involves a scoping change to SPDX.

kestewart:
Hmm... We should probably have talked this, when we met earlier today @swinslow   
Its something that FOSSology has requested, as they are collecting this information today and need to be able to share it with other tools,  so it should go in 2.2 - they've been waiting long enough.   Advertising clauses in licenses need to be complied with, esp. when they've been customized, and being able to pull this out is key for generating some compliance documentation, esp. for binaries.  So it is complying with the license to provide the information in the advertising clause.   If you don't like my revised wording, feel free to suggest something.

swinslow:
I'm happy to weigh in on the language, but it isn't clear to me what data they want this field to contain. I could imagine it being any of the following:

1) The text that needs to be included to comply with license notice requirements generally (e.g., not specific to "advertising clauses"). This is what the File example in this PR seems to show: _This file is free software; the Free Software Foundation gives unlimited permission to copy and/or distribute it, with or without modifications, as long as this notice is preserved._ -- in most cases, this would presumably include the full text of most licenses.

or

2) The text of just the clause _describing_ an "advertising clause" requirement. This is what the Package example in this PR seems to show: _All advertising materials mentioning features or use of this software must display the following acknowledgement:  This product includes software developed by the AT&T._

or

3) The actual text that needs to be reproduced in advertising, documentation, etc.: _This product includes software developed by the AT&T._

I don't know which of these options they want to go in this field. If you or they know which of these it is, or if it's something different, I'm happy to help wordsmith.

(For what it's worth, "advertising clause" isn't really a consistently defined term across licenses -- some require notices in documentation, some in advertising, I'm sure some are worded to cover other things. I'm not sure which of these they want to capture, if it isn't just "all license text" generally.)

swinslow:
Discussed on 2020-02-25 tech team call: suggest not specifying where it needs to be reproduced; explicitly not intended to be full license text; intended to be other acknowledgements to be reproduced somewhere beyond just the retention of the license text. cardinality: optional, many

carmenbianca:
Small complication that was mentioned *outside of the scope* of this issue: If one wanted to embed the notice in a source file, the notice would have to be single-line, which may or may not be limiting.

e.g.:

```
# SPDX-FileAttributionText: Copyright (C) 2004 Free Software Foundation, Inc.
#   Written by Scott James Remnant, 2004
#
# This file is free software; the Free Software Foundation gives
# unlimited permission to copy and/or distribute it, with or without
# modifications, as long as this notice is preserved.
```

^ This is not valid.

chapters/3-package-information.md
```diff
@@ -1047,6 +1047,38 @@ Example:
...
+
+**3.23.4** Data Format: free form text that can span multiple lines.
+
+**3.23.5** Tag: `PackageAttributionText:`
+
+In `tag:value` format multiple lines are delimited by `<text> .. </text>`.
+
+Example: 
+
+    PackageAttributionText: <text>
```
mcjaeger: In case you would like to have a license text here (and not a company): 

```
This product includes software developed by the organization.
```

from:

https://spdx.org/licenses/BSD-4-Clause.html

PR #209, opened 2020-02-24
App. X: Define custom license namespaces

### LINKED ISSUE #113: Expand License namespaces available
From Mark Atwood: 

**Proposed**: two methods for for organizations to self-declare SPDX license identifiers.
The two methods will be referred to a “organization names” and “dns names”.

**Organization Name:**
The SPDX Project maintains a list of organization names in github.
Organizations submit their entry into this list via github pull requests.
SPDX core maintainers sanity check the entry, discuss in meeting, and accept the pull request.
The entry contains a URL pointing to a SPDX doc maintained by the organization.
That file can contain the license definitions.
That file can contain XML URL external refers that point to more SPDX docs.
Completely ordinary XML parsing and external reference handling.

An example SPDX license tag:  “License-Ref--amazon-ASL-1.1”.  Notice the two hyphens.

The URL for the entry could be https://github.com/aws/spdx/registration.xml
The URL for the ASL-1.1 license could be https://github.com/aws/spdx/asl-1.1.xml

**DNS Name:**
An organization already has a DNS name, assigned via the ordinary DNS registration and name arbitration system.
The organization uses its DNS name, but reversed, similar to a JVM class name.
For example: “.com.amazon.”.  Notice the leading and trailing dots, which are usually implicit in the DNS, but are made explicit here.
By using the DNS name, the organization can immediately start using their namespace without concern that it will conflict with a different organization.

An example SPDX license tag:  “License-Ref-.com.amazon.-ASL-1.1”.

The organization may and should send a pull request to the SPDX github, treat similar to the “Organization Name” process.

swinslow:
This commit is an initial attempt to define the custom license namespace concept that was agreed upon in Spring 2019, as reflected in https://github.com/spdx/spdx-spec/issues/113 and tech team minutes, but which was not subsequently specified.

This defines the DNS-style version of the `LicenseRef-` format. It also describes the process for a namespace maintainer to create and publish an SPDX document at a URL within that domain, where anyone can find the license text corresponding to each identifier defined within that namespace.

This is a first cut to capture discussions from nearly a year ago, and I am sure that things could be worded better. Please, please throw rocks at it.  :)  cc @goneall @kestewart @MarkAtwood @pombredanne @zvr @tsteenbe @iamwillbar as well as anyone else who wants to weigh in.

Fixes #113 

### DIFF
--- a/chapters/appendix-X-custom-license-namespaces.md
+++ b/chapters/appendix-X-custom-license-namespaces.md
@@ -12,18 +12,18 @@ Subsequently, interest has been expressed in being able to use stable identifier
 
 ## Brief overview
 
-This appendix defines a format and mechanism for using the existing `LicenseRef-` format in a manner that enables stable identifier usage. Under this system:
+This appendix defines two formats and mechanisms for using the existing `LicenseRef-` format in a manner that enables stable identifier usage. Under this system:
 
-1. each license "namespace" consists of a standard order DNS name (such as `example.com`) that is under the control of the namespace maintainer's organization;
-2. the namespace maintainer creates an SPDX document defining license texts and corresponding identifiers with the DNS name included, as described below, and publishes it publicly at a URL under the namespace domain;
+1. each license "namespace" consists of either a standard order DNS name (such as `example.com`) or an organization name (such as `ExampleCorp`) that is under the control of the namespace maintainer's organization;
+2. the namespace maintainer creates an SPDX document defining license texts and corresponding identifiers with the DNS name or organization name included, as described below, and publishes it publicly at a URL under the namespace domain;
 3. the namespace and the SPDX document URL is registered with the SPDX project; and
 4. identifiers in the format described below can then be used by anyone, as long as they are defined in the SPDX document available at the registered URL.
 
-## Format
+## Format 1: DNS name
 
-For purposes of this appendix, a valid "namespace" is a standard order DNS name, such as `example.com`. Each custom license identifier under this namespace consists of the following, in order:
+For purposes of this appendix, a valid "namespace" using the DNS name format is a standard order DNS name, such as `example.com`. The DNS name, when represented here, can only use characters permitted by the license expression syntax (e.g., letters, digits, "-" and ".". It should not contain a period followed by a hyphen, as that denotes the end of the namespace. Each custom license identifier under this namespace consists of the following, in order:
 1. the prefix `LicenseRef-`
-2. the namespace, preceded by a period, and followed by a period and hyphen (e.g., `.example.com.-`)
+2. the namespace, preceded by a **period**, and followed by a **period and hyphen** (e.g., `.example.com.-`)
 3. any other characters permitted by the license expression syntax (e.g., letters, digits, "-" and ".")
 
 For example, for the `example.com` namespace, the following could be valid identifiers:
@@ -31,11 +31,23 @@ For example, for the `example.com` namespace, the following could be valid ident
 * `LicenseRef-.example.com.-EULA-v3.1`
 * `LicenseRef-.example.com.-anything`
 
+## Format 2: Organization name
+
+For purposes of this appendix, a valid "namespace" using the organization name format is the organization's name, such as `ExampleCorp`. The organization name, when represented here, can only use characters permitted by the license expression syntax (e.g., letters, digits, "-" and ".". It should not contain a double hyphen, as that denotes the end of the namespace. Each custom license identifier under this namespace consists of the following, in order:
+1. the prefix `LicenseRef-`
+2. the namespace, preceded by **two hyphens**, and followed by **two hyphens** (e.g., `--ExampleCorp--`)
+3. any other characters permitted by the license expression syntax (e.g., letters, digits, "-" and ".")
+
+For example, for the `ExampleCorp` namespace, the following could be valid identifiers:
+* `LicenseRef--ExampleCorp--XYZ-1.0`
+* `LicenseRef--ExampleCorp--EULA-v3.1`
+* `LicenseRef--ExampleCorp--anything`
+
 ## Prerequisites
 
 There are two required prerequisites before using the custom license namespace format.
 
-First, the creator of the identifiers must create an SPDX document that defines the custom identifiers and their corresponding license texts. This should be done using [Other Licensing Information](./6-other-licensing-information-detected.md) sections. The defined identifiers should be in the format described above. Namespace maintainers may update this document from time to time in order to add new identifiers, but they must not modify or remove previously-published identifiers.
+First, the creator of the identifiers must create an SPDX document that defines the custom identifiers and their corresponding license texts. This should be done using [Other Licensing Information](./6-other-licensing-information-detected.md) sections. The defined identifiers should be in the applicable format described above. Namespace maintainers may update this document from time to time in order to add new identifiers, but they must not modify or remove previously-published identifiers. The SPDX document should not define any licenses with identifiers other than those in its own namespace.
 
 Second, the URL for the namespace's definition document must be registered with the SPDX project. The purpose for this is to ensure that consumers of the namespaced license identifiers are able to find the document defining the license text corresponding to those identifiers. This registration should occur using a mechanism to be provided by the SPDX project and made publicly available, which may consist of pull requests to a public repository or another custom tool for this purpose.
 
--- a/chapters/appendix-X-custom-license-namespaces.md
+++ b/chapters/appendix-X-custom-license-namespaces.md
@@ -1,0 +1,48 @@
+# Appendix X: Custom License Namespaces
+
+## Rationale
+
+The [SPDX License List](https://spdx.org/licenses) defines a curated subset of the most commonly used free software / open source licenses, together with other licenses that have been used in historical, prominent and/or widely used projects for other forms of open content. When a license is added to the SPDX License List, it receives a stable license identifier aligned to a specified license text, with markup and matching guidelines to capture non-substantive variations to the text. The SPDX Legal Team maintains the License List at https://github.com/spdx/license-list-XML and enables community participation in the review and addition of licenses.
+
+The SPDX specification also defines a mechanism for creators of SPDX documents to define custom license texts and identifiers. This enables SPDX documents to refer to licenses that might never be included on the SPDX License List, for a variety of reasons (whether due to limited breadth of use, failure to meet the [license inclusion principles](https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md), or otherwise).
+
+These custom license identifiers are prefixed with `LicenseRef-` as defined in [Appendix IV of the SPDX 2.2 Specification](./appendix-IV-SPDX-license-expressions.md). The `LicenseRef-` identifier format was originally assumed to be used in the context of an SPDX document, so that the license text for that identifier would be specified in a corresponding [Other Licensing Information](./6-other-licensing-information-detected.md) section.
+
+Subsequently, interest has been expressed in being able to use stable identifiers for custom licenses outside the context of a single SPDX document. For example, developers might wish to insert [SPDX short-form license identifiers](https://spdx.org/ids) into their source code, for licenses that are not on the License List. Additionally, tools may want to interchange information about custom licenses without having to assume that they are both using the same internal format to generate `LicenseRef-` identifiers, or without having to compare defined text from an Other License Information section of an SPDX document to confirm that they are in fact the same text.
+
+## Brief overview
+
+This appendix defines a format and mechanism for using the existing `LicenseRef-` format in a manner that enables stable identifier usage. Under this system:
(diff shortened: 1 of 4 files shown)


goneall:
@swinslow Thanks for pulling this together!  A nicely detailed proposal.

It would be good to have the legal team review - especially Mark Atwood who made the original proposal.

I recall 2 separate discussions which would impact the proposal.  I can't find the notes on the exact discussions, so this is a bit from memory.  I'm hoping some folks on the tech or legal team can provide a bit more detail.

- Optional rather than required registration of the namespace - For the DNS namespace, the registration and public documentation of the licenses would be encouraged, but not required.  I recall some on the legal team wanted to maintain an internal list and still allow references in SPDX documents.
- Non-DNS formatted namespaces - I believe this came from a tech call.  There was a concern that some maintainers of license lists may not have a DNS registered domain.  I recall we agreed that non-DNS formatted namespaces were required to register their namespaces to avoid namespace conflicts.

zvr:
As I recall, we had discussed two (independent) license namespaces:

1. the "DNS" namespace, which is described in this PR, where the owner of example.com could use identifiers like `LicenseRef-.example.com.-something` (and obliviously publish the licenses) 

2. the "reserved name" namespace, where identifiers would be of the form `LicenseRef--example-something` where the SPDX group would serve as a simply "registry" for the names (example, in this case) and the "owner" would publish the licenses via SPDX. Examples of such namespaces could be large organizations, scanning tools, etc. We had not advanced in the discussion about how to operate the registry, etc.

The former was called dash-dot and the latter dash-dash, from the characters following `LicenseRef`.

swinslow:
Hi @goneall and @zvr, thanks for your comments. I agree, in #113 we had at least originally discussed having a second format option which was a "flat" or organization-name-based namespace as well.

I had omitted it from the draft here, because to @zvr's point I don't think we had worked our a process where that would be authenticated or managed. For DNS names, it's pretty much an objective check to see -- a yes/no of whether or not the URL they're seeking to register is within the domain name from the DNS entry. For an organization name, it's more of a subjective question of, do we know that this person represents XYZ organization or company? Maybe that's acceptable, I'd just feel better if we had a process in place for overseeing this (or at least agreement on who would handle it).

@goneall, thanks for the details on your understanding of registration being optional for DNS identifiers. I guess my question then is, how people would be able to derive the actual license text -- if all they have is a DNS-based identifier, but it's unregistered so there is no link back to an SPDX document with the license text. Maybe that's something we can discuss during the joint call tomorrow.

swinslow:
cc @jlovejoy -- apologies, I meant to include you in the list of mass-cc's at the top of this pull request! 

This is my draft proposal for the "license list namespace" writeup, to implement what we'd discussed and agreed on last year. Grateful if you are able to weigh in as well, if you have input here.

chapters/appendix-X-custom-license-namespaces.md
```diff
@@ -0,0 +1,48 @@
...
+
+The SPDX specification also defines a mechanism for creators of SPDX documents to define custom license texts and identifiers. This enables SPDX documents to refer to licenses that might never be included on the SPDX License List, for a variety of reasons (whether due to limited breadth of use, failure to meet the [license inclusion principles](https://github.com/spdx/license-list-XML/blob/master/DOCS/license-inclusion-principles.md), or otherwise).
+
+These custom license identifiers are prefixed with `LicenseRef-` as defined in [Appendix IV of the SPDX 2.2 Specification](./appendix-IV-SPDX-license-expressions.md). The `LicenseRef-` identifier format was originally assumed to be used in the context of an SPDX document, so that the license text for that identifier would be specified in a corresponding [Other Licensing Information](./6-other-licensing-information-detected.md) section.
```
swinslow: Should only define LicenseRef-'s within own namespace definition document -- thanks Carmen!

zvr: it's already specified:
> the namespace maintainer creates an SPDX document defining license texts and corresponding identifiers *with the DNS name included*
so example.org cannot publish licenses of somebodyelse.org

chapters/appendix-X-custom-license-namespaces.md
```diff
@@ -0,0 +1,48 @@
...
+This appendix defines a format and mechanism for using the existing `LicenseRef-` format in a manner that enables stable identifier usage. Under this system:
+
+1. each license "namespace" consists of a standard order DNS name (such as `example.com`) that is under the control of the namespace maintainer's organization;
+2. the namespace maintainer creates an SPDX document defining license texts and corresponding identifiers with the DNS name included, as described below, and publishes it publicly at a URL under the namespace domain;
+3. the namespace and the SPDX document URL is registered with the SPDX project; and
+4. identifiers in the format described below can then be used by anyone, as long as they are defined in the SPDX document available at the registered URL.
+
+## Format
+
+For purposes of this appendix, a valid "namespace" is a standard order DNS name, such as `example.com`. Each custom license identifier under this namespace consists of the following, in order:
```
swinslow: Update to also include org name as a second format option, per discussion on tech team call 2020-02-25

chapters/appendix-X-custom-license-namespaces.md
```diff
@@ -0,0 +1,48 @@
...
+1. the prefix `LicenseRef-`
+2. the namespace, preceded by a period, and followed by a period and hyphen (e.g., `.example.com.-`)
+3. any other characters permitted by the license expression syntax (e.g., letters, digits, "-" and ".")
+
+For example, for the `example.com` namespace, the following could be valid identifiers:
+* `LicenseRef-.example.com.-XYZ-1.0`
```
swinslow: Suggested on tech team call - also recommend that "--" and "-." should not be used for LicenseRef's that aren't part of namespaces.

Important to define hierarchy of resolution

swinslow: Syntax should be both intuitive, compatible and automatable

goneall:
An alternative proposal - we could use the external document referenced licenseRef's as defined in the [License Expresions Appendix - Simple License Expressions](https://github.com/spdx/spdx-spec/blob/development/v2.2/chapters/appendix-IV-SPDX-license-expressions.md#simple-license-expressions-).  

We could use a custom syntax for the External Document Ref to denote a "registered external document" containing licenses.

swinslow:
Proposals from 2020-03-03 tech team call (thanks @zvr!):

`NamespaceRef-amazon:LicenseRef-MyLicense `

`NamespaceRef-([^:]+):LicenseRef-(.+)`

look up \2 in namespace \1

goneall:
Updated proposal from the call:

`(NamespaceRef-|^(DocumentRef-))([^:]+):LicenseRef-(.+)`

Examples:

`NamespaceRef-amazon:LicenseRef-myLicense`
`amazon:LicenseRef-myLicense`

zvr:
We can even simplify it further, deleting the `NamespaceRef-` part, going towards:
`intel:LicenseRef-MyLicense-1.0`

If the name includes a colon (':'), the part before the colon is the namespace (or something starting with DocumentRef-).

swinslow:
Could we swap it around? with `LicenseRef-` still as the prefix? So for instance:

`LicenseRef-intel:MyLicense-1.0`

That way `LicenseRef-` is still consistently the prefix for non-official license identifiers. And after parsing that, you can look for a colon (or not) to determine whether it's namespaced (or not).

goneall:
> Could we swap it around? with LicenseRef- still as the prefix?

@swinslow you raise a good point about making it easier to parse (both for computers and humans).

Since the `DocumentRef-XXX:` is already a prefix, we already have situations where we have non-listed licenses that do not start with `LicenseRef-`.

I think the cleanest approach would be to require the `NamespaceRef-XXX`, that way we could easily recognize we are dealing with namespaces.  The rules for parsing would be:

- If the license ID starts with `LicenseRef-` then it will be found in the SPDX document
- If the license ID starts with 'DocumementRef-` then it will be found in an external SPDX document defined by the external document reference section of the SPDX document
- If the license ID starts with `NamespaceRef-` then it will be found in an SPDX document defined by the external license namespace git repository
- If none of the above apply, then it is an SPDX listed license

If we don't require the `NamespaceRef-` prefix, then the rules would be:
- If the license ID starts with `LicenseRef-` then it will be found in the SPDX document
- If the license ID contains a colon ':' and starts with 'DocumementRef-` then it will be found in an external SPDX document defined by the external document reference section of the SPDX document
- If the license ID contains a colon ':' and does NOT starts with 'DocumementRef-`  then it will be found in an SPDX document defined by the external license namespace git repository
- If none of the above apply, then it is an SPDX listed license

swinslow:
Thanks @goneall. I'm in favor of the first of the two formats you proposed, using the appropriate prefix to indicate which type of reference it is.

zvr:
Actually, we could have it both ways:
* If the license ID starts with `LicenseRef-` then it will be found in the SPDX document
* If the license ID starts with `DocumementRef-` then it will be found in an external SPDX document defined by the external document reference section of the SPDX document
* If the license ID starts with `NamespaceRef-` then it will be found in an SPDX document defined in the registration of the external license namespace
* If the license ID contains a colon `:` (and does not start with `DocumementRef-`) then it will be found in an SPDX document defined in the registration of the external license namespace
* If none of the above apply, then it is a short identifier of a license in the SPDX license list

This way the "Namespace-" prefix is optional.
I envision that it will be used in SPDX documents (and the prefix may be used for other uses in a 3.x spec), but in standalone uses people might choose the shorter format.

Example in a source file:

`SPDX-License-Id: intel:LicenseRef-ISSL-202002`

zvr:
Heh, of course if we go by my comment above, there is nothing stopping us from saying that the fourth alternative is actually:

`SPDX-License-Id: intel:ISSL-202002`

since the presence of the colon signifies a non-standard namespace, which makes the need for the "LicenseRef-" prefix redundant.

To put it another way, if `:` must always be followed by `LicenseRef-`, one can substitute `:LicenseRef-` by `:` without introducing any misunderstandings.

swinslow:
I really think it needs to be crystal-clear to people about whether they're using a curated license identifier that's on the SPDX License List, or a custom license namespace.

If it's on the License List then they know that it's been vetted, it has a stable URL and a settled reference text / markup.

If it's a custom namespace then anyone using it is buying into the controller of that namespace being able to change its meaning at their whim.

I'm really concerned about an assumption that the broader community will be able to distinguish between the implications of these two very different cases, just based on the presence of a colon. I'd strongly encourage that the custom namespace route have something that makes it really explicit, _to a casual observer_, that it's not an official and vetted license / identifier. That's why I'm a fan of something that puts `NamespaceRef-` or `LicenseRef-` front and center.

(cc @jlovejoy for visibility)

goneall:
> To put it another way, if : must always be followed by LicenseRef-, one can substitute :LicenseRef- by : without introducing any misunderstandings.

This is true for the syntax within the SPDX document making the external reference.

The proposed solution for finding the actual text of the referenced licenses is to use SPDX documents.

If we use an SPDX document to document the licenses referenced by the namespace, they would need to be preceeded by `LicenseRef-` to match the current spec.  If we don't use the existing SPDX document spec, we would be free to not require the `LicenseRef-` prefix.

chapters/appendix-X-custom-license-namespaces.md
swinslow: added in 5976c1d

chapters/appendix-X-custom-license-namespaces.md
swinslow: This is still open, and subject to outcome of conversation in the main thread.

chapters/appendix-X-custom-license-namespaces.md
swinslow: I've made this a bit more explicit, by adding the following (which is referring to the document created by the namespace maintainer):

> The SPDX document should not define any licenses with identifiers other than those in its own namespace.

I think this should address the concern indicated on the call -- thanks @carmenbianca and @zvr!

swinslow:
All, I've updated the PR to add in the second, "organization name" format, and made some other conforming changes.

I believe the only remaining open issue is whether to use the identifier format that was agreed upon last year, or whether to change it to something else (such as by incorporating `NamespaceRef-` as a prefix, or using colons, etc).

My guess is that if we are taking one of those different approaches, there will need to be more extensive changes to e.g. the license expression syntax and grammar, as well as getting buy-in from the legal team on another joint legal/tech call. So I'd suggest that if we want to go that route, then we defer all of this to 3.0. I'm not sure what the timing looks like for 2.2, but if we want to add this in the near-term then I'd encourage we use the format that is described here, since (I think) it won't require those sort of further changes and discussions.

kestewart:
Based on discussions on the call today,  there is strong support for Namespace- proposal, and possibly using ":" symbol to separate.    Alexios has volunteered to write up a pull request for this in 3.0,  and Steve has volunteered to help, and coordinate with legal team.    Since this seems to be the direction,  we'll close this for 2.2 and not apply.

PR #219, opened 2020-03-23
Add JSON schema

goneall:
The JSON schema was generated from the OWL ontology document.

There is also a minor fix to the JSON example to match the generated schema.


kestewart:
@tsteenbe - can you take a look at this?

Issue #137, opened 2019-09-03
Define  SPDXREF-NONE & SPDXREF-NOASSERTION 


kestewart:
We already have SPDXREF-DOCUMENT to refer to the entire SPDX document, and it can be used in relationships. 

There are some use cases that have been flagged in the [NTIA workgroup](https://docs.google.com/document/d/1YzpLd8VtLIrAQtULZhRf6ZENqmaZ-w71wfhQdpNSpcs/edit#heading=h.f9iwzux6yxmj),  where people want to refer to the "known unknowns" in RELATIONSHIPS.     

By introducing SPDXREF-NONE we can model that an SPDX element (package/file/snippet) has no other elements connected by some relationship to it.

By introducing SPDXREF-NOASSERTION we can model that the author is not asserting where other elements are connected by relationships to it or not.  ie.  There could be some, but the author is not asserting one way or another. 

Examples:    
Relationship:  SPDXREF-SPDXRef-CompressionEng-v3.1 CONTAINS SPDXREF-NONE 
Can be used to explicitly indicate there are NO dependencies

Relationship: Relationship:  SPDXRef-Buffer-v2.2 CONTAINS SPDXREF-NOASSERTION
Can be used to explicitly indicate its not clear if there are dependencies on this package or not.

kestewart:
Review NTIA Framing,  and align with the terminology.

kestewart:
Moving this to 3.0,  and will refine further there.

kestewart:
Move this back to 2.2, as @tsteenbe has a use case with repository managers that trim dependency trees, and this mechanism would enable a way of capturing this knowledge.

goneall:
Would these only be used on the right hand side of relationships?  If so, I can see how we can make this work in the tooling, but it may be a challenge to reflect this in the UML model accurately.

Is there any possibility this would break existing documents since we are introducing reserved ID's?

swinslow:
I don't have a particular preference either way on the overall issue, but I think @goneall is right. Existing SPDX 2.1 or earlier documents could define `SPDXRef-NONE` or `SPDXRef-NOASSERTION` to point to an existing element in the document, and this would change the definition of those identifiers, so it seems that it would break compatibility.

kestewart:
Its unlikely they've used NONE and NOASSERTION,  but the point is valid.   How about we introduce a new character permitted in IDSTRING to deliminate these values.   ie. "*"
[...]

PR #221, opened 2020-03-24
Add ability to use NONE & NOASSERTION to relationships

kestewart:
In the NTIA Framing Workgroup Phase 1, there was the requirement to be able to specify the known-unknowns (see: section 2.4.1 in https://www.ntia.gov/files/ntia/publications/framingsbom_20191112.pdf)  

Based on discussion in ISSUE #137, the use of the keywords of NONE and NOASSERTION can extend to represent these concepts, and be used with relationships.   To support tooling and different formats,  these keywords are initially being limited to be on the right hand side of a relationship.  

In order to be able to express these concepts "DEPENDS_ON" relationship has been introduced, to be the mirror of the "DEPENDENCY_OF".

### DIFF
--- a/chapters/7-relationships-between-SPDX-elements.md
+++ b/chapters/7-relationships-between-SPDX-elements.md
@@ -2,7 +2,13 @@
 
 ## 7.1 Relationship <a name="7.1"></a>
 
-**7.1.1** Purpose: This field provides information about the relationship between two SPDX elements. For example, you can represent a relationship between two different Files, between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument. The relationships between two elements that are supported are:
+**7.1.1** Purpose: This field provides information about the relationship between two SPDX elements. For example, you can represent a relationship between two different Files, between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument. 
+
+In cases where there are "known unknowns", the use of the keyword "NOASSERTION" can be used on the right hand side of a relationship to indicate that the author is not asserting whether there are other SPDX elements (package/file/snippet) are connected by relationships or not.  ie. There could be some, but the author is not asserting one way or another.  
+
+Similarly, the use of the keywords "NONE" can be used to indicate that an SPDX element (package/file/snippet) has no other elements connected by some relationship to it.
+
+The relationships between two SPDX elements that are supported are:
 
 | Relationship           | Description | Example |
 |------------------------|-------------|---------|
--- a/chapters/7-relationships-between-SPDX-elements.md
+++ b/chapters/7-relationships-between-SPDX-elements.md
@@ -4,9 +4,9 @@
 
 **7.1.1** Purpose: This field provides information about the relationship between two SPDX elements. For example, you can represent a relationship between two different Files, between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument. 
 
-In cases where there are "known unknowns", the use of the keyword NOASSERTION" can be used on the right hand side of a relationship to indicate that the author is not asserting whether there are other SPDX elements (package/file/snippet) are connected by relationships or not.  ie. There could be some, but the author is not asserting one way or another.  
+In cases where there are "known unknowns", the use of the keyword `NOASSERTION` can be used on the right hand side of a relationship to indicate that the author is not asserting whether there are other SPDX elements (package/file/snippet) are connected by relationships or not.  ie. There could be some, but the author is not asserting one way or another.  
 
-Similarly, the use of the keywords "NONE" can be used to indicate that an SPDX element (package/file/snippet) has no other elements connected by some relationship to it.
+Similarly, the use of the keywords `NONE` can be used to indicate that an SPDX element (package/file/snippet) has no other elements connected by some relationship to it.
 
 The relationships between two SPDX elements that are supported are:
 
--- a/chapters/7-relationships-between-SPDX-elements.md
+++ b/chapters/7-relationships-between-SPDX-elements.md
@@ -4,7 +4,7 @@
 
 **7.1.1** Purpose: This field provides information about the relationship between two SPDX elements. For example, you can represent a relationship between two different Files, between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument. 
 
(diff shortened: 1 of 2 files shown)


chapters/7-relationships-between-SPDX-elements.md
```diff
@@ -2,16 +2,23 @@
 
 ## 7.1 Relationship <a name="7.1"></a>
 
-**7.1.1** Purpose: This field provides information about the relationship between two SPDX elements. For example, you can represent a relationship between two different Files, between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument. The relationships between two elements that are supported are:
+**7.1.1** Purpose: This field provides information about the relationship between two SPDX elements. For example, you can represent a relationship between two different Files, between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument. 
+
+In cases where there are "known unknowns", the use of the keyword `NOASSERTION` can be used on the right hand side of a relationship to indicate that the author is not asserting whether there are other SPDX elements (package/file/snippet) are connected by relationships or not.  ie. There could be some, but the author is not asserting one way or another.  
+
```
swinslow: Suggest adding the following paragraph just after the above: "The use of NOASSERTION is not mandatory for any relationship. If no relationship of a particular type is specified, then the document author is not presumed to be asserting whether or not there are relationships of that type. If some relationships of a particular type are specified, then the document author is not presumed to be asserting whether there are more possible relationships of that type."

chapters/7-relationships-between-SPDX-elements.md
```diff
@@ -2,16 +2,23 @@
...
 | DESCRIBED_BY           | Is to be used when SPDXRef-A is described by SPDXREF-Document                                         | The package ‘WildFly’ is described by SPDX document `Wildfly.spdx`. |
 | CONTAINS               | Is to be used when SPDXRef-A contains SPDXRef-B.                                                      | An ARCHIVE file `bar.tgz` contains a SOURCE file `foo.c`. |
 | CONTAINED_BY           | Is to be used when SPDXRef-A is contained by SPDXRef-B.                                               | A SOURCE file `foo.c` is contained by ARCHIVE file `bar.tgz` |
-| DEPENDENCY_MANIFEST_OF | Is to be used when SPDXRef-A is a manifest file that lists a set of dependencies for SPDXRef-B        | A file `package.json` is the dependency manifest of a package `foo`. Note that only one manifest should be used to define the same dependency graph. | 
+| DEPENDS_ON             | Is to be used when SPDXRef-A depends on SPDXRef-B. | Package A depends on the presence of package B in order to build and run |
```
swinslow: I see why it might sound better to add this. However, my worry is that because we've added a bunch of different *_DEPENDENCY_OF types (e.g., BUILD, DEV, etc), adding a new reciprocal one here would suggest that we'd also need reciprocal versions of each of those.

Can we reach the same outcome by keeping just the existing one-way relationship types, but flipping around the ordering of elements? e.g.:

`NONE DEPENDENCY_OF SPDXRef-B`

`NOASSERTION BUILD_DEPENDENCY_OF SPDXRef-B`

It might not flow quite as well as having the reciprocal versions, but I worry that the additon makes it more complex for tools to have to handle both cases of every reciprocal relationship type when we add new reciprocal ones.

chapters/7-relationships-between-SPDX-elements.md
```diff
@@ -57,14 +64,18 @@
...
 
 where "DocumentRef-"`[idstring]`":" is an optional reference to an external SPDX document as described in [section 2.6](2-document-creation-information.md#2.6)
 
 where `SPDXID` is a string containing letters, numbers, `.` and/or `-`. as described in sections (2.3, 3.2, 4.2).
 
 where `<relationship>` is one of the documented relationship types in table 7.1.1.
 
+where `NONE` can be used to explicitly indicate there are NO other relationships.
+
+where `NOASSERTION` can be used to explicitly indicate its not clear if there are relationships that may apply or not.
```
swinslow: minor nit: "its" => "it's" or "it is"

chapters/7-relationships-between-SPDX-elements.md
```diff
@@ -76,6 +87,14 @@ Examples:
...
 
     RelationshipComment: This current document is an amendment of the SPDXA document.
+    
+    Relationship: SPDXRef-CarolCompression DEPENDS_ON NONE
+    
+    RelationshipComment: The package CarolCompression can be treated as a root with no dependencies.
+    
+    Relationship: SPDXRef-BobBrowser DEPENDS_ON NOASSERTION
+    
+    RelationshipComment: The package BobBrowser may have other packages required as dependencies, but the author has insufficient information to threat this as other than unknown at this point in time.
```
swinslow: minor nit: "threat" => "treat"

chapters/7-relationships-between-SPDX-elements.md
```diff
@@ -76,6 +87,14 @@ Examples:
     Relationship: SPDXRef-DOCUMENT AMENDS DocumentRef-SPDXA:SPDXRef-DOCUMENT
 
     RelationshipComment: This current document is an amendment of the SPDXA document.
+    
+    Relationship: SPDXRef-CarolCompression DEPENDS_ON NONE
```
swinslow: For this and the next example below, see my comment above about whether we want to add a new DEPENDS_ON relationship type. If we don't, then these examples could be reversed:

`NONE DEPENDENCY_OF SPDXRef-CarolCompression`

chapters/7-relationships-between-SPDX-elements.md
kestewart: See @goneall's comment in the issue about problems if they aren't on the RHS of the relationship.     
Rather than going into all the variants of dependencies, mirroring the most generic version seemed the best compromise.

kestewart:
From call,   Gary notes that this is going to be easier to keep NONE & NOASSERTION on RHS, so we'll go that way.   We'll just stick with just adding DEPENDS_ON for now as the reciprocal, rather than all the specific dependency relationships. 

Some relationships may not make sense with NONE & NOASSERTION and this should be noted.

chapters/7-relationships-between-SPDX-elements.md
kestewart: Added the sentences to the spec.

chapters/7-relationships-between-SPDX-elements.md
kestewart: See decision from 2020/3/24 meeting.    Path of least pain is to keep NONE on RHS of expression.

kestewart:
@swinslow - can you re-review this to double check I've addressed the issues you've flagged.

PR #222, opened 2020-03-24
Address case sensitivity for license expressions

### LINKED ISSUE #63: Clarify case sensitivity of Short Form licenses - for list and tools.
This has been moved from bugzilla: https://bugs.linuxfoundation.org/show_bug.cgi?id=1327

**Kate Stewart 2015-11-19 19:00:15 UTC**
Jilayne wrote: 
in http://lists.spdx.org/pipermail/spdx-tech/2015-November/002905.html

- in http://wiki.spdx.org/view/Technical_Team/Minutes/2014-09-16#
Case_sensitivity_for_license_information - the tech team discussed this on 16 Sept 2014, note saying “License ID’s case sensitive”

- and then the legal team discussed it - http://wiki.spdx.org/view/Legal_Team/Minutes/2014-09-18 - and concluded:
        • Mark raised issue of whether SPDX License List short identifiers and (new) license expression operators should be case sensitive with the Tech Team and discussed further here: decided that for purposes of spec, in terms of a legitimate value, both could be case insensitive (but best practice would be to display with precise capitalization). Mark to go back to tech team with this decision.

So… looks like maybe we didn’t really capture this elsewhere?  In any case, I don’t see a reason to have them be case sensitive in terms of matching (for tools), but have them display with the upper/lower case as they are shown in the SPDX License List - it’s easier for humans to read/spot :)

**Kate Stewart 2015-11-19 19:01:50 UTC**
I'll add it to the 2.1 version of the spec.   Also consider adding this as an appendum/erratta for 2.0.

**Kate Stewart 2015-12-22 18:13:49 UTC**
Discussed on 12/22 - no concerns,  going forward with documenting.

**Bill Schineller 2016-05-10 17:53:56 UTC**
didn't jump out at me where / if we made edit yet to SPDX 2.2
todo

**Kate Stewart 2016-05-17 17:01:29 UTC**
Have proposed edit to 6.1, and Appendix I.   Lets review.

**Kate Stewart 2016-05-17 17:14:40 UTC**
In discussion, some concern about other tools and matching in future. 

Circling back this discussion to include Mark Gisi.

**Bill Schineller 2016-05-17 17:15:33 UTC**
fwiw:
from http://lists.w3.org/Archives/Public/www-rdf-interest/2003Aug/0002.html
RDF is case-sensitive. From the last call Concepts working draft:

     Two RDF URI references are equal if and only if they compare as
     equal, character by character, as Unicode strings.

     -- http://www.w3.org/TR/rdf-concepts/#section-Graph-URIref

An upper-case 'A' and a lower-case 'a' are different Unicode characters.

**Bill Schineller 2016-05-24 17:13:32 UTC**
Kate / Jilayne agreed to leave the Spec language as-is for 2.1

as-is means 'it IS case-sensitive'

leaving ticket open with Version 'unspecified'  in case we want to revisit in the future.

We were reluctant to make case-insensitive now for 2.1 without understanding the impacts case might have on URIs  (website, other tools, RDF graphs, ...)

swinslow:
This commit attempts to reflect the outcome of the discussion at
https://github.com/spdx/spdx-spec/issues/63 regarding whether
license expression operators and identifiers should be matched in
a case-sensitive manner.

Specifically it attempts to reflect the comment at
https://github.com/spdx/spdx-spec/issues/63#issuecomment-394805189
regarding the outcome of the tech team discussion on 2018-06-05.

Closes #63 

@goneall -- would greatly appreciate your review, as I'm not certain whether I'm accurately capturing the intent of the comment linked above. Thanks!

### DIFF
--- a/chapters/appendix-IV-SPDX-license-expressions.md
+++ b/chapters/appendix-IV-SPDX-license-expressions.md
@@ -41,6 +41,14 @@ A valid `<license-expression>` string consists of either:
 
 There MUST NOT be whitespace between a license-id and any following `+`. This supports easy parsing and backwards compatibility. There MUST be whitespace on either side of the operator "WITH". There MUST be whitespace and/or parentheses on either side of the operators `AND` and `OR`.
 
+## Case sensitivity <a name="case-sensitivity"></a>
+
+License expression operators (`AND`, `OR` and `WITH`) should be matched in a *case-sensitive* manner.
+
+License identifiers used in SPDX documents or source code files should be matched in a *case-insensitive* manner. In other words, `MIT`, `Mit` and `mIt` should all be treated as the same identifier and referring to the same license.
+
+However, please be aware that it is often important to match with the case of the canonical license identifier on the [SPDX License List](https://spdx.org/licenses). This is because the canonical identifier's case is used in the URL of the license's entry on the List, and because the canonical identifier is translated to a URI in RDF documents.
+
 ## Simple License Expressions <a name="simple-expr"></a>
 
 A simple `<license-expression>` is composed one of the following:
--- a/chapters/appendix-IV-SPDX-license-expressions.md
+++ b/chapters/appendix-IV-SPDX-license-expressions.md
@@ -45,9 +45,9 @@ There MUST NOT be whitespace between a license-id and any following `+`. This su
 
 License expression operators (`AND`, `OR` and `WITH`) should be matched in a *case-sensitive* manner.
 
-License identifiers used in SPDX documents or source code files should be matched in a *case-insensitive* manner. In other words, `MIT`, `Mit` and `mIt` should all be treated as the same identifier and referring to the same license.
+License identifiers (including license exception identifiers) used in SPDX documents or source code files should be matched in a *case-insensitive* manner. In other words, `MIT`, `Mit` and `mIt` should all be treated as the same identifier and referring to the same license.
 
-However, please be aware that it is often important to match with the case of the canonical license identifier on the [SPDX License List](https://spdx.org/licenses). This is because the canonical identifier's case is used in the URL of the license's entry on the List, and because the canonical identifier is translated to a URI in RDF documents.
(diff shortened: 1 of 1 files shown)


chapters/appendix-IV-SPDX-license-expressions.md
```diff
@@ -41,6 +41,14 @@ A valid `<license-expression>` string consists of either:
 
 There MUST NOT be whitespace between a license-id and any following `+`. This supports easy parsing and backwards compatibility. There MUST be whitespace on either side of the operator "WITH". There MUST be whitespace and/or parentheses on either side of the operators `AND` and `OR`.
 
+## Case sensitivity <a name="case-sensitivity"></a>
+
+License expression operators (`AND`, `OR` and `WITH`) should be matched in a *case-sensitive* manner.
+
+License identifiers used in SPDX documents or source code files should be matched in a *case-insensitive* manner. In other words, `MIT`, `Mit` and `mIt` should all be treated as the same identifier and referring to the same license.
```
goneall: I checked and the SPDX tools look up the license by using the case sensitive URI - so implementing a case insensitive license ID would be somewhat substantial change to the SPDX tools (and any other library that uses the URL based on the ID - which could be quite a few tools since this has been the reccomended approach.

Recommend we change this to case sensitive ID's.

swinslow: Thanks @goneall -- I'm fine with changing it, but wonder if we should discuss on the tech team call today, since I'd understood the prior consensus at https://github.com/spdx/spdx-spec/issues/63#issuecomment-394805189 to be that they would be treated as case-insensitive.

goneall: I just found a PR from Feb 2018 when this issue was first discussed: https://github.com/spdx/tools/pull/153/files  This PR makes the entire expression (including the and's and or's) case insensitive.

Based on this and the prior consensus, we can leave the license ID's case insensitive as proposed.  I'll merge in the PR to the tools.

It would be worth raising on the call since other tools may not be implementing the case sensitivity - just want to give a heads up to other tool writers they should check their parser implementations.

kestewart:
Thomas asked about license exception identifiers,   Steve added clarity based.    
Alexios suggested just saying canonical identifier, to remove ambiguities as well.

swinslow:
closed in favor of #228 due to merge conflicts here

PR #224, opened 2020-03-31
Remove multi-line license exprs and mandatory parens

### LINKED ISSUE #44: Clarify the use of newlines in license expressions in Appendix V
The current license expression syntax states that whitespace must be used between elements of a compound expression.  However, it does not explicitly state if a newline, CR, or LF is included in the definition of whitespace.

Suggest adding a definition of white space which include new line.

### LINKED ISSUE #45: Clarify use of newlines for license expressions used in source files
In Appendix V, it is not clear if multiple lines are to be used for compound set of licenses.  Suggest changing the following statement:

The SPDX License Identifier syntax may consist of  a single license (represented by a short identifier from the SPDX license list) or a compound set of licenses (represented by joining together multiple licenses using the license expression syntax).

to:

The SPDX License Identifier syntax may consist of  a single license (represented by a short identifier from the SPDX license list) or a compound set of licenses (represented by joining together multiple licenses using the license expression syntax which is enclosed in parenthesis and may span multiple lines).

### LINKED ISSUE #52: Make parens optional in license expressions
There are no good reasons for `(` parens  `)` to be mandatory for most compound expressions.
The only cases I can fathom would be:
1. multiline expressions, and IMHO these should be banned
2. rare cases where the operator precedence is not enough
3. to make expressions more easy to read

Therefore, they should be optional and best left to the cases where they are needed only. This would make things much simpler. 
Tool wise, any decent boolean expression handler does not care much about un-needed parens. If they are that do, they should be fixed.

swinslow:
This changes appendices IV and V to prohibit multiple lines (e.g.,
line breaks) in a single license expression. This is applicable
for both tag-value SPDX documents as well as short-form license
identifier expressions in source code.

Because multiple lines are no longer permitted, this also makes
the outer set of parentheses optional, rather than mandatory, for
license expressions with more than one license identifier.

Fixes #44
Fixes #45
Fixes #52

### DIFF
--- a/chapters/appendix-IV-SPDX-license-expressions.md
+++ b/chapters/appendix-IV-SPDX-license-expressions.md
@@ -41,6 +41,8 @@ A valid `<license-expression>` string consists of either:
 
 There MUST NOT be whitespace between a license-id and any following `+`. This supports easy parsing and backwards compatibility. There MUST be whitespace on either side of the operator "WITH". There MUST be whitespace and/or parentheses on either side of the operators `AND` and `OR`.
 
+In the `tag:value` format, a license expression MUST be on a single line, and MUST NOT include a line break in the middle of the expression.
+
 ## Simple License Expressions <a name="simple-expr"></a>
 
 A simple `<license-expression>` is composed one of the following:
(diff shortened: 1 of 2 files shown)


kestewart:
Discussed on call,  no other concerns raised,  merging.

PR #226, opened 2020-04-07
Adding SoftwareHeritage external references

### LINKED ISSUE #97: Add External Reference to "Identity of Digital Object"
see discussion in:
https://annex.softwareheritage.org/public/talks/2018/2018-09-25-iPres2018.pdf

Would like to be able to refer to artifacts contained in Software Heritage.

zvr:
Added `swh` type in a new `PERSISTENT-ID` category of external references in appendix VI (and in chapter 3 where it is referenced).

Closes #97

### DIFF
--- a/chapters/appendix-VI-external-repository-identifiers.md
+++ b/chapters/appendix-VI-external-repository-identifiers.md
@@ -1,12 +1,10 @@
 # Appendix VI: External Repository Identifiers
 
-***
-When `<category>` = `SECURITY`:
-***
+## When `<category>` = `SECURITY`:
 
-## `<type>` cpe22Type <a name="cpe22"></a>
+### `<type>` cpe22Type <a name="cpe22"></a>
 
-### `<locator>` Information
+#### `<locator>` Information
 
 Locator Format:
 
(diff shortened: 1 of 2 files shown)


chapters/appendix-VI-external-repository-identifiers.md
```diff
@@ -133,12 +130,51 @@ External Reference Site: [https://github.com/package-url/purl-spec](https://gith
...
+These point to objects present in the Software Heritage archive by the means of persistent identifiers
+that are guaranteed to remain stable (persistent) over time.
+Their syntax is described below.
+Note that they are identifiers and not URLs.
+
+A persistent identifier can point to any software artifact (or “object”)
+available in the Software Heritage archive.
+Objects come in different types, and most notably:
+
+* contents
```
goneall, line 149: Should we add the short form from the grammer in the description?  either "cnt - contents" or "contents (cnt)"?

zvr: I've put them in the right order between description and grammar. ;-)
I tried to not overload the document with details, although providing more info than the minimal CPE types. I think linking to external documentation covers it.

kestewart: Agree with @zvr here, its following the pattern with the others, so lets go forward with this as is right now, and revisit how its described in 3.0.

Issue #159, opened 2019-12-05
Re-evaluating CC0-1.0 as DataLicense for SPDX 3.0


swinslow:
The SPDX 3.0 discussions have included questions being raised about whether CC0-1.0 should be retained as the mandatory `DataLicense` field for SPDX documents.

The SPDX legal team is gathering details about the historical rationales for why the CC0-1.0 license was initially chosen, and how that is seen as operating in SPDX documents. (Much of these rationales are currently present in the existing spec and on the SPDX wiki.)

For those folks who have asked to make a change to the current CC0-1.0 DataLicense, in order to help evaluate this request, I'd ask that they add comments to this issue explaining specifically:

1. why they believe that CC0-1.0 is not satisfactory for their anticipated use cases for 3.0, in light of the existing rationales in the spec; and
2. what specific alternative they would want to see (e.g. `DataLicense` as a field that can take any license expression, or remove the field altogether, etc.)

kestewart:
@iamwillbar, @pombredanne - if you have examples,  can you share your use cases here, so we can justify doing this.    Thanks!

jlovejoy:
I think I posted this on the mailing list but putting link here too, to make sure folks have the background. We had a write-up explaining the rationale for CC-0 - https://wiki.spdx.org/images/SPDX-TR-2014-1.v1.1.pdf

Also, don't forget the preamble (which should be somewhere besides this...) https://wiki.spdx.org/view/Legal_Team/Decisions/SPDX_Metadata_License:_Preamble_and_CC0_1.0_Universal

swinslow:
I've raised this on a few prior tech team calls requesting input, and haven't seen any feedback in response to the questions raised above to the folks seeking a license change. So I am inclined to close this issue and stick with CC0-1.0 as the document DataLicense.

swinslow:
No responses, so closing issue.

MarkAtwood:
I would like to reopen this issue.  Amazon has severe resevations about being required to tag the SBOMs of our internal services and delivered products as CC0, even if there is also an NDA in place.  We especially don't want to have "you put a CC0 on it" when someone else publishes something that was provided to them by someone breaking their NDA.  The other SBOM standards do not require a CC0 or other license tag.

PR #257, opened 2020-04-10
dataLicense can be only one license (CC0-1.0)

davaya:
According to the description: "By using the SPDX specification ..., you hereby agree that any copyright rights ... shall be subject to the terms of the Creative Commons CC0 1.0 Universal license."

The value of Document Information / dataLicense thus cannot be any arbitrary SPDXExpression, it must be only the expression "CC0-1.0" as shown in the examples.  For consistency with the Description, the schema should specify that any other value of dataLicense is invalid.

(Note: the name of the constant type CC0-LICENSEID is arbitrary as long as it is a string matching the regex ^CC0-1\\.0$).

### DIFF
--- a/chapters/2-base-profile.md
+++ b/chapters/2-base-profile.md
@@ -79,7 +79,7 @@ Information about the SPDX document itself.
                 fontsize = 10
         ]
 
-        DocumentMetadata [label = "{DocumentMetadata|+ spdxVersion : string\l+ dataLicense : SPDXExpression\l+ SPDXID : string\l+ documentNamespace : string\l+ documentName : string\l+ created : DateTime\l+ creators : Identity[1..*]\l+ externalDocumentReferences : ExternalDocumentReference[0..*]\l|}"]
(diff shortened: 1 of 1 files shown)


swinslow:
Thanks @davaya -- I'd note also that there have been discussions about whether this should be changed for 3.0 so that other licenses besides CC0-1.0 are permitted. If you (or others) are interested, I'd encourage taking a look at issue #159 and weighing in on the two open questions there. Thanks!

davaya:
Hi Steve, my perspective is as a data scientist, so I have no position with respect to issue #159.  Organizing SPDX into legal and technical teams is an excellent approach, and makes it easy to articulate that whatever the legal team decides, the data format must be able to accommodate it.

I wasn't aware that the issue was under discussion, so thanks for pointing that out.  If a decision is made to allow any license, or a defined subset of licenses, or only a single license, then we'll be able to implement it using appropriate data.

Issue #270, opened 2020-04-17
JSON is not a supported file format


davaya:
Section 1.7.6 says: Current supported formats include YAML, JSON, RDF, flat text, .xls spreadsheets.

But JSON does not appear to be supported - it is mentioned only three times: once in section 7 as example "package.json", once in Appendix I as the JSON License, and once in Appendix IV as "package.json".   Is there information separate from the rendered 2.2 spec that defines the contents of a file in .spdx.json format?

goneall:
@davaya There is a JSON schema at https://github.com/spdx/spdx-spec/blob/development/v2.2/schemas/spdx-schema.json.

I'm thinking we should add a pointer to the schema in the document.

@kestewart What do you think?  Should we add it to section 1.7.6?

kestewart:
@goneall - am fine with adding references to the appropriate schemas in 1.7.6.
You willing to submit the PR?

PR #276, opened 2020-04-21
Add reference to the JSON schema.  Resolves issue #270

goneall:
Note that the hyperlink in the doc will not work until the development branch is merged into master.

### DIFF
--- a/chapters/1-rationale.md
+++ b/chapters/1-rationale.md
@@ -73,7 +73,7 @@ In an SPDX document, Relationship elements can be used to indicate relationships
 **1.7.6** Multiple file formats can be used to represent the information being exchanged.   Current supported formats include:
 * YAML 1.2  see: https://yaml.org/spec/1.2/spec.html
 * JavaScript Object Notation (JSON) see: ECMA-404(https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)
-    The JSON Schema for SPDX can be found in the [SPDX Spec Git Repository Schema directory](https://github.com/spdx/spdx-spec/schemas/spdx-schema.json)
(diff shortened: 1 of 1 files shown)


chapters/1-rationale.md
```diff
@@ -74,6 +74,7 @@ In an SPDX document, Relationship elements can be used to indicate relationships
 * YAML 1.2  see: https://yaml.org/spec/1.2/spec.html
 * JavaScript Object Notation (JSON) see: ECMA-404(https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)
 * Resource Description Framework (RDF also referred to as RDF/XML) see: https://www.w3.org/TR/rdf-syntax-grammar/
+    The JSON Schema for SPDX can be found in the [SPDX Spec Git Repository Schema directory](https://github.com/spdx/spdx-spec/schemas/spdx-schema.json)
```
swinslow, line 77: Hi @goneall, content-wise this looks good, but putting it here looks like it's getting folded into the RDF bullet point when it renders. Can this maybe be moved one line up and made into a sub-bullet?

carmenbianca:
@kestewart The URL in this PR leads to a 404. I assume it needs to lead to https://github.com/spdx/spdx-spec/blob/development/v2.2/schemas/spdx-schema.json. `blob/devepment/v2.2` should probably be replaced by `blob/master`, although `master` doesn't currently have this file yet.

kestewart:
Thanks for flagging @carmenbianca.   @goneall -  where do you want it to reside long terms so we can clean this up?

goneall:
This should be resolved once this is merged into master.  If we pointed it to 2.2, it would create a 404 after the development branch is deleted.

goneall:
@carmenbianca @kestewart I just checked and indeed the link was still broken after merging to master.  I just created a PR to fix.

Thanks @carmenbianca for pointing this out.

Issue #280, opened 2020-04-22
Proposal - make the Attribution Text be a property of the SpdxItem


goneall:
Currently in the 2.2 spec, Attribution Text is a property for the package and the file.  The definition of an SPDX item is anything that is copyrightable and can have a license.  It seems that it would be consistent and logical to associate the attribution text with all copyrightable items since they could carry a license which requires attribution.

The practical implication is we would add Attribution Text as a property to Snippets.  Although this may not be used much (we all agree Snippets are not used nearly as widely as Files and Packages), it would be consistent and I could argue for some use cases where it may be helpful.

Another consistent approach would be to remove this as a property from the File making the argument that Packages are the distributed element and is the only class that needs the attribution text.

Having this as a direct property to Package and File but not Snippet seems inconsistent IMO.

goneall:
@kestewart If you agree with the proposal, I can write up a PR with the spec changes.

kestewart:
@goneall - I agree with the proposal.   We should make this consistent.

PR #281, opened 2020-04-22
Update Ontology and JSON Schema to the SPDX 2.2 release

goneall:
The OWL, JSON Context and JSON Schema files should now all be consistent with the candidate release for 2.2 with one exception.

The model for the ontology attaches the new property `attributionText` to the SpdxItem rather than individually to the SpdxPackage and SpdxFile.  

I created a proposal to address the inconsistency with issue #280 

Below is a summary of the changes made:

* Additional checksums added:
    * checksumAlgorithm_sha224 (SHA224)
	* checksumAlgorithm_sha384 (SHA384)
	* checksumAlgorithm_sha512 (SHA512)
	* checksumAlgorithm_md2 (MD2)
	* checksumAlgorithm_md4 (MD4)
	* checksumAlgorithm_md6 (MD6)
* Added external reference category referenceCategory_persistenId
* Added property attributionText
    * Property added to SpdxItem which is now inconsistent with the spec.  Added Issue #28 to track the spec inconsistency
* Added new relationships
    * relationshipType_buildDependencyOf (BUILD_DEPENDENCY_OF)
    * relationshipType_dependencyManifestOf (DEPENDENCY_MANIFEST_OF)
    * relationshipType_dependencyOf (DEPENDENCY_OF)
    * relationshipType_dependsOn (DEPENDS_ON)
    * relationshipType_devDependencyOf (DEV_DEPENDENCY_OF)
    * relationshipType_devToolOf (DEV_TOOL_OF)
    * relationshipType_exampleOf (EXAMPLE_OF)
    * relationshipType_optionalDependencyOf (OPTIONAL_DEPENDENCY_OF)
    * relationshipType_providedDependencyOf (PROVIDED_DEPENDENCY_OF)
    * relationshipType_runtimeDependencyOf (RUNTIME_DEPENDENCY_OF)
    * relationshipType_testDependencyOf (TEST_DEPENDENCY_OF)
    * relationshipType_testOf (TEST_OF)
    * relationshipType_testToolOf (TEST_TOOL_OF)
    * relationshipType_none (NONE)
    * relationshipType_noAssertion (NOASSERTION)

Note: No changes were necessary to support puurl.


zvr:
Github does not make it easy to review such files, so I'll do it offline in Protege and write back here...

goneall:
@zvr Thanks for the review!  I'll commit the changes once you're done reviewing.

> has exactly 1 Checksum; should be min 1 (like File has) 

I think exactly 1 is correct - but we can discuss in the other issue

> comment says algorithm is defined in section 4.7 of the spec; it's 3.9.4 now

Will be fixed in the next commit

> attributionText is owl:rational; should be xsd:string

Will be fixed in the next commit

> owl:Thing

Removed all three references - will be fixed in the next commit

zvr:
Comparing OWL with the spec sections:

* SpdxDocument has `spdxVersion`; [2.1](https://spdx.github.io/spdx-spec/v2-draft/2-document-creation-information/#21-spdx-version) has `specVersion`
* homepage in a Package is `xsd:anyURI`; [3.11.6](https://spdx.github.io/spdx-spec/v2-draft/3-package-information/#311-package-home-page) says `doap:homepage` 
* licenseInfoFromFiles in Package has `min 1`; [3.14.3](https://spdx.github.io/spdx-spec/v2-draft/3-package-information/#314-all-licenses-information-from-files) says it must be omitted if filesAnalyzed is false
* ReferenceType in ExternalRef has `contextualExample`, `documentation`, and `externalReferenceSite`; [3.21.6](https://spdx.github.io/spdx-spec/v2-draft/3-package-information/#321-external-reference) mentions none of these
* licenseInfoInSnippet in Snippet has `min 1`; [5.6.3](https://spdx.github.io/spdx-spec/v2-draft/5-snippet-information/#56-license-information-in-snippet) says it's optional
* reviewer in Review has `exactly 1`; [9.1.3](https://spdx.github.io/spdx-spec/v2-draft/9-review-information-deprecated/#91-reviewer-deprecated) says it's optional
* comment in Review has `exactly 1`; 9.3.3 says it's optional

sorry, @goneall for the long list...

goneall:
@zvr I updated the schema and context files base on your review.  If you could just double check that the updates resolves the identified issues.

Below are the identified issues and status as of this update:

> SpdxDocument has spdxVersion; 2.1 has specVersion

Changed this back to specVersion to be compatible with 2.1.  In 3.0, we can change it to spdxVersion to be compatible with the JSON and YAML property names

> homepage in a Package is xsd:anyURI; 3.11.6 says doap:homepage

The spec is referring to the property URI which is doap:homepage (rather than spdx:homepage).  The type is not specified in the spec document, but a URI should be correct.  This also hasn't changed from the 1.1 spec.

> licenseInfoFromFiles in Package has min 1; 3.14.3 says it must be omitted if filesAnalyzed is false

Changed the cardinality to min 0

> ReferenceType in ExternalRef has contextualExample, documentation, and externalReferenceSite; 3.21.6 mentions none of these

This is a known issue - these are defined in an appendix in the spec.  No change from 2.0

> licenseInfoInSnippet in Snippet has min 1; 5.6.3 says it's optional

Changed to min 0

> reviewer in Review has exactly 1; 9.1.3 says it's optional

Changed to max 1

comment in Review has exactly 1; 9.3.3 says it's optional

Changed to max 1

zvr:
I agree with everything (and I completely missed the homepage namespace), except:

>> ReferenceType in ExternalRef has contextualExample, documentation, and externalReferenceSite; 3.21.6 mentions none of these

> This is a known issue - these are defined in an appendix in the spec. No change from 2.0

I did not find anything like that in any appendix.

OK, we know that external references have 3 parts: category, type and locator. These are enumerated in appendix VI. Where is the `ReferenceType` part further broken down?

If there are such elements in `ReferenceType`, I am concerned I need to provide something for the new `swh` we introduced for persistent identifiers.

zvr:
Of course, this should not be blocking for the 2.2 release; I understand that all this comes from the past...
But we should fix it in 2.2.1 and not have it in ISO

goneall:
Regarding the `ReferenceType`

> I did not find anything like that in any appendix.

In Appendix VI, most reference type has 4 categories of information "Locator Format", "Contextual Example", "External Reference Site" and "Documentation".  The last 3 correspond to the properties `contextualExample`, `externalReferenceSite`, and `documentation`.  

In looking at the persistent identifiers, all of the information is there but the headings are a little different.  The external reference site would the the same URI as the documentation and example would be renamed contextual example.

In practice, I don't think it matters about the format.  I don't even know if the structure for `ReferenceType` is ever used.

I recall when we added this, there was a proposal we create a website similar to the license list where we could programatically access this structure and use it to connect the graphs between different resources for the same component which would enable some pretty powerful semantic reasoning.  We never created such a website and the champion for this proposal, Bill from Black Duck, moved on from the group.

goneall:
> Of course, this should not be blocking for the 2.2 release; I understand that all this comes from the past...
But we should fix it in 2.2.1 and not have it in ISO

I agree on both points.  Removing it now would break the SPDX tools, but we could resolve this in the 2.2.1 timeframe.

@zvr Let me know if there are any other changes you think are needed for 2.2, otherwise we'll go ahead and merge.

zvr:
Go head and merge close this one for now.
We have work to do, but it will be in the next iteration.

Issue #285, opened 2020-04-27
Rendering issues


zvr:
I've gone through all pages of the spec and marked some issues:

1. list and links in 1.7.6 not rendered
   * https://spdx.github.io/spdx-spec/v2-draft/1-rationale/#17-format-requirements
1. example tag-value in 4.16.5 is in red
   * https://spdx.github.io/spdx-spec/v2-draft/4-file-information/#416-file-dependencies-deprecated
1. example RDF in 5.3.6 is in red
   * https://spdx.github.io/spdx-spec/v2-draft/5-snippet-information/#53-snippet-byte-range
1. classes in 5.3.6 are not in `code` (cf. 5.4.6)
   * https://spdx.github.io/spdx-spec/v2-draft/5-snippet-information/#53-snippet-byte-range
1. reference to "Appendix IV" in 5.5.1 is not a link
   * https://spdx.github.io/spdx-spec/v2-draft/5-snippet-information/#55-snippet-concluded-license
1. should the list of all relationship types be sorted by name?
   * https://spdx.github.io/spdx-spec/v2-draft/7-relationships-between-SPDX-elements/#71-relationship
1. example RDF in 7.1.6 is incorrectly indented
   * https://spdx.github.io/spdx-spec/v2-draft/7-relationships-between-SPDX-elements/#71-relationship
1. property and class in 8.2.6 not in `code`
   * https://spdx.github.io/spdx-spec/v2-draft/8-annotations/#82-annotation-date
1. sections 9.2 and 9.3 do not appear in table of contents
   * https://spdx.github.io/spdx-spec/v2-draft/9-review-information-deprecated/
1. no sections of Appendix II appear in table of contents
   * https://spdx.github.io/spdx-spec/v2-draft/appendix-II-license-matching-guidelines-and-templates/
1. numbering in CC-BY-3.0 license text in appendix restarts from 1 in different places
   * https://spdx.github.io/spdx-spec/v2-draft/appendix-VII-creative-commons-attribution-license-3.0-unported/

kestewart:
@zvr,  can you re-check the renderings of the lists, when you get on line?   I applied William's fix this evening, so hopefully the list issues are sorted.

kestewart:
@tsteenbe - can you look into why some parts are rendering in "red",   no clue from looking at the source why this is so.

tsteenbe:
[...]

PR #295, opened 2020-04-28
Appendix VII: Fix rendering section headers

tsteenbe:
Fix Markdown incorrect thinking the CC-BY-3.0 license text
is on big list.

Signed-off-by: Thomas Steenbergen <thomas.steenbergen@here.com>


tsteenbe:
@zvr  @kestewart  @swinslow  To fix the markdown issue raised by @zvr in point 11 of https://github.com/spdx/spdx-spec/issues/285 I had to make minor change the text/lay-out of the CC-BY-3.0 license. Hope this is acceptable for you.

swinslow:
Agreed with @kestewart -- OK with the bullet point numbers either being all bolded or all unbolded, but should be consistent. It looks like 1 is the only one where the number itself is bolded, so might be easiest to change it to align with the others.

tsteenbe:
Changed bolding of point 1 so all the bullet point numbering is now consistent.

PR #296, opened 2020-04-29
couple of fixes for rendering

zvr:
related to #285

- missed another list in 1.7.5 (point 1 in list)
- missed a closing parenthesis in 5.3.6 (point 4 in list)


swinslow:
Hi @zvr, the PR I submitted at #299 included the line break for 1.7.5, along with a few other nits. I've also now added the 5.3.6 fix you caught. So I think #299 will handle both of these. Thanks!

kestewart:
Closing this, as I believe they've been handled by Steve's just now.   Please reopen if I'm mis reading something.

PR #297, opened 2020-04-29
Fixes in XML definitions in the spec text

### LINKED ISSUE #284: Errors in XML definitions in the spec text
1. 2.3.6, example: 0305E82C33... --> 0305E82C33123#SPDXRef-Document
1. 2.4.6: Document --> spdx:SpdxDocument
1. 2.6.6: spdx:Document --> spdx:SpdxDocument
1. 3.2.6, first example: rdf:ID="SPDXRef-1" --> rdf:about="#SPDXRef-1"
1. 3.2.6, second example: 28A9B0349A1B --> 28A9B0349A1B#SPDXRef-1
1. 3.16.4: In RDF, it is delimited by \<licenseComment\>. --> 
1. 3.22.4: In RDF, it is delimited by \<ExternalRefComment\>. --> 
1. 3.23.6: attributionText --> spdx:attributionText
1. 4.2.6, first example: rdf:ID="SPDXRef-1" --> rdf:about="#SPDXRef-1"
1. 4.14.6: fileContributor --> spdx:fileContributor
1. 5.1.6, first example: rdf:ID="SPDXRef-1" --> rdf:about="#SPDXRef-1"
1. 5.1.6, second example: 28A9B0349... --> 28A9B0349A1B#SPDXRef-1
1. 5.3.6: spdx:byteRange --> spdx:range
1. 5.4.6: spdx:byteRange --> spdx:range
1. 5.10.6: spdx:snippetName --> spdx:name
1. 6.3.6: spdx:licenseName --> spdx:name
1. 7.1.6: relationship --> spdx:relationship
1. 7.1.6 (and in example): `SpdxElement` is never defined. Use Document, Package, File, or Snippet.
1. 8.3.6: annotationType --> spdx:annotationType
1. 8.3.6, example: spdx:annotationType --> annotationType
1. 8.4.6 (and in example): SpdxElement is never defined. Use Document, Package, File, or Snippet.
1. IV.5, end of section (License Exceptions): the attributes should be in `code` in one words and with lower first letter
   * seeAlso: (optional) --> (optional, one or more)
   * example: --> (optional)

@goneall let me know whether you agree with these...

zvr:
Fixes #284

### DIFF
--- a/chapters/8-annotations.md
+++ b/chapters/8-annotations.md
@@ -91,13 +91,13 @@ Example:
 AnnotationType: REVIEW
 ```
 
-**8.3.6** RDF: property `annotationType` in class `spdx:Annotation`
+**8.3.6** RDF: property `spdx:annotationType` in class `spdx:Annotation`
 
 Example:
 
 ```text
 <Annotation>
-    <spdx:annotationType rdf:resource="http://spdx.org/rdf/terms#annotationType_other"/>
+    <annotationType rdf:resource="http://spdx.org/rdf/terms#annotationType_other"/>
 </Annotation>
 ```
 
@@ -132,16 +132,16 @@ SPDXREF: DocumentRef-spdx-tool-1.2:SPDXRef-5
 
 **8.4.6** RDF:
 
-For RDF, the annotations are a property of the SPDX element it is annotationg.
+For RDF, the annotations are a property of the SPDX Document, Package, File, or Snippet they are annotationg.
 
 ```text
-<SpdxElement rdf:about="#SPDXRef-45">
+<File rdf:about="#SPDXRef-45">
     <annotation>
         <Annotation>
             ...
         </Annotation>
     </annotation>
-</SpdxElement rdf:about="#SPDXRef-45">
+</File>
 ```
 
 ## 8.5 Annotation Comment <a name="8.5"></a>
--- a/chapters/8-annotations.md
+++ b/chapters/8-annotations.md
@@ -132,7 +132,7 @@ SPDXREF: DocumentRef-spdx-tool-1.2:SPDXRef-5
 
 **8.4.6** RDF:
 
-For RDF, the annotations are a property of the SPDX Document, Package, File, or Snippet they are annotationg.
+For RDF, the annotations are a property of the SPDX Document, Package, File, or Snippet they are annotating.
 
 ```text
 <File rdf:about="#SPDXRef-45">
--- a/chapters/8-annotations.md
+++ b/chapters/8-annotations.md
@@ -69,7 +69,7 @@ Example:
 
 ```text
 </Annotation>
-    <annotationDate> 2010-01-29T18:30:22Z </annotation Date>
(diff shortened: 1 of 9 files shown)


chapters/8-annotations.md
```diff
@@ -132,16 +132,16 @@ SPDXREF: DocumentRef-spdx-tool-1.2:SPDXRef-5
 
 **8.4.6** RDF:
 
-For RDF, the annotations are a property of the SPDX element it is annotationg.
+For RDF, the annotations are a property of the SPDX Document, Package, File, or Snippet they are annotationg.
```
swinslow, line 135: minor nit (in original, not from your edit) -- "annotationg" => "annotating"

zvr: done

zvr:
It seems that @goneall or @swinslow have to mark the "requested changes" as satisfied...

kestewart:
@zvr - please review Thomas' changes.

goneall:
@zvr Thomas just updated the PR with commit a28a965 adding `spdx:` to the examples.  I did a review and it looks good to me - if you don't mind doing one more review.  Thanks.

zvr:
I remember when we removed all the `spdx:` namespace  from all the code examples in the spec, to present a simpler view -- since realistically, the XML documents for SPDX will have it as a default namespace.
Why do we want to re-introduce it to everything?

kestewart:
@zvr - from discussion it seems new people were confused on how to use the rdf, and so there was resistance to adopting it.  @tsteenbe felt this would be easier for others to understand this way, and @goneall was ok with trying it this way.

goneall:
> I remember when we removed all the spdx: namespace from all the code examples in the spec, to present a simpler view -- since realistically, the XML documents for SPDX will have it as a default namespace.

@zvr You have a better memory than I do.  I forgot about removing them before.

I'm OK either way.  @tsteenbe pointed out that with the recent changes of adding the `spdx:` namespace the examples looked inconsistent.  I thought he made a good point since the properties used a namespace right before the examples.  

@zvr Let me know if you feel it would be more readable without the namespaces or if you are OK with adding them back in. It would be easy to revert back to the non-spdx namespaces since they are all in one commit.

zvr:
I believe it's visual clutter and the spec will seem much more complicated; it's simpler to see things like `Package` that includes something called `filesAnalyzed` rather than `spdx:Package` and `spdx:filesAnalyzed`.

I added the namespace explicitly in all the definitions (i.e., `spdx:filesAnalyzed` inside `spdx:Package`) in this PR for this reason..

I don't mind the namespace in the standalone example files, but the spec has snippets (ha!) of the XML file, and in the same way we don't show the complete hierarchy we should only present a minimal (valid) text to show the classes and properties...

goneall:
@tsteenbe @kestewart Unless you disagree, we'll revert the last commit and go with the simpler approach suggested by @zvr

goneall:
No disagreements so far - so I'll revert the last commit so we can have the release out this weekend

PR #300, opened 2020-04-29
update to make explicit SWHIDs and cleanup PURL entry

kestewart:
Update references per request from stefano - "official name: "SWHID", the spec has been
  updated to reflect that https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html
  Please help us promoting this new, mnemonic and pronounceable name for
  them :-)" 

Remove duplicate reference in PURL


kestewart:
This has already been handled (for the key parts by #298 .  Closing this.

PR #303, opened 2020-04-30
Added supporting documentation about the ontology

zvr:
I've created and added the following:

- a single HTML page presenting all the OWL ontology
- a diagram showing all OWL classes and inheritance relationships
- a simplified diagram, hiding the license parts

We can leave them in the `ontology` directory,
until we decide where and how to publish them.


zvr:
Sure @goneall -- next time.

Although I got confused because looking at Protege, things like `artifactOf` have `term_status: stable` but also refer to `owl:deprecatedProperty`... :smiley:

We'll talk before the next clean-up.

goneall:
> Although I got confused because looking at Protege, things like artifactOf have term_status: stable but also refer to owl:deprecatedProperty..

Oops - something to be fixed for the next release.

PR #307, opened 2020-04-30
Remove the RelationshipType_None and RelationshipType_NoAssertion

goneall:
This was an incorrect interpretation of an update to the spec to allow NONE and NOASSERTION.  The NONE and NOASSERTION applies to the SPDX Element on the right side of the relationships.


goneall:
Close - going to re open with a new branch

PR #308, opened 2020-04-30
Remove the Relationship types NONE and NOASSERTION

goneall:
These were introduced in error when interpreting the NONE and NOASSERTION for relationships.  The NONE and NOASSERTION applies to the SPDX element on the right hand side - not to the type.


kestewart:
Was an error to introduce originally,  don't add to change log.

PR #319, opened 2020-05-05
Add new top level scope

kestewart:
Set up place holder for rex to fill in.


RexJaeschke:
Kate, I reopened this thinking that I would be able to add another Commit, which contained the (very small amount of) text for this clause. But I'm not seeing how to do that, especially as the scope.md file does not actually exist until this PR is merged.

If you are able to do that for me, here is the entire content for that file:

# Scope

This Software Package Data Exchange (SPDX®) specification defines a standard data format for communicating the component and metadata information associated with software packages. An SPDX file can be associated with a set of software packages, set of files or snippets and contains information about the software in the SPDX format described in this specification.

PR #322, opened 2020-05-06
Create and populate normative-references.md

RexJaeschke:
Kate, I left all the URLs as raw text, as in "https://maven.apache.org/", which renders correctly in GitHub. If you need me to convert them to the md form, as in

> `[https://maven.apache.org/](https://maven.apache.org/)`

let me know and I'll make another Commit before you merge this PR; however, this verbose approach seems redundant. That said, if we will use a conversion tool to produce the final Word document, it might be useful to be consistent in having all links have the same format in md.


kestewart:
Agree we need the links to be standardized throughout, but we can merge this now, and fix up if necessary.    Key for this version of the spec is to go from md to web,  and md to word cleanly, so lets align with what gives smoothest path.   Topic for discussion.

PR #336, opened 2020-05-09
Revise Clause F

RexJaeschke:


chapters/composition-of-an-SPDX-document.md
```diff
@@ -43,9 +43,23 @@ Annotations and Relationships for the package may appear after the Package Infor
 
 [see Clause X](xxx.md) for details of the field in this kind of Fact.
 
-### **1.4.3** File Information Fact
+### File information Fact
```
kestewart, line 46: Should be Fields ( not Facts)

RexJaeschke: @kestewart I made a new commit to address this.

PR #337, opened 2020-05-09
Revise Clause S

RexJaeschke:


chapters/composition-of-an-SPDX-document.md
```diff
@@ -47,9 +47,20 @@ Annotations and Relationships for the package may appear after the Package Infor
 
 [see Clause X](xxx.md) for details of the fields in this kind of Fact.
 
-### **1.4.4** Snippet Information Fact
+### Snippet information Fact
```
kestewart, line 50: Shouldn't be Fact,  should be fields to match above.

kestewart: Approving this for now,  with follow up to be tracked in issue #347

PR #339, opened 2020-05-09
Revise Clause R

RexJaeschke:


chapters/7-relationships-between-SPDX-elements.md
```diff
@@ -1,6 +1,6 @@
-# 7 Relationships between SPDX Elements
+# Relationships between SPDX elements fields
```
kestewart, line 1: "elements fields" doesn't make  sense.
Lets go with "Relationships between SPDX fields".    Elements is in the model, but not really visible in the spec at this time.

chapters/composition-of-an-SPDX-document.md
```diff
@@ -55,9 +55,11 @@ Annotations and Relationships for the package may appear after the Package Infor
 
 [see Clause X](xxx.md) for details of the fields in this kind of Fact.
 
-### **1.4.6** Relationships Between SPDX Elements Fact
+### Relationships Between SPDX Elements Fact
```
kestewart, line 58: Drop the "Elements Fact" and use "fields" to be consistent with other file.

chapters/7-relationships-between-SPDX-elements.md
kestewart: applying this for now, and we'll resolve as part of issue #347

PR #370, opened 2020-05-20
Renumber Headers and Links in Annex H

RexJaeschke:
@kestewart: I retained the leading uppercase "L" in "Lite" in all headers, as I figured this was part of a proper noun and need to be preserved in every use.


kestewart:
Agree,  that's how they defined it, so it should be retained.

PR #371, opened 2020-05-20
Renumber Headers and Links in Annex J

RexJaeschke:
@kestewart I'll fill in the differences between older versions separately. For now, I've left header stubs.


kestewart:
So, to confirm we're going to need to fill in all the history since the first version?

RexJaeschke:
That is your call; we don’t have to, but I’ve made provision for it. Remember, going forward, ISO doesn’t make available any but the latest edition of a spec.

 

Rex

PR #389, opened 2020-06-03
Reformat some terms-and-definitions entries

RexJaeschke:


kestewart:
There are some definitions that need adjustment, but applying these changes right now, so we have a clean base to edit further.

PR #397, opened 2020-06-08
Reformat 7-relationships-between-SPDX-elements

jmudge:
Rename Purpose headings to Description
Convert metadata to tabular format
Remove pseudo-subsection numbers


jmudge:
@kestewart well spotted! A table of relationships number & caption has been inserted and the subsequent table numbers in this file incremented.

Issue #409, opened 2020-06-11
Missing copyrightText for `SPDXRef-Saxon` in examples


tsteenbe:
In SPDX examples, the package `SPDXRef-Saxon` is missing mandatory field `copyrightText ` see
- https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXYAMLExample-2.2.spdx.yaml#L166-L179
- https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json#L130-L144
- https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXXMLExample-v2.2.spdx.xml#L257-L273

goneall:
@tsteenbe Good catch - I'll add that in the next update.

goneall:
Fixed in #416

Issue #401, opened 2020-06-09
Excludes serialization incorrect in JSON and YAML examples


tsteenbe:
Examples contain `packageVerificationCodeExcludedFiles` that do not make sense for people familiar with JSON and YAML.

```
packageVerificationCodeExcludedFiles:
      - "excludes: ./package.spdx"
```
should be 
```
packageVerificationCodeExcludedFiles:
      - "./package.spdx"
```
see https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXYAMLExample-2.2.spdx.yaml#L231

Same issue in JSON `"packageVerificationCodeExcludedFiles" : [ "excludes: ./package.spdx" ],` should simple be `"packageVerificationCodeExcludedFiles" : [ " ./package.spdx" ],` see example https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json#L102

goneall:
Completely agree.  I would call this a bug.  I'll try to update the examples and fix the Java tools in the next week or so.

goneall:
Fixed in #416

Issue #400, opened 2020-06-09
Remove `Document` wrapper from JSON and YAML examples


tsteenbe:
Both of the SPDX JSON and YAML examples have a `Document` wrapper for the SpdxDocument while this may make sense for XML and RDF this causes an issue in JSON and YAML as there is now no easy way to have multiple SPDX documents in the same file.

Propose to remove `Document` wrapper for the JSON and YAML formats.

[1] https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXYAMLExample-2.2.spdx.yaml#L2
[2] https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json#L2

goneall:
Let's discuss - I'm not sure I agree that it is so difficult to have multiple SPDX documents in one file - you would just wrap each document.  Wrapping seems to be the approach used in some of the JSON examples: https://json.org/example.html

tsteenbe:
Aware of the JSON.org example that use wrapper help people who know XML to learn JSON.
In the unwritten rule in the JSON / YAML community is that you only use a wrapper element if there other elements at root level or if you are planning to add other elements in the future. 

In this case SPDX document will be the only element at the root so wrapper element is just extra baggage that should be removed.

Having `Document` will also make it more difficult for JSON/YAML creators and consumer to handle SPDX as items are represented the same way in code. As a consumer it now harder to loop over a response with multiple docs - I can't you can just ‘slurp in the JSON’ and you immediately have your domain object but I first have to remove `Document`. Same for creators they will have to artificially insert `Document` when they serialize their classes

tsteenbe:
>  It makes it easier for me to group the fields and makes it more human readable when combined with other JSON elements. 

JSON and YAML have implicit object type definition that why https://json-schema.org/ exists. If a document creator wants to include SPDX documents in a larger document then it that creator's responsibility to wrap the SPDX document(s) in field within their output.

We already made the decision in the SPDX tech call of November 26, 2019 to remove the Document, Package and File wrappers see https://github.com/spdx/spdx-spec/issues/157 / https://wiki.spdx.org/view/Technical_Team/Minutes/2019-11-19.

@zvr  As requested some examples below with and without the `Document` wrapper.

Current spec with `Document` wrapper:

```
{
  "Document" : {
    "SPDXID" : "SPDXRef-DOCUMENT",
[...]

PR #416, opened 2020-06-11
Update SPDX examples based on review feedback

goneall:
Resolves issue #400 
Resolves issue #401 
Resolves issue #409


goneall:
Sorry about the horrible diffs - these are all machine generated

examples/SPDXYAMLExample-2.2.spdx.yaml
```diff
@@ -1,392 +1,390 @@
...
+creationInfo:
+  comment: "This package has been shipped in source and binary form.\nThe binaries\
+    \ were created with gcc 4.5.1 and expect to link to\ncompatible system run time\
+    \ libraries."
+  created: "2010-01-29T18:30:22Z"
+  creators:
+  - "Tool: LicenseFind-1.0"
+  - "Organization: ExampleCodeInspect ()"
+  - "Person: Jane Doe ()"
+  licenseListVersion: "3.8"
```
tsteenbe, line 13: Minor issue but the license list is now at 3.9 should we update `licenseListVersion` in the examples?

goneall: Manually updated the license list to 3.9 (much faster than regenerating all the files).

tsteenbe:
@goneall  Why does `packages` and not  `package` like RDF?

https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXXMLExample-v2.2.spdx.xml#L246-L256 vs. https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L1821-L1831

Issue #254, opened 2020-04-09
Decide what to keep from Chapter 1: Rationale and how to use it


RexJaeschke:
I propose deleting this chapter and moving its content into the new ISO structure, as follows.

1.1 Charter and 1.2 Definition

- Together, they provide the content for the new clause "Scope." **DONE**

1.3 Why is a common format for data exchange needed?

- Content is a candidate for the new "Introduction." **DONE**

1.4 What does this specification cover?

- Move to new clause, "Composition of an SPDX document". **DONE**

1.5 What is not covered in the specification?

- Move to new clause, "Composition of an SPDX document". **DONE**

1.6 What does "Package" mean in the context of SPDX?

A short definition of "package" goes in "Terms and definitions." **NO FURTHER ACTION NEEDED HERE**

1.6 Format Requirements and 1.7 Conformance

- This content goes in a new "Conformance" clause. **DONE**

1.9 Differences from SPDX Specification 2.1

- Goes into a new, informative annex. **DONE**

RexJaeschke:
As all actions have been completed, I'm closing this.

WilsonHorse1:
ISC License

Copyright (c) [year], [fullname]

Permission to use, copy, modify, and/or distribute this software for any
purpose with or without fee is hereby granted, provided that the above
copyright notice and this permission notice appear in all copies.

THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.

PR #426, opened 2020-06-19
Rework this section for better flow

kestewart:
Add in missing sections,  update diagram,  make cardinality explicit for all sections.

### DIFF
--- a/chapters/composition-of-an-SPDX-document.md
+++ b/chapters/composition-of-an-SPDX-document.md
@@ -22,7 +22,7 @@ See [Clause 6](2-document-creation-information.md) for details of the fields in 
 
 ### 5.2.2 Package information section <a name="5.2.2"></a>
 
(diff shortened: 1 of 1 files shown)


RexJaeschke:
Looks good to me.

swinslow:
On the Packages description though, just to mention -- I realize this is a short summary, but I do want to make sure we preserve somewhere the [longer description of Packages](https://spdx.github.io/spdx-spec/1-rationale/#16-what-does-package-mean-in-the-context-of-spdx) that we worked through last year.

I gathered that the "Rationale" part of the spec might be changing so I just wanted to mention that we should make sure at least that that description is retained, as there was a lot of input from various folks on how best to define "Packages" for purposes of the spec.  Thanks  :)

RexJaeschke:
The Rational chapter is no longer part of the spec. See Issue #254 , which states <<A short definition of "package" goes in "Terms and definitions." The body of the text goes into the new clause, "Composition of an SPDX document." DONE>> [The intent was to redistribute that content into other clauses, one of which was new.] However, I don't see that the bulk of that was in fact moved to the new clause. Yet is sounds like it should be. If so, do you have any suggestion as to where it should be inserted into composition-of-an-SPDX-document.md?

swinslow:
Thanks @RexJaeschke -- apologies, I had missed that issue discussion previously. I think `5.2.2 Package information section` is probably the right place for that content to go, probably just prior to the Cardinality line in that section.

Issue #253, opened 2020-04-09
Create and populate the new "Terms and definitions" clause


RexJaeschke:
[The decisions made in Issue #232 will impact the way in which this md file’s name and headings are chosen.]

1. Create an md file for the new, third, numbered clause with header, “Terms and definitions.” **DONE**

1. Identify which terms are _fundamental_ to the spec and define them here using the (strict) limits required by ISO (must be no more than a phrase that could replace every occurrence of the term). (Lesser terms can defined at their first use and by setting the term name in italics in that one place.) **PENDING**; Rex made a first pass at this.

1. For a small number (20 or so) of entries, it is easier to simply hard code their numbers, bearing in mind that adding, removing, and re-arranging entries might cause some editorial effort.
While md cannot help with auto-numbering such entries, Word can via a series of SEQ fields. 
If it is decided to hard code entry numbers in the md, then do so. **Handled by #381**

1. If ISO does retain bad-break inhibitors, inhibit bad page breaks between the first, second, and third lines, of each entry in the Word spec. **PENDING** A publication-only issue

[In the final ISO spec, this will be Clause 3, hence the 3. prefixes.]

Entries must be numbered (as in 3.1, 3.2, ..., 3.n) possibly with nested, numbered sublists (as in 3.1.1, 3.1.2, ..., 3.1.n).

An entry's format is
```
3.x (in bold)
<term> (in bold>
definition ...
```

RexJaeschke:
# Specific Questions:

## Q1: SPDX document vs. SPDX file

What is the difference between an SPDX document and an SPDX file?

## Q2: The need for a name for top-level thingies

I invented the (placeholder) term “blob”, so I can talk about the top-level thingies inside a document (such as Document Creation Information thingies, Package Information thingies, and so on), as I don’t see a term for these. Might this be what is referred to as a “section”? If “section” is something else, it should be defined here as a term. (Initially I used “element” but found that term used differently in other places. What is the definition of “element”?)

Here’s just one example of why we need such a clearly defined, top-level term, like “blob”: In Chapter 3, “Package Information”, it says, “In `tag:value` format, the order in which package and files occur is syntactically significant.” Say what? My guess is that given my invention of the term blob this should read “... the order in which package [blob(s)] and file [information blob(s)] occur [in an SPDX Document|file] is syntactically significant.”

## Q3: Package

I don’t see a definition of “package.” We’ll also want one for “sub-package.”

## Q4: File

[...]

PR #443, opened 2020-06-23
Update terms and definitions

goneall:
Partially resolves issue #253 

### DIFF
--- a/chapters/terms-and-definitions.md
+++ b/chapters/terms-and-definitions.md
@@ -4,44 +4,50 @@ For the purposes of this document, the following terms and definitions apply.
 
 **3.x**  
 **annotation section**  
-*section* (3.x) type, an instance of which contains information about when and by whom an *SPDX file* (3.x) was reviewed
+*section* (3.x) type, an instance of which contains comments about an SPDX document, SPDX file, SPDX package, or SPDX snippet
 
 **3.x**  
 **document creation information section**  
-*section* (3.x) type, an instance of which contains metadata that associates analysis results with a specific version of an *SPDX file* (3.x) and license for use, and provides information on how, when, and by whom the SPDX file was created
+*section* (3.x) type, an instance of which contains metadata that associates analysis results with a specific version of an *SPDX document* (3.x) and license for use, and provides information on how, when, and by whom the SPDX file was created
 
 **3.x**  
 **section**  
-> TBD set of fields describing ...
+a part of this SPDX specification
 
 **3.x**  
 **file information section**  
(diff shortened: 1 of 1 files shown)


goneall:
Feel free to create pull requests against the updateterms branch if you would like to propose wording changes.

swinslow:
Thanks for this @goneall. I have a few questions for the group -- I'm happy to push a commit to address these, but wanted to get your input first:

* **what does "section" mean?**  It looks like we're potentially using "section" in two different contexts:
  * on its own, to mean a portion of the spec; and
  * together with e.g. "file information," etc., to mean basically a corresponding portion of an SPDX document

It looks to me like we intend for "section" to refer to just the latter point? Or I might be misunderstanding based on the way the definitions are referring to an "instance" of a portion of the spec.

* **"SPDX file" vs. "file information section"**: I gather we're now using "SPDX document" to refer to a complete document -- e.g. .spdx, .spdx.rdf, etc. Within an SPDX document, when we're talking about e.g. the information it contains about a single file, it looks like sometimes we're using "file information section" and sometimes we're using "SPDX file" to mean this. There are similar definitions for both terms. I also see "SPDX package" and "SPDX snippet" used but not define. I lean towards using just "file information section", etc. for these terms, and not defining or using "SPDX file" particularly because of the risk of overlap / confusion.

* **"information" vs. "facts"**: In some of the definitions we refer to sections containing "information", in others we refer to "facts". Unless there's a distinction I'm missing, I'm inclined to use "information" for each of these cases. I think "information" is a bit more general of a term and is probably more accurate than "facts" to refer to sections which can contain comment fields.

There's also a couple of missing line breaks that I can add when pushing a commit to address the things above.

RexJaeschke:
Re "section", ISO calls numbered text document sections "subclauses" while a whole chapter is a "clause."

At one time, Kate had me use the term "fact" for what we know call "section," so if there are occurrences of "fact" still floating around, they should (probably) be changed to "section."

goneall:
re: sections - I am confused about that this means myself.  If someone could replace my definition with a more appropriate definition that would be great.

goneall:
> "SPDX file" vs. "file information section"

I'm OK removing this definition to be consistent now that SPDX document is defined.  @RexJaeschke @kestewart If you're OK with that, I'll update the PR.

goneall:
@swinslow No opinion on the "information" vs "fact".  BTW - feel free to commit directly to the branch or create a PR if you want to update this or any of the other items.

goneall:
@swinslow @kestewart I removed the SPDX file definition.  I'll go ahead and merge this, so any update please create a new PR.

Issue #454, opened 2020-07-06
Cleanup for 2.2.1 draft spec


swinslow:
__Updated 4__: virtually all of the issues _should_ be fixed after #460 is merged. The only remaining items are:

- [x] **Clause 4.4**: confirm that indented sub-bullets are displaying correctly at https://spdx.github.io/spdx-spec/v2-draft/conformance/ after PR is merged; if they aren't, fix them
- [x] **Annex J.2**: Replace with the actual changes from v2.1 to v2.2 (what's currently shown are the changes from v2.0 to v2.1)

= = =

I'm in the midst of reviewing the draft 2.2.1 SPDX spec at https://spdx.github.io/spdx-spec/v2-draft/ for typos, and in a couple of places for minor stylistic issues.

__Updated__: completed read-through, additional issues noted below. I'll do my best to prepare a PR to fix some or all of these, but may not have it ready before tech team call tomorrow.

__Updated 2__: in process of preparing PR; I'll be checking these off as I go through, but note that they aren't addressed in the main branch until the PR is actually submitted and merged.

__Updated 3__: added items from @zvr below

- [x] **across many sections**: need to align on capitalization for “SPDX Document” vs. “SPDX document”
  - also update in SPDX ontology and JSON schema files? => yes, tracked in #459
- [x] **Copyright** and **Clause 1**: also add (R) after first instance of “Software Package Data Exchange”
- [x] **Foreword**: ISO links don’t work, not absolute
- [x] **Foreword**: Missing close paren in 3rd para
- [x] **Foreword**: Last paragraph: Rex -- Does it need to say only this? Can we change or supplement the language here to clarify that the SPDX spec development occurs within the SPDX workgroup, rather than in ISO itself?
- [x] **Introduction**: “the SPDX working group” => “the SPDX workgroup”
- [x] **Clause 1**: “set of files or snippets” => remove “set of”, already appears before “software packages”
- [x] **Clause 2**: the urls aren’t rendering as links
- [x] **Clause 3**: the “.x” section numbers will need to be filled in
- [x] **Clause 3**: I assume the terms should be alphabetized?
- [x] **Clause 3**: "file information section" includes a reference to "files (3.x)", but "file" is not defined -- should either define "file" and update section numbers / references, or else remove "(3.x)" from end of this line
  - remove parenthetical ref
- [x] **Clause 3**: after prior edits, definitions are now wrapping onto same line as terms; needs extra hard line break after each term
[...]

PR #455, opened 2020-07-07
Fix minor issues for 2.2.1

swinslow:
This addresses some, but not all, of the issues noted in #454.


goneall:
Thanks @swinslow for fixing all my Document capitalization errors as well as the other fixes!

PR #460, opened 2020-07-11
Finish most remaining cleanup edits

swinslow:
This fixes additional cleanup edits, and nearly (but not quite) completes the list of issues in #454.


swinslow:
@kestewart I _think_ that this fixes _almost_ all of the issues listed in #454, after it is merged. Can you review and merge if it looks okay?

There are two remaining to address after this is merged, which I've bumped to the top of #454:
* confirm that the sub-bullet points display correctly for the formats list -- I'll check this after the PR is merged and the draft page re-renders
* fix the "differences from v2.1 to v2.2" in Annex J -- I think you said on the call that you would be able to address this? I'm happy to be a second set of eyes afterwards.

Thank you!

RexJaeschke:
Are the formatting changes to "chapters/license-matching-guidelines-and-templates.md" correct? It seems that not all that text should be code.

swinslow:
Thanks @RexJaeschke, good point. I've reverted my edits there and instead made it so that just the `<<var . . .>>` segments themselves get backticks. Feel free to let me know if this still looks incorrect.

RexJaeschke:
@swinslow Looks good now.

Issue #456, opened 2020-07-07
license expressions grammar


zvr:
This is about **Annex D**, where we define the license expressions.

The ABNF grammar does not look correct to me...

In both definitions of `compound-expression` and `license-expression` we can delete the superfluous initial `1*1`, which says "minimum 1, maximum 1", which of course is the same as omitting it ("exactly 1").

Then, the definition of `license-expression` includes both `simple-expression` and `compound-expression`, but `compound-expression` can already be a single `simple-expression`, so there is no need to have them both.

Cynical-Optimist:
#### "`1*1` is superfluous"
I agree completely. +1

#### "`compound-expression` makes `simple-expression` superfluous"
Not quite. Look at the `"WITH"` line.
`compound-expression "WITH" license-exception-id` isn't (in general) a valid license expression. It specifically has to be a `simple-expression`. That's why they're not interchangeable.

Cynical-Optimist:
Two more issues:

* Redundant `1*`
The definition of `license-ref` involves `1*(idstring)`, but the definition of `idstring` already used a `1*`. so the second `1*` is redundant.

* Markdown removes spaces at the start lines
There are three lines that start with 2 spaces. According to the definition of ABNF, these spaces are an important part of the definition. They show that these lines are a continuation of the line above.

The spaces are present in the raw markdown file, and they're correct (they ought to be there), but they dissappear when the document is actually rendered (because markdown ignores them).

The whole definition should probably be quoted with triple backticks, and treated like a code snippet, so that markdown won't change the text.

This would also mean we could remove the backslash in the license-ref definition. (Which I assume is there just to escape the asterisk).

goneall:
It looks like this was partially resolved in 2.3, but we still have the redundant expressions.

Since we're past the review period for 2.3, I'm pushing this to a 3.0 fix.

zvr:
Closing this, as the grammar of 3.0 has the fixes already integrated.

PR #474, opened 2020-09-23
Reduce redundancy in ABNF description

Cynical-Optimist:
"1*1" is redundant in ABNF. It just means "exactly one of", which is the same thing you get if you don't put an asterisk at all.

Due to the way idstring is defined, "1*idstring" is equivalent to "idstring".

Resolves (part of) #456

### DIFF
--- a/chapters/SPDX-license-expressions.md
+++ b/chapters/SPDX-license-expressions.md
@@ -15,11 +15,11 @@ license-id = <short form license identifier in Annex A.1>
 
 license-exception-id = <short form license exception identifier in Annex A.2>
 
-license-ref = ["DocumentRef-"1\*(idstring)":"]"LicenseRef-"1*(idstring)
+license-ref = ["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring)
 
 simple-expression = license-id / license-id"+" / license-ref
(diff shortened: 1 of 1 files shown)


kestewart (commented):
From https://datatracker.ietf.org/doc/html/rfc5234#page-9

@zvr  - can you review and weigh in on the interpretation of the following. 

3.6.  Variable Repetition:  *Rule

   The operator "*" preceding an element indicates repetition.  The full
   form is:

         <a>*<b>element

   where <a> and <b> are optional decimal values, indicating at least
   <a> and at most <b> occurrences of the element.

   Default values are 0 and infinity so that *<element> allows any
   number, including zero; 1*<element> requires at least one;
   3*3<element> allows exactly 3; and 1*2<element> allows one or two.

kestewart:
Henk reviewing....

kestewart:
Moving this to 2.3.

henkbirkholz (approved):
1*1 is redundant and as well is the notation in 1*1

zvr:
This single change is correct, but we'll have a new grammar for 2.3 anyway...

kestewart (approved):

Issue #476, opened 2020-09-25
Remove deprecated sections and fields


iamwillbar:
For the 3.0 version of the specification we should remove all sections and fields that were marked as deprecated in SPDX 2.x.

goneall:
I believe this has been completed.

If there are any deprecated properties or classes, please open a new issue or (better yet) a PR.

PR #477, opened 2020-09-25
Remove deprecated clauses and fields

### LINKED ISSUE #475: Reverse order of Annex J
Currently the way Annex J is ordered every time we release a new version we will have to re-number every section on that page. If we reverse the order so that it is oldest to newest then we can append new versions without having to re-number every section.

iamwillbar:
Resolves #475, #476


kestewart:
Believe this is stale.  Closing now.   If there's a revised version, please open a new PR.

PR #494, opened 2021-01-14
Remove encapsulating 'Document' key in JSON example

rnjudge:
The JSON example document on 'master' contains an encapsulating
'Document' key. This is incorrect and the document will not
validate with it present. This commit removes the "Document"
encapsulation for the JSON example file so that it properly
validates.

### DIFF
--- a/examples/SPDXJSONExample-v2.2.spdx.json
+++ b/examples/SPDXJSONExample-v2.2.spdx.json
@@ -1,280 +1,278 @@
 {
-  "Document" : {
-    "SPDXID" : "SPDXRef-DOCUMENT",
-    "spdxVersion" : "SPDX-2.2",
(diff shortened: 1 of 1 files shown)


goneall (approved):
The order of the JSON elements is a bit different which makes the diff look more substantial than the actual changes.  The changes look OK considering the order difference.

goneall (approved):

rnjudge:
> The order of the JSON elements is a bit different which makes the diff look more substantial than the actual changes. The changes look OK considering the order difference.

Thanks Gary! Just for the record, I didn't change the order of anything in my changes, all I did was remove the encapsulating `Document` and then re-align the remaining content which is why the change looks so huge :) Thanks for merging!

PR #503, opened 2021-03-29
licensing: Add profile draft for discussion

swinslow:
This adds an incomplete, starting-point draft of the Licensing profile in Markdown format.

The draft is based on the latest 3.0 template from the tech team for profile sections, and incorporates a portion of the content from the joint legal/tech discussions last fall about what should be in the Licensing profile for 3.0.

It is incomplete in that it has placeholders without content for some of the sections at the bottom (e.g. "Other License" section; the License Expression model). It also includes _many_ **TBD** markers where I ran into questions or ambiguities while preparing this.

I'd like to ask if comments could initially focus not on the _substance_ of the profile, e.g. which fields should or should not be included. Let's start instead by focusing on the syntax and setup of the template, so that we can make sure we're aligned on that. I think that'll help unblock other profile section drafters as well. After that we can come back to the Licensing specifics as needed.

### DIFF
--- a/chapters/profile-licensing.md
+++ b/chapters/profile-licensing.md
@@ -1,0 +1,348 @@
+# N Licensing Profile
+
+The Licensing profile describes fields and classes used to convey information regarding the licenses applicable to Artifacts, as well as related information such as copyright notices.
+
+**(TBD how to express the profile identifier)**
+
+## N.1 Document
+
+### N.1.1 Summary
+
+The Licensing profile adds a field to the Document class from the Base profile.
+
+### N.1.2 Metadata
+
+**(TBD how to express this in the context of adding fields to the Document Class)**
+
+| Attribute | Value |
+| --------- | ----- |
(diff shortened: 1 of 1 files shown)


goneall (approved):
Although this is a work in progress, as discussed on the 1 April legal team, I think we should go ahead and merge this into the development branch so that PR's can be made against this proposal.

swinslow:
Reviews here are just for merging this in draft format, not for the substance  :)

kestewart:
@goneall - this needs to be reviewed to see what is stale, and what should be pulled into 3.0.

goneall (changes requested):
I believe this is no longer required as the model files will generate a licensing profile description.

@swinslow - if you agree, we can close this.

swinslow:
@goneall Agreed, this is old and all relevant should have been moved to spdx-3-model, so I'll go ahead and close.

Issue #473, opened 2020-09-17
Add PURL and CPE into the SPDX examples file


goneall:
Now that Package URL's and CPE references are supported, we should add those to the [examples](https://github.com/spdx/spdx-spec/tree/development/v2.2.1/examples)

MarkusTe:
That would help us a lot ;-)

MarkusTe:
Still not being successful in using PURL with SPDX.
Trying to [convert](https://tools.spdx.org/app/convert/) this [example](https://github.com/oss-review-toolkit/ort/blob/a6621fef22395d332ffcec2765b717229c44c0f6/reporter/src/funTest/assets/spdx-document-reporter-expected-output.spdx.yml) into Tag format --> error message in the context of package mgrs.

goneall:
@MarkusTe The error is due to [issue 16](https://github.com/spdx/spdx-java-jackson-store/issues/16) in the SPDX Java Jackson Store.

The spec isn't clear on if we should be using the Tag/Value format of `PACKAGE-MANAGER` or a more computer language friendly form of `PACKAGE_MANAGER` in the YAML and JSON versions.

It looks like ORT is using the dashes while the other SPDX tools are expecting the underscrore.

@tsteenbe @zvr any opinions on which of these formats should be used in JSON/YAML?

If the consensus is to use dashes, I can fix the Java code be compatible.

zvr:
Why am I reminded of #58?

goneall:
Good memory @zvr !  It is a bit of a pain to deal with dashes in a serialization format that doesn't support them.

Java, for examples, doesn't support dashes in enums.

sschuberth:
> It looks like ORT is using the dashes while the other SPDX tools are expecting the underscrore.

FYI, @tsteenbe fixed ORT to use underscores in https://github.com/oss-review-toolkit/ort/pull/3867/files some two weeks ago.

goneall:
> FYI, @tsteenbe fixed ORT to use underscores in oss-review-toolkit/ort#3867 some two weeks ago.

Based on this, I'll leave the Java implementation as is - it will be consistent with ORT

goneall:
I just added a PR to add a purl example.  Please review PR #509 and let me know if this works.

My apologies for the randomizing of the order - a side-effect of using tools to generate the examples.

MarkusTe:
Assuming that the examples are accepted as correct, here my findings
[...]

PR #508, opened 2021-04-27
Add a PURL external ref to all examples

goneall:
Resolves issue #473 

Note that the examples already contain a CPE external ref.

Since these example files are machine generated, the order of the elements are somewhat random making the diffs rather useless.  You can, however, run these files through compare and viewer or other verification tools to review.


goneall:
Some additional commits were accidentally added - closing and re-opening a different PR

Issue #451, opened 2020-07-03
Use tag:value instead of RDF for the value of referenceType in JSON/YAML


tsteenbe:
See https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXYAMLExample-2.2.spdx.yaml#L177 propose to  make below change so this field aligns with other field where we do not use RDF but tag:value. 

```
referenceType: "http://spdx.org/rdf/references/cpe23Type"
```

to

```
referenceType: "cpe23Type"
```

goneall:
Agreed - I'll update this on the next round of example updates.

htttd35:
❤️

goneall:
@tsteenbe I just updated the examples in PR #519 which should resolve this.

goneall:
Resolved by PR #519

PR #509, opened 2021-04-27
Add an external-ref of type purl to the example files.

goneall:
Resolves issue #473

Note that the examples already contain a CPE external ref.

Since these example files are machine generated, the order of the elements are somewhat random making the diffs rather useless. You can, however, run these files through compare and viewer or other verification tools to review.


examples/SPDXYAMLExample-2.2.spdx.yaml
```diff
@@ -158,28 +158,28 @@ packages:
...
+    checksumValue: "11b6d3ee554eedf79299905a98f9b9a04e498210b59f15094c916c91d150efcd"
   copyrightText: "Copyright 2008-2010 John Smith"
   description: "The GNU C Library defines functions that are specified by the ISO\
     \ C standard, as well as additional features specific to POSIX and other derivatives\
     \ of the Unix operating system, and extensions specific to GNU systems."
   downloadLocation: "http://ftp.gnu.org/gnu/glibc/glibc-ports-2.15.tar.gz"
   externalRefs:
+  - referenceCategory: "SECURITY"
+    referenceLocator: "cpe:2.3:a:pivotal_software:spring_framework:4.1.0:*:*:*:*:*:*:*"
+    referenceType: "http://spdx.org/rdf/references/cpe23Type"
```
sschuberth, line 173: According to https://github.com/spdx/spdx-spec/issues/451, shouldn't this be just "cpe23Type"?

sschuberth: Similar in other places, and also in the JSON.

goneall: > According to #451, shouldn't this be just "cpe23Type"?

@sschuberth I agree - this is a bug in the Java tools or libraries.  I added an issue to track: https://github.com/spdx/spdx-java-jackson-store/issues/18

goneall: I just added a commit to manually fix the reference type.

BTW - the reference type for non-SPDX defined types are still full URI's to make sure they are unique.

goneall: Just FYI - I'll be offline for a few days and may be a bit slow to responding to any additional review comments

sschuberth (approved):
I only looked at the added references. I did not verify that there are no other changes besides the ordering.

swinslow (approved):
Just looked at purl additions, not other changes, but LGTM

PR #521, opened 2021-06-23
document why Artifact of is deprecated

tardyp:
The information was hidden in the changelog for 1.8.1

### DIFF
--- a/chapters/file-information.md
+++ b/chapters/file-information.md
@@ -422,6 +422,8 @@ EXAMPLE 2 RDF: Property `spdx:copyrightText` in class `spdx:File`
 
 ## 8.9 Artifact of project name field (deprecated) <a name="8.9"></a>
 
+This field is deprecated since 1.8.4, as they can be replaced by a relationship to the more descriptive External Packages.
+
 ### 8.9.1 Description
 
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM - Thanks @tardyp

swinslow (approved):

tardyp:
@goneall Actually, I just reproduced the notice in the release notes, but I have hard time understanding how we are supposed to replace that by "external packages", actually looking for external packages in relationship section gives me nothing.

relationship is supposed to point to another spdx document, while artifactOf looks like more made to reference a DOAP project.

for context, I am working on tools-python spdx parser, which cannot understand the artifactOf relations generated by BlackDuck.

goneall:
@tardyp There are a couple of approaches to replacing an ArtifactOf with a relationship.

- Create an SPDX package with the DOAP information within the same SPDX document with `filesAnalyzed=false`.  You can then create a relationship with a relationshipType of `GENERATED_FROM` if the file is a binary generated from the DOAP Artifact.  Other relationship types may be more applicable depending on the artifact.
- Replace `ArtifactOf` with an [ExternalRef](https://spdx.github.io/spdx-spec/3-package-information/#321-external-reference).

kestewart (approved):

Issue #527, opened 2021-08-16
Conversion of tag-value to JSON creates a field not in the schema


SamuraiAku:
When I run the [official tag-value example file](https://github.com/spdx/spdx-spec/blob/v2.2/examples/SPDXTagExample-v2.2.spdx) through the online converter tool to JSON, it turns the tag "DocumentNamespace" into the JSON field "documentNamespace".  Which makes a lot of sense, except that field does not exist in the [JSON schema](https://github.com/spdx/spdx-spec/blob/v2.2/schemas/spdx-schema.json).  Instead it's believed that the field "spdxDocument" is the intended field for the namespace.  See this [issue in SPDX-specs](https://github.com/spdx/spdx-spec/issues/520).

Personally I think it would be clearest if the schema was changed to match what the tool is putting out in this instance, but either way this is a discrepancy.

goneall:
Resolved with PR #528

PR #528, opened 2021-08-16
Update JSON schema

goneall:
Update the JSON schema to be consistent with the current JSON example file.

- Remove the top level Document property
- Add documentNamespace

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -5,9 +5,7 @@
   "type" : "object",
   "properties" : {
     "SPDXID" : {
-      "type" : "string",
-      "minItems" : 1,
-      "maxItems" : 1
+      "type" : "string"
     },
     "revieweds" : {
       "description" : "Reviewed",
@@ -17,20 +15,17 @@
         "properties" : {
           "reviewer" : {
             "description" : "The name and, optionally, contact information of the person who performed the review. Values of this property must conform to the agent and tool syntax.",
-            "maxItems" : 1,
             "type" : "string"
           },
           "comment" : {
-            "maxItems" : 1,
             "type" : "string"
           },
           "reviewDate" : {
             "description" : "The date and time at which the SpdxDocument was reviewed. This value must be in UTC and have 'Z' as its timezone indicator.",
-            "minItems" : 1,
-            "maxItems" : 1,
             "type" : "string"
           }
-        }
+        },
+        "required" : [ "reviewDate" ]
       }
     },
     "hasExtractedLicensingInfos" : {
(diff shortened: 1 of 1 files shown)


goneall:
Resolves issue #527

goneall:
@SamuraiAku - this PR should resolve the issue, however, when I run it through one of the [online JSON schema validators](https://www.liquid-technologies.com/online-json-schema-validator) with the current JSON example, it seems that everything validates - even an empty JSON file.

Let me know if you have any ideas on how to improve the JSON file and I'll update the PR.

BTW - This JSON schema file is generated by the [SPDX Java tools](https://github.com/spdx/tools-java/blob/master/src/main/java/org/spdx/tools/schema/OwlToJsonSchema.java)

schemas/spdx-schema.json
```diff
@@ -4,606 +4,95 @@
...
-                    "enum" : [ "SHA256", "SHA1", "SHA384", "MD2", "MD4", "SHA512", "MD6", "MD5", "SHA224" ]
-                  },
-                  "checksumValue" : {
-                    "description" : "The checksumValue property provides a lower case hexidecimal encoded digest value produced using a specific algorithm.",
-                    "type" : "string"
-                  }
-                },
-                "description" : "A Checksum is value that allows the contents of a file to be authenticated. Even small changes to the content of the file will change its checksum. This class allows the results of a variety of checksum and cryptographic message digest algorithms to be represented."
-              },
-              "spdxDocument" : {
```
SamuraiAku, line 135: All I wanted was to change this item from spdxDocument to documentNamespace

SamuraiAku: And this is buried inside the external reference property.  There's nothing for the namespace for this document

SamuraiAku (commented):
Edit: Nevermind, I was mistyping when searching for the change.

~~@goneall If you want to remove the Document overall property I guess that's OK. But that's not what I was looking for. We need a field for documentNamespace, section 2.5 of the spec. There is nothing for that now.~~

goneall:
I plan on improving the detail of the schema substantially over the next week.  I'll be posting the improvements to this PR.

goneall:
Added licenseConcluded and licenseDeclared to schema - resolve issue #530

goneall:
Added the missing spdxElementId from relationships to the schema - resolves issue #531

goneall:
Schema updated with required fields which forces validation.  Resolved an issue where both `fileName` and `name` properties were on the `files` array - should only be `fileName`.

goneall:
This schema should be good to go - I'd appreciate any reviews.

goneall:
Renamed property describesPackages to documentDescribes - resolves issue #533

goneall:
@SamuraiAku I would like to go ahead and merge this to resolve the following issues:
- #533 
- #531 
- #530 
- #529 
- #527 

Let me know if you agree these are resolved.  

I'll leave the other issues you identified open for further discussion.

swinslow:
Hi @goneall, is this modifying the existing properties for JSON documents for SPDX 2.2?

I'm asking because @specter25 has been working on the JSON parsing / saving for the Golang tools for SPDX 2.2 as his GSoC project this summer, which has now concluded. Though I believe he has mostly been working from the example JSON file as the definition of SPDX JSON, so if this is just aligning the schema definition to fit with that example file, then hopefully that should be fine.

goneall:
> is this modifying the existing properties for JSON documents for SPDX 2.2?

@swinslow The intent is not to modify any of the existing properties.  The intent is to make the current schema consistent with the JSON example file - which is our only other documentation on the JSON serialization format.

I did test [this schema](https://raw.githubusercontent.com/spdx/spdx-spec/updatejsonschema/schemas/spdx-schema.json) against the [current JSON example](https://raw.githubusercontent.com/spdx/spdx-spec/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json) using the [JSON Schema Lint online tool](https://jsonschemalint.com/#!/version/draft-07/markup/json) and it passed.

If you have a chance, it would be helpful to run the output of the golang tools against the same schema.  If you find any issues, we can discuss here and decide if it is a schema or tool issue.

davaya:
In order to prevent the schema from validating everything including empty files, it's a good practice to include `"additionalProperties": false` on every object type.  I also include string size limits (I picked 5,000 characters out of the air) for extracted texts - the precise number isn't that important, but an SPDX file with megabytes or gigabytes in one extracted license text is probably not valid.  

We should have some invalid SPDX test files to ensure that the schema doesn't pass them - I'm sure there are plenty of examples from the NTIA plugfests, but haven't systematically gone through them yet.  Checking for false positives is as important as eliminating false negatives.

A [JSON schema](https://gist.github.com/davaya/ef0177d96874a6cd78ac8d41bb5578f7) generated from my [SPDX-v2.2 information model](https://gist.github.com/davaya/52e1910282f4b4fc64a286641b7b429a) also validates the current JSON example using JSON Schema Lint.

goneall:
Thanks @davaya for the review!  I added the additionalProperties as suggested.  I didn't add the length restrictions since they are not currently included in the spec (something we should probably add).

BTW - it looks like your schema is much more complete.  We should consider using your schema in future releases.

SamuraiAku:
> @SamuraiAku I would like to go ahead and merge this to resolve the following issues:
> 
> * [JSON schema/example has describesPackages/documentDescribes property that has no equivalent in the spec #533](https://github.com/spdx/spdx-spec/issues/533)
> * [JSON schema is missing the spdxElementId element from relationships #531](https://github.com/spdx/spdx-spec/issues/531)
> * [JSON schema is missing licenseConcluded and licenseDeclared properties in the package level #530](https://github.com/spdx/spdx-spec/issues/530)
> * [JSON schema is missing SPDXID property at the Document level #529](https://github.com/spdx/spdx-spec/issues/529)
> * [Conversion of tag-value to JSON creates a field not in the schema #527](https://github.com/spdx/spdx-spec/issues/527)
> 
> Let me know if you agree these are resolved.
> 
> I'll leave the other issues you identified open for further discussion.

@goneall Sorry for not replying earlier.  I agree that you have resolved these issues.

PR #535, opened 2021-08-20
Update field names and descriptions in externalDocumentRefs object of…

SamuraiAku:
externalDocumentRefs has three properties.  The property "checksum" is clear enough and matches up with the specification (section 2.6). But I get very confused trying to line up with other two properties "spdxDocument" and "externalDocumentId" with the corresponding properties in the specification ("DocumentRef-[idstring]" and [SPDX Document URI])

I would propose renaming "externalDocumentId" to just "SPDXID".  This makes it consistent with how IDs are named for packages, files and the whole document. 

I would propose renaming "spdxDocument" to "documentNamespace".  This makes it consistent with how the Namespace is named for the whole document

This makes it much clearer to a user trying to fill in the fields to understand what each one does, and to tie them back to the specification.

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -155,8 +155,8 @@
       "items" : {
         "type" : "object",
         "properties" : {
-          "externalDocumentId" : {
-            "description" : "externalDocumentId is a string containing letters, numbers, ., - and/or + which uniquely identifies an external document within this document.",
+          "SPDXID" : {
+            "description" : "A string with the format DocumentRef-[idstring] where [idstring] is a unique string containing letters, numbers, ., - and/or + which uniquely identifies an external document within this document.",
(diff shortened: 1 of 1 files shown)


goneall:
I agree with changing both of these field names to be clearer.

Since these changes would introduce incompatibilities, we would need to wait until the SPDX spec 3.0 to make the changes.

Note - we are doing a rather significant re-working of the external document reference model, so it should result in renaming the fields.

The current proposal is to change to a map for external ID's called `ExternalMap` with properties `externalId` which replaces `externalDocumentId`, `verifiedUsing` replacing `checksum` and `elementURL` replacing `spdxDocument`.  These are best document in the [Model diagram](https://github.com/spdx/meetings/blob/master/tech/Model.png) in the SPDX tech team minutes.

PR #543, opened 2021-09-05
Make subject and verb agree

Jayman2000:


goneall (approved):
Thanks @Jayman2000 for catching and correcting this.

PR #544, opened 2021-09-05
Clarify Package Supplier example

Jayman2000:


goneall (commented):
There is already an 'as' after `[glibc][] and [Red Hat][]`, so I don't think we want to add this additional `as`

Jayman2000:
It depends on what the intent is here. The way I read it at the moment, “the package glibc” and “Red Hat” are both “the Package Supplier”. This doesn’t seem like what was intended. Here’s why I say that:

- What package would “the package glibc” supply? Does it supply itself? Does it supply a sub-package?
- glibc is created by the GNU Project which is supported by the FSF. Red Hat is owned by IBM. It doesn’t seem correct for both of them to be the same Package Supplier because they are so separate.
- According to section 3.5, “The name of the Package Supplier must be an organization or recognized author and not a web site.” glibc is a piece of software, so I don’t think that it can be a member of an organization. I don’t think that there’s an organization that contains both glibc and Red Hat.

chapters/package-information.md
```diff
@@ -225,7 +225,7 @@ EXAMPLE 2 RDF: property `spdx:supplier` in class `spdx:Package`
 
 ### 7.6.1 Description
 
-If the package identified in the SPDX document originated from a different person or organization than identified as Package Supplier (see [7.5](#3.5) above), this field identifies from where or whom the package originally came. In some cases a package may be created and originally distributed by a different third party than the Package Supplier of the package. For example, the SPDX document identifies the package [glibc][] and [Red Hat][] as the Package Supplier, but the [Free Software Foundation][FSF] is the Package Originator.
+If the package identified in the SPDX document originated from a different person or organization than identified as Package Supplier (see [7.5](#3.5) above), this field identifies from where or whom the package originally came. In some cases a package may be created and originally distributed by a different third party than the Package Supplier of the package. For example, the SPDX document identifies the package as [glibc][] and [Red Hat][] as the Package Supplier, but the [Free Software Foundation][FSF] is the Package Originator.
```
goneall, line 228: How about:

```
For example, the SPDX document identifies the package as [glibc][] and the package supplier as [Red Hat][],
```

to make it clearer and a bit more consistent.

goneall (commented):
Agree with your comment that adding the `as` is correct.  I made a suggestion that I think makes it a bit clearer. Let me know if you agree.

chapters/package-information.md
Jayman2000: I think that that’s clearer and more consistent. I force pushed a new version of the commit that uses your suggestion, but I removed my sign-off. I’m not very familiar with DCOs, but I think that I need your permission before I can sign-off on the commit again.

goneall (commented):
LGTM - Thanks @Jayman2000

goneall:
@Jayman2000 - looks good - can you ammend your commit to add a signoff per https://github.com/spdx/spdx-spec/pull/544/checks?check_run_id=3538590402?  Otherwise, just reply here they you signoff and I'll do the merge.

Jayman2000:
It’s signed off now.

goneall:
Thanks @Jayman2000

seabass-labrax:
@Jayman2000, thank you for your contributions here and in other pull requests; we always appreciate new perspectives on SPDX!

If you'd like to be involved with developing the next version of the specification, SPDX 3.0, you are very welcome to join our weekly Tech Team calls - joining details are [available on our wiki](https://wiki.spdx.org/view/Technical_Team). Indeed, the weekly Outreach Team call is today at 15:00 UTC; do feel welcome to [join us then](https://meet.jit.si/SPDXOutreachMeeting) if press releases are more/also your cup of tea :)

PR #545, opened 2021-09-05
Correct expansion of URI/URL

Jayman2000:
See: https://www.rfc-editor.org/rfc/rfc3986.html#section-1.1.3


goneall (approved):
Agree this should be changed to Uniform consistent with the RFC.

@zvr let me know if you agree and we can merge.

PR #547, opened 2021-09-08
Remove redefinitions of external ontology elements

### LINKED ISSUE #539: Ontology should not redefine standard OWL and RDFS properties
There is no need to redefine the standard elements rdfs:comment, owl:deprecatedClass, owl:deprecatedProperty, owl:qualifiedCardinality - and in fact, the way it's done can cause problems in applications importing spdx along with other ontologies: specifically my company had a fairly hard to track down reasoning problem caused by the redefinition of rdfs:comment.
 
Further, OWL does not even have owl:deprecatedClass, owl:deprecatedProperty - it has the property owl:deprcated and the classes owl:DeprecatedClass, owl:DeprecatedProperty (notice capital D).

### LINKED ISSUE #540: Problem in ontology definition of ns:term_status
As defined it does not make sense: it's defining ns:term_status as an annotation property but then it's defining a triple with itself as subject with value "stable".
I think the intent is to assert that the spdx ontology is itself stable. 
In which case use the property with the ontology as the subject
So you'd have:

<owl:Ontology rdf:about="http://spdx.org/rdf/terms">
    <ns:term_status>stable<ns:term_status> ;

rivettp:
Fixes #539 and fixes #540
Remove local definitions of elements from external ontologies rdfs: and pointers: 
Remove restrictions on  rdfs:comment and rdfs:seeAlso which, as annotation properties, are not subject to reasoning
Use owl:deprecated instead of non-existent owl:deprecatedClass/Property and set to be consistent with term_status
Use rdfs:comment for deprecation reason previously the value of owl:deprecatedClass/Property

### DIFF
--- a/ontology/spdx-ontology.owl.xml
+++ b/ontology/spdx-ontology.owl.xml
@@ -19,8 +19,6 @@ Known issues:</rdfs:comment>
         <owl:versionInfo rdf:datatype="http://www.w3.org/2001/XMLSchema#decimal">2.2</owl:versionInfo>
     </owl:Ontology>
     
-
-
     <!-- 
     ///////////////////////////////////////////////////////////////////////////////////////
     //
@@ -30,42 +28,6 @@ Known issues:</rdfs:comment>
      -->
 
     
-
-
-    <!-- http://www.w3.org/2000/01/rdf-schema#comment -->
-
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2000/01/rdf-schema#comment">
-        <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
-    </owl:AnnotationProperty>
-    
-
-
-    <!-- http://www.w3.org/2002/07/owl#deprecatedClass -->
-
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2002/07/owl#deprecatedClass"/>
-    
-
-
-    <!-- http://www.w3.org/2002/07/owl#deprecatedProperty -->
-
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2002/07/owl#deprecatedProperty"/>
-    
-
-
-    <!-- http://www.w3.org/2002/07/owl#qualifiedCardinality -->
-
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2002/07/owl#qualifiedCardinality"/>
-    
-
-
-    <!-- http://www.w3.org/2003/06/sw-vocab-status/ns#term_status -->
-
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2003/06/sw-vocab-status/ns#term_status">
-        <rdfs:isDefinedBy>http://spdx.org/rdf/terms#</rdfs:isDefinedBy>
-        <ns:term_status>stable</ns:term_status>
-    </owl:AnnotationProperty>
-    
-
 
     <!-- 
     ///////////////////////////////////////////////////////////////////////////////////////
@@ -147,8 +109,9 @@ Known issues:</rdfs:comment>
         <rdfs:domain rdf:resource="http://spdx.org/rdf/terms#SpdxElement"/>
         <rdfs:range rdf:resource="http://usefulinc.com/ns/doap#Project"/>
         <rdfs:comment xml:lang="en">Indicates the project in which the SpdxElement originated. Tools must preserve doap:homepage and doap:name properties and the URI (if one is known) of doap:Project resources that are values of this property. All other properties of doap:Projects are not directly supported by SPDX and may be dropped when translating to or from some SPDX formats.</rdfs:comment>
-        <owl:deprecatedProperty xml:lang="en">Deprecated as of version 2.1</owl:deprecatedProperty>
-        <ns:term_status>stable</ns:term_status>
+        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
+        <rdfs:comment xml:lang="en">Deprecated as of version 2.1</rdfs:comment>
+        <ns:term_status xml:lang="en">deprecated</ns:term_status>
     </owl:ObjectProperty>
     
 
@@ -205,8 +168,8 @@ Known issues:</rdfs:comment>
         <rdfs:domain rdf:resource="http://spdx.org/rdf/terms#SpdxDocument"/>
         <rdfs:range rdf:resource="http://spdx.org/rdf/terms#Package"/>
         <rdfs:comment xml:lang="en">The describesPackage property relates an SpdxDocument to the package which it describes.</rdfs:comment>
-        <owl:deprecatedProperty></owl:deprecatedProperty>
-        <ns:term_status xml:lang="en">stable</ns:term_status>
+        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
+        <ns:term_status xml:lang="en">deprecated</ns:term_status>
     </owl:ObjectProperty>
     
 
@@ -237,7 +200,9 @@ Known issues:</rdfs:comment>
     <owl:ObjectProperty rdf:about="http://spdx.org/rdf/terms#fileDependency">
         <rdfs:domain rdf:resource="http://spdx.org/rdf/terms#File"/>
         <rdfs:range rdf:resource="http://spdx.org/rdf/terms#File"/>
-        <owl:deprecated xml:lang="en">This field is deprecated since SPDX 2.0 in favor of using Section 7 which provides more granularity about relationships.</owl:deprecated>
+        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
+        <rdfs:comment xml:lang="en">This field is deprecated since SPDX 2.0 in favor of using Section 7 which provides more granularity about relationships.</rdfs:comment>
+        <ns:term_status xml:lang="en">deprecated</ns:term_status>
     </owl:ObjectProperty>
     
 
@@ -390,7 +355,8 @@ Known issues:</rdfs:comment>
             </owl:Class>
         </rdfs:range>
         <rdfs:comment xml:lang="en">Licensing information that was discovered directly in the subject file. This is also considered a declared license for the file.</rdfs:comment>
-        <ns:term_status>deprecated</ns:term_status>
+        <ns:term_status xml:lang="en">deprecated</ns:term_status>
+        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
     </owl:ObjectProperty>
     
 
@@ -507,7 +473,8 @@ Known issues:</rdfs:comment>
         <rdfs:domain rdf:resource="http://spdx.org/rdf/terms#SpdxDocument"/>
         <rdfs:range rdf:resource="http://spdx.org/rdf/terms#File"/>
         <rdfs:comment xml:lang="en">Indicates that a particular file belongs as part of the set of analyzed files in the SpdxDocument.</rdfs:comment>
-        <owl:deprecatedProperty xml:lang="en">This property has been replaced by a relationship between the SPDX document and file with a &quot;contains&quot; relationship type.</owl:deprecatedProperty>
+        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
+        <rdfs:comment xml:lang="en">This property has been replaced by a relationship between the SPDX document and file with a &quot;contains&quot; relationship type.</rdfs:comment>
         <ns:term_status xml:lang="en">deprecated</ns:term_status>
     </owl:ObjectProperty>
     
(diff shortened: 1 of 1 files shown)


ontology/spdx-ontology.owl.xml
```diff
@@ -1736,13 +1611,6 @@ Known issues:</rdfs:comment>
                 <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
             </owl:Restriction>
         </rdfs:subClassOf>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.w3.org/2000/01/rdf-schema#comment"/>
-                <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
-                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
-            </owl:Restriction>
-        </rdfs:subClassOf>
```
goneall, line 1745: Same comment as above

ontology/spdx-ontology.owl.xml
```diff
@@ -1996,13 +1864,6 @@ Known issues:</rdfs:comment>
                 <owl:onClass rdf:resource="http://spdx.org/rdf/terms#RelationshipType"/>
             </owl:Restriction>
         </rdfs:subClassOf>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.w3.org/2000/01/rdf-schema#comment"/>
-                <owl:maxQualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:maxQualifiedCardinality>
-                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
-            </owl:Restriction>
-        </rdfs:subClassOf>
```
goneall, line 2005: Same comment as above

ontology/spdx-ontology.owl.xml
```diff
@@ -2035,14 +1896,8 @@ Known issues:</rdfs:comment>
                 <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
             </owl:Restriction>
         </rdfs:subClassOf>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.w3.org/2000/01/rdf-schema#comment"/>
-                <owl:maxQualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:maxQualifiedCardinality>
-                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
-            </owl:Restriction>
-        </rdfs:subClassOf>
```
goneall, line 2044: Same comment as above

ontology/spdx-ontology.owl.xml
```diff
@@ -2052,13 +1907,6 @@ Known issues:</rdfs:comment>
 
     <owl:Class rdf:about="http://spdx.org/rdf/terms#SimpleLicensingInfo">
         <rdfs:subClassOf rdf:resource="http://spdx.org/rdf/terms#AnyLicenseInfo"/>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.w3.org/2000/01/rdf-schema#seeAlso"/>
-                <owl:minQualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">0</owl:minQualifiedCardinality>
-                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#anyURI"/>
-            </owl:Restriction>
```
goneall, line 2060: Similar comment but on the rdfs:seeAlso type

ontology/spdx-ontology.owl.xml
```diff
@@ -2073,13 +1921,6 @@ Known issues:</rdfs:comment>
                 <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
             </owl:Restriction>
         </rdfs:subClassOf>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.w3.org/2000/01/rdf-schema#comment"/>
-                <owl:maxQualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:maxQualifiedCardinality>
-                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
-            </owl:Restriction>
-        </rdfs:subClassOf>
```
goneall, line 2082: Same comment as above

ontology/spdx-ontology.owl.xml
```diff
@@ -2206,13 +2047,6 @@ Known issues:</rdfs:comment>
...
```
goneall, line 2215: Same comment as above

ontology/spdx-ontology.owl.xml
```diff
@@ -3146,15 +2867,6 @@ IMAGE if the file is assoicated with an picture image file (MIME type of image/*
     
 
 
-    <!-- http://spdx.org/rdf/terms#reviewed -->
-
-    <owl:NamedIndividual rdf:about="http://spdx.org/rdf/terms#reviewed"/>
-    <rdf:Description>
-        <owl:qualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:qualifiedCardinality>
-    </rdf:Description>
```
goneall, line 3154: Curious what this individual was removed - does not seem related to the rest of the PR

ontology/spdx-ontology.owl.xml
```diff
@@ -706,40 +673,6 @@ Known issues:</rdfs:comment>
...
-
-    <!-- http://www.w3.org/2009/pointers#startPointer -->
-
-    <owl:ObjectProperty rdf:about="http://www.w3.org/2009/pointers#startPointer">
-        <rdfs:domain rdf:resource="http://www.w3.org/2009/pointers#CompoundPointer"/>
-        <rdfs:range rdf:resource="http://www.w3.org/2009/pointers#SinglePointer"/>
-        <ns:term_status xml:lang="en">stable</ns:term_status>
-    </owl:ObjectProperty>
-    
-
```
goneall, line 742: Agree this should be removed from the OWL document to prevent redefinitions.  This will, however, cause a problem in some of the SPDX tools implementations that depends on having a definition of these types/classes in the OWL document.  I'll need to look into how to include the OWL definitions in a separate file, or ideally reference a pre-existing definition from this OWL document.

goneall (changes requested):
Thanks @rivettp for creating the PR.  The annotation changes look fine to me.

The rdfs:comment and pointer classes are used by some utilities to generate schemas and code to check conforming SPDX document.

I made some specific comments in the file reviews.

If there is a way to reference pre-existing definitions rather than re-defining in this OWL documents, we may be able to retain the restrictions while not re-defining the types/classes.

Note that when we go to SPDX 3.0, we plan to move to SHACL which should make this much easier to implement in the tools.

rivettp:
@goneall the problems with trying to use restrictions on these annotation properties are:
- they won't have any effect on any OWL reasoner since the feature of AnnotationProperties is that they are not used for reasoning at all
- their attempted usage in such restrictions causes some tools like Protege to assume they must be  DatatypeProperties, which is why I believe the extra redefinitions of rdfs:comment etc were included in the OWL file

So to enforce the conditions you're looking for, OWL restrictions are not the way to do it. Other options are:
- use SPARQL queries as "hygiene tests" which are run against proposed ontologies. This is the approach taken by the FIBO ontology - there is a directory of tests (here  https://github.com/edmcouncil/fibo/tree/master/etc/testing/hygiene) which are picked up and run when someone checks in a new branch in GitHub.
- express the constraints using SHACL: which does not have the same limitations on checking annotation properties

ontology/spdx-ontology.owl.xml
```diff
@@ -1455,13 +1351,6 @@ Known issues:</rdfs:comment>
                 <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
             </owl:Restriction>
         </rdfs:subClassOf>
-        <rdfs:subClassOf>
-            <owl:Restriction>
-                <owl:onProperty rdf:resource="http://www.w3.org/2000/01/rdf-schema#comment"/>
-                <owl:maxQualifiedCardinality rdf:datatype="http://www.w3.org/2001/XMLSchema#nonNegativeInteger">1</owl:maxQualifiedCardinality>
-                <owl:onDataRange rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
-            </owl:Restriction>
-        </rdfs:subClassOf>
```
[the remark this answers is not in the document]
rivettp: Can you say more about how the code generator works.

ontology/spdx-ontology.owl.xml
rivettp: Sorry, I forgot to mention this one in the PR: this seemed to be another pair of spurious redefinitions added by Protege. 
1) Elsewhere #reviewed is also defined as an ObjectProperty. And separately in the file there is a rdf:Description that sets its status to deprecated (ideally this would be in the same XML element). The text I deleted is a 3rd place in the file that further defines it as a NamedIndividual (which would cause undesirable OWL punning to have it as a NamedIndividual and a Property) 
2) the qualifiedCardinality makes no sense standalone like this. It's not part of the NamedIndividual element. Even if it were, qualifiedCardinality can only validly be used as a property of an owl:Restriction.

ontology/spdx-ontology.owl.xml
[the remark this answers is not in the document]
goneall: There are 3 projects that use the RDF OWL document for checking restrictions:
- [spdx-java-rdf-store](https://github.com/spdx/spdx-java-rdf-store) supports the serialization and deserialization of SPDX files in RDF format
- [OwlToJsonSchema](https://github.com/spdx/tools-java/blob/master/src/main/java/org/spdx/tools/schema/OwlToJsonSchema.java) converts the OWL ontology into a JSON schema
- [OwlToJsonContext](https://github.com/spdx/tools-java/blob/master/src/main/java/org/spdx/tools/schema/OwlToJsonContext.java) converts the OWL ontology into a JSON schema format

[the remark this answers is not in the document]
goneall: These projects use the restrictions in the OWL document to associate properties with classes and to determine the cardinality restrictions on those properties.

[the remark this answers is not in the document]
goneall: In looking at the restrictions - since they are nested inside `<owl:Class rdf:about="http://spdx.org/rdf/terms#CreationInfo">`, would that limit the scope of this restriction to the `rdf-schema#comment` within this class or does this restriction end up being applied to all `rdf-schema#comment` uses?

ontology/spdx-ontology.owl.xml
goneall: Thanks for the explanation - I'm not exactly sure how this ended up this way, but it looks like something we should remove.

goneall:
@rivettp we are `rdfs:comment` and `rdfs:seeAlso` to represent properties in some of the SPDX classes (example - `rdfs:comment` is a property of the `spdx:SpdxElement` class).

In looking at your [explanation above](https://github.com/spdx/spdx-spec/pull/547#issuecomment-916939964), it looks like using these properties in the SPDX ontology as members of classes rather than purely as annotations for the OWL schema is causing the issue.

A couple questions for you:
- Is using these annotation properties as members of the SPDX classes improper?  Should we use something like `spdx:comment` instead of the `rdfs:comment`?
- Is there a way to continue to use these properties but change how we define the restrictions so it does not cause any redefinition of these properties outside the scope of how they are used within the SPDX classes?

If we need to change the URI for these properties, we would introduce a compatibility issue since there are may SPDX documents out there which use these properties URI's.  I can think of a few solutions, but I wanted to check my understanding of the issue first.

goneall:
@zvr Bringing you into this conversation to see if you have any suggestions.

We started using these property terms several years ago - they are part of SPDX 1.0.  I recall discussing the goal of not introducing new ontology terms if we could re-use existing standards.  I don't, however, recall any discussions specific to `rdfs:comment` or `rdfs:seeAlso`.

rivettp:
The cleanest approach might be to carry on using those annotation properties, but a different approach (not OWL restrictions) for requiring their usage. I suggested a couple of options, but not sure how the how current code generation works and hence might need to be changed.

goneall:
@rivettp Most of the tools that rely on the OWL document are maintained by myself and @zvr.  If you could let us know your thoughts in general, we can let you know if making the changes is something we can do in the short term.

Also - if you would like to discuss on a real-time webconference, we can schedule this as a topic for the SPDX tech call (information at the bottom of the [SPDX tech web page](https://spdx.dev/participate/tech/)).

A couple other thoughts.

If there is an approach where the information is available in the OWL document, that would be easiest to change in the short term.

I would like to switch over to SHACL in the future.  To change everything over to SHACL is a pretty big effort.  Something I plan to do when we release the SPDX 3.0 version of the spec.  I wonder there there is a way to incrementally move over to SHACL?

ontology/spdx-ontology.owl.xml
```diff
@@ -390,7 +355,8 @@ Known issues:</rdfs:comment>
             </owl:Class>
         </rdfs:range>
         <rdfs:comment xml:lang="en">Licensing information that was discovered directly in the subject file. This is also considered a declared license for the file.</rdfs:comment>
-        <ns:term_status>deprecated</ns:term_status>
+        <ns:term_status xml:lang="en">deprecated</ns:term_status>
+        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
     </owl:ObjectProperty>
```
goneall, line 360: I'm manually merging these changes into a new PR - Just a note, this term is not deprecated, the term-status is in error.  I'll change this to `stable` in the new PR

goneall:
I copied over the non-breaking changes to PR #716 

Closing this PR.  Once we start the schema work for 3.0, we can remove the (re)definitions of the external ontology elements.

PR #567, opened 2021-09-15
ISO-required license-related edits

RexJaeschke:


goneall:
@kestewart Another one I'd like you to review.  Removing the CC0 license references may not be appropriate for the version we maintain in the repo.  On the other hand, we want a document to mirror what was approved by ISO.

@tsteenbe suggested we upload a PDF of the ISO version to the release - that may be sufficient.  Another possibility is to maintain a separate branch for the ISO version which has the CC0 removed - that would be more work, but would solve the issue.

@RexJaeschke - Feel free to comment

RexJaeschke:
I think it best to have a base version that exactly matches the ISO one. But if you can justify otherwise, you'll need to make a note somewhere to make the same adjustment in future ISO submissions.

goneall (approved):

goneall:
Approved by @kestewart per email

goneall:
@tsteenbe Another conflict - interesting that these are all popping up a day after I successfully merged several other PR's

Issue #570, opened 2021-09-17
JSON-LD context file contains unused properties


goneall:
in the [JSON-LD context file](https://github.com/spdx/spdx-spec/blob/development/v2.2.1/ontology/spdx-ontology.context.json), every property which represents a list, there is a singular version of the property with the type and a plural version representing the set.

Example:
```
 "relationship" : {
      "@id" : "spdx:relationship",
      "@type" : "spdx:Relationship"
    },
    "relationships" : {
      "@id" : "spdx:relationships",
      "@type" : "spdx:Relationship",
      "@container" : "@set"
    },
```

This is causing validation issues in the tools Java - see https://github.com/spdx/spdx-java-jackson-store/issues/21 for details.

Since the singular form isn't really used in these contexts and is causing an validation issue, I would propose removing them.

goneall:
@zvr @sschuberth @tsteenbe Let me know if you agree we should remove the singular forms.

It may break some JSON files generated using the singular forms of the properties, but it probably shouldn't be valid.

sschuberth:
> Since the singular form isn't really used in these contexts and is causing an validation issue, I would propose removing them.

I can't really comment on that as I'm not sure about the potential side effects. However, as AFAIK neither JSON nor YAML should ever contain a field named `relationship` (singular), so removing the singular forms sounds reasonable to me.

goneall:
I created PR #573 with an updated JSON-LD context file.

@sschuberth @zvr @tsteenbe - if this looks good to you, I'll merge and update the libraries to resolve the original issue

kestewart:
@goneall - I believed this was resolved in 2.2.1.   Please reopen if you believe it's still an issue.

PR #572, opened 2021-09-20
Update JSON-LD Context file

goneall:
Resolves issue #570


goneall:
Oops - need a better base for this PR

PR #573, opened 2021-09-20
Update JSON-LD Context file - resolves issue #570

goneall:
JSON-LD file generated with OWL to JSON LD utility after merging https://github.com/spdx/tools-java/pull/30


goneall:
It looks like this revision of the JSON-LD context file includes several updates between revision 10 and revision 16 of the SPDX 2.2 OWL document.  This should now be current.

goneall:
This should resolve issue #570

sschuberth:
The non-deterministic order of properties makes this really hard to review. Some properties, like `"member"` (singular), are not gone but only changed positions. Could we also change the `OwlToJsonContext.java` tool to write out properties in alphabetical order from now on?

goneall:
> Could we also change the OwlToJsonContext.java tool to write out properties in alphabetical order from now on?

Agree.  Added https://github.com/spdx/tools-java/issues/31 to track.

goneall:
I updated the OWL to JSON Context tool to sort the context file for easier comparison.  I also updated the OWL XML document to include the CrossRef class for the listed licenses to bring it up to current status.

I've attached several files that compare the output of the now sorted JSON Context files:
- spdx-ontology.context-orig.json.txt - a sorted version of the JSON-LD context file using the version of the OWL document that generated the JSON-LD in the development/v2.2.1 branch
- spdx-ontology.context-schema-16.json.txt - a sorted JSON-LD using the updated OWL XML document
- diff-ontv10-ontv16.txt - Diff output comparing spdx-ontology.context-orig.json.txt to spdx-ontology.context-schema-16.json.txt
- diff-ontv16-issue570.txt - Diff output comparing spdx-ontology.context-schema-16.json.txt to the current spec

[spdx-ontology.context-orig.json.txt](https://github.com/spdx/spdx-spec/files/7230183/spdx-ontology.context-orig.json.txt)
[spdx-ontology.context-schema-16.json.txt](https://github.com/spdx/spdx-spec/files/7230184/spdx-ontology.context-schema-16.json.txt)
[diff-ontv10-ontv16.txt](https://github.com/spdx/spdx-spec/files/7230185/diff-ontv10-ontv16.txt)
[diff-ontv16-issue570.txt](https://github.com/spdx/spdx-spec/files/7230186/diff-ontv16-issue570.txt)

goneall:
In reviewing the diffs (attached above), diff-ontv10-ontv16.txt is as expected.

There were more changes in diff-ontv16-issue570.txt than I expected.  After a careful review, I believe all of the changes are correct.  Below are my notes from the analysis:

The following properties were removed due to having a max cardinality greater than 1
- annotation
- artifactOf
- attributionText
- creator
- crossRef
- describesPackage
- externalDocumentRef
- externalRef
- fileContributor
- fileDependency
- fileType
- hasExtractedLicensingInfo
- hasFile
- licenseInfoInFile
- licenseInfoInSnippet
- member
- packageVerificationCodeExcludedFile
- referencesFile
- relationship
- reviewed
- seeAlso

Removed due to having more concrete subclasses used in the serialized schema:
- date

Removed due to being an annotation field not used in the serialized classes:
- deprecatedClass
- deprecatedProperty
- qualifiedCardinality
- term_status

Removed due to no longer being referenced (probably should be removed from the OWL document):
- packageName (replaced by name)
- snippetName (replaced by name)

goneall:
@sschuberth If you could do a quick review and let me know if this looks reasonable, I'll go ahead and merge.

BTW - I probably overloaded this PR by adding in the OWL document changes, but it made the comparison of JSON-LD context files much easier due to the way my tooling was setup.

goneall:
I just realized, adding the OWL XML document would create a merge conflict with another PR - so I'll remove it from this PR.  I added issue #578 to track adding this to the repo.

This does create a minor inconsistency where the CrossRef details are added to the JSON-LD context before the OWL XML document has been updated.

ontology/spdx-ontology.context.json
```diff
@@ -7,472 +7,416 @@
...
+    "comment" : {
+      "@id" : "rdfs:comment",
+      "@type" : "xs:string"
     },
-    "reference" : {
-      "@id" : "rdfpointer:reference",
-      "@type" : "spdx:File"
+    "contextualExample" : {
+      "@id" : "spdx:contextualExample",
+      "@type" : "xs:anyURI"
```
sschuberth, line 69: Is the change from `"@type": "xs:string"` intended here?

sschuberth:
> Removed due to being an annotation field not used in the serialized classes:
> 
>     * deprecatedClass
>     * deprecatedProperty

Just curious, is it correct that `deprecatedVersion` is still present?

sschuberth:
> The following properties were removed due to having a max cardinality greater than 1
> 
>     * member

Note that "members" (plural) is also gone. Is this correct?

ontology/spdx-ontology.context.json
sschuberth: Similar for the `"referenceLocator"` field.

goneall:
@sschuberth 

> Just curious, is it correct that deprecatedVersion is still present?

Yes - this field is actually used as a property in the listed license, so it should remain

> Note that "members" (plural) is also gone. Is this correct?

Great catch - I missed this one.  It should not have been removed.  I'll need to do some research to find out why it was dropped.  I'll update the PR once I figure this out.

goneall:
@sschuberth I added back in the missing `member` and `members`.

Unfortunately, the fix results in a different sort since they are now sorted by ID rather than by property name.

I did a review of the changes and it looks like the only changes to the context file are the addition of the above mentioned properties and resorting any properties that were not in the SPDX namespace.

This was introduced when I added the sort - turns out the property names are not always unique without their namespace.  I added [this commit](https://github.com/spdx/tools-java/pull/30/commits/b01080ac71b547d65992b8ce86784eca0d7776d7) to fix the OWL to JSON Context utility.

BTW - the reason I missed this was I only reviewed the post-sorted version.

@sschuberth Thanks again for catching this!

sschuberth:
> I just need someone with mkdocs experience to review the updated version to see if there is any compatibility issues going from version 1.1 to version 1.2.2.

Maybe @pombredanne, being a Python guy, has experience here?

goneall:
@sschuberth @tsteenbe If one of you could approve this PR, I'll merge it in now that we have the CI working

sschuberth:
This is a bit unpleasant to review (on a commit-by-commit basis) with the revert and merge commits included, but as there seem to be no updates since my [last comment](https://github.com/spdx/spdx-spec/pull/573#issuecomment-941291206) I'm approving *without another thorough review*.

sschuberth (approved):

PR #576, opened 2021-09-23
ISO-required wrap example code

RexJaeschke:
ISO requires constant-width lines to be set in Courier New, which is a wide font. As such, more than a few example lines got wrapped by Word, in the middle of words/tokens. In the ISO spec, some still have "bad line breaks," but when I pushed this family of fixes back into the md, I *only* broke lines at spaces, so things look better. However, this *does* not exactly match the ISO spec, but the content is the same. I *don't* see that being a problem.


goneall:
@RexJaeschke If you could sign these off - it would also be nice to squash these commits to a single commit, but not necessary

goneall (approved):

PR #579, opened 2021-09-28
Update to spec version to 2.2.2 and update copyright to 2022

tsteenbe:
See individual commits for details

PR #588, opened 2021-10-29
Bump copyright year

Jayman2000:


Jayman2000:
After creating this PR, I found out that #579 already does this. Closing in favor of that one.

PR #590, opened 2021-10-29
Capitalize site name

Jayman2000:


kestewart:
@tsteenbe  can you confirm we should update the site to be 2.2.2 at the same time as fixing the capitalization?

seabass-labrax (approved):

kestewart (approved):

PR #592, opened 2021-10-29
Bibliography fixes

Jayman2000:


kestewart (changes requested):
Not sure that "<"...">" does not makes these clickable links in the spdx.github.iio docs, so not really seeing the purpose of this change.     Syntax used elsewhere is "( link text)[link]", so if it's desirable to make them clickable, probably should align with that.

seabass-labrax (approved):
Nice; thanks @Jayman2000! Who doesn't like clickable hyperlinks? :)

seabass-labrax:
I just ran a local build of @Jayman2000's branch; I can confirm that the `<url>` syntax works:

image

PR #597, opened 2021-11-01
Annex lettering fixes

Jayman2000:
At the moment, the spec has two problems with annex lettering:

1. The edition on [spdx.dev](https://spdx.dev) and the ISO edition use different letters for some annexes.
2. At least in the [spdx.dev](https://spdx.dev) edition, the annex lettering is inconsistent. For example, both [Annex G](https://spdx.github.io/spdx-spec/creative-commons-attribution-license-3.0-unported/) and [Annex H](https://spdx.github.io/spdx-spec/SPDX-Lite/) call themselves “Annex G”.

This PR fixes those two issues, documents the changes in “Differences from previous editions”, and makes it less likely for these problems to crop up again in the future.

### DIFF
--- a/chapters/conformance.md
+++ b/chapters/conformance.md
@@ -2,7 +2,7 @@
 
 ## 4.1 SPDX Current and Previous Versions <a name="4.1"></a>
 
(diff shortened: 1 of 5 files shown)


seabass-labrax (approved):
Thank you, @Jayman2000! From your table in Annex I (the new Annex I!) It looks like the ordering was disturbed by the ISO version not being explicitly under the Creative Commons licence.

PR #599, opened 2021-11-01
Make nested lists actually nest

Jayman2000:


kestewart (approved):
Thanks for fixing up the indentation!

PR #609, opened 2021-11-18
Add ExternalRef field to SPDX Lite

heliocastro:
Introduce Clause 7.21, External Ref field to SPDX Lite as information
like security CPE strings became mandatory in some industries.

Signed-off-by: Helio Chissini de Castro <helio.chissini-de-castro@bmw.de>


heliocastro:
@kestewart Would be nice too, considering that this can became a blocker for NTIA adoption to the lite format and then force to rely on the full SPDX doc

kestewart:
Reviewed in call, and agreed this should be added to 2.3.   @seabass-labrax will generate a new one against 2.3

kestewart:
Thanks to William's coaching, we've changed this.    Applying.

Issue #612, opened 2021-12-23
JSON schema does not support PERSISTENT-ID category of external references


coderpatros:
The schema only supports `[ "OTHER", "SECURITY", "PACKAGE_MANAGER" ]` for reference category.

Relevant section https://github.com/spdx/spdx-spec/blob/2be70fcdf33c7e54290c17924a2b471d87afa0f0/schemas/spdx-schema.json#L340-L344

seabass-labrax:
Hello @coderpatros; thank you for opening this issue! I've made pull request #613 to fix this. Happy Christmas! :smiley::christmas_tree::snowman_with_snow:

goneall:
Resolved with PR #613

PR #613, opened 2021-12-23
Add PERSISTENT-ID externalRef type to JSON schema

seabass-labrax:
Resolves issue #612. Thanks to @coderpatros for the report!

Signed-off-by: Sebastian Crane <seabass-labrax@gmx.com>

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -340,7 +340,7 @@
                 "referenceCategory" : {
                   "description" : "Category for the external reference",
                   "type" : "string",
-                  "enum" : [ "OTHER", "SECURITY", "PACKAGE_MANAGER" ]
+                  "enum" : [ "OTHER", "SECURITY", "PACKAGE_MANAGER", "PERSISTENT-ID" ]
                 },
                 "referenceLocator" : {
                   "description" : "The unique string with no spaces necessary to access the package-specific information, metadata, or content within the target location. The format of the locator is subject to constraints defined by the <type>.",


schemas/spdx-schema.json
```diff
@@ -340,7 +340,7 @@
                 "referenceCategory" : {
                   "description" : "Category for the external reference",
                   "type" : "string",
-                  "enum" : [ "OTHER", "SECURITY", "PACKAGE_MANAGER" ]
+                  "enum" : [ "OTHER", "SECURITY", "PACKAGE_MANAGER", "PERSISTENT-ID" ]
                 },
                 "referenceLocator" : {
                   "description" : "The unique string with no spaces necessary to access the package-specific information, metadata, or content within the target location. The format of the locator is subject to constraints defined by the <type>.",
```
zvr, line 346: Do you mind also changing the final `<type>` here to `<referenceType>` ?
It's an unrelated change, but it's nearby ;-)

goneall:
The RDF OWL document also needs to be updated.

I'm actually generating the JSON schema file from the OWL document.

I have another fix for the JSON file - I'll update the OWL document and regenerate the JSON file in a separate PR within the next couple of days.

seabass-labrax:
Ah, thanks for the information @goneall. Looks like it's already in the OWL document: https://github.com/spdx/spdx-spec/blob/96e7beab922a3afe2d2b334da2d999d9890fa36d/ontology/spdx-ontology.owl.xml#L2692

goneall:
> Looks like it's already in the OWL document

Interesting - it should be in the generated JSON file.  I'll have to review this and see why it's not getting populated.

The JSON schema file is generated with the [OWL to JSON](https://github.com/spdx/tools-java/blob/master/src/main/java/org/spdx/tools/schema/OwlToJsonSchema.java) utility.  Perhaps I ran this with the wrong version of the schema?

zvr:
@seabass-labrax FYI it's not in the OWL ontology.
You should be looking at what can be a `referenceCategory` and you'll see that it only has the original three alternatives:

https://github.com/spdx/spdx-spec/blob/2be70fcdf33c7e54290c17924a2b471d87afa0f0/ontology/spdx-ontology.owl.xml#L465-L489

goneall:
Turns out the there was a bug in the utility that produces the schema - a typo in the enumerations class.  I just created a [pull request](https://github.com/spdx/Spdx-Java-Library/pull/73) to resolve.

In processing the OWL document, I look at both the type and the domain to construct the enumerations - so the OWL document is OK.

I'm thinking of removing the domain references since it is out of date for both the checksum algorithms and reference categories.  This can be done in a future PR.

goneall:
> Do you mind also changing the final <type> here to <referenceType>

@zvr I think type type should be string since the enumeration values are just a category based on a brief google search on types for enums in JSON schemas (e.g. this [stackoverflow article](https://stackoverflow.com/questions/30924271/correct-way-to-define-array-of-enums-in-json-schema)) - but I could be wrong in this.

If you think these should have a different value, let's track in a separate issue.  I would need to update the schema generator to produce this.

goneall:
The schema generator is now correctly producing the PERSISTENT-ID, however, it doesn't sort the properties so it is a bit of a mess to review.

Suggest we accept this PR and I'll change the schema generator to sort the properties and create a separate PR with the sorted schema.

goneall (approved):

goneall:
I'm going to go ahead and merge this PR.

PR #619, opened 2022-01-09
Update JSON Schema PERSISTENT-ID to PERSISTENT_ID

goneall:
Resolves issue#618

This schema was generated from the OWL document using a different sort
order than the previous schema.  Going forward, the sort will be more
consistent.

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -8,112 +8,6 @@
       "type" : "string",
       "description" : "Uniquely identify any element in an SPDX document which may be referenced by other elements."
     },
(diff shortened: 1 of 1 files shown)


goneall:
I reviewed the 2 versions using a JSON diff and the only differences I found was the order of properties and the change from `PERSISTENT-ID` to `PERSISTENT_ID`

kestewart:
@goneall - should this be 2.2.2 or include in 2.3?

goneall:
>  should this be 2.2.2 or include in 2.3?

@kestewart I'm thinking 2.2.2

PR #622, opened 2022-02-08
Add Package Purpose field

### LINKED ISSUE #621: Add a type parameter to SPDX Package
When discussing the security vulnerability use cases, it would be valuable to know if an SPDX Package represents source files used to compile a package, a binary distribution of a package, or a container.

This would also allow for higher fidelity translations between CycloneDX and SPDX SBOM types.

Propose the cardinality be 0..* since more than one type may apply (e.g. the packageFile contains the binary executables plus all the source files used to create it).  Making it optional would allow compatibility with the current SPDX release.

Proposed initial set of cardinality values:
- sourcedistro - A source distribution of a package where all the source files necessary to compile or run an application or library is present
- application (compatible with CycloneDX)
- framework (compatible with CycloneDX)
- library (compatible with CycloneDX)
- container (compatible with CycloneDX)
- operating-system (compatible with CycloneDX)
- device (compatible with CycloneDX)
- firmware (compatible with CycloneDX)
- binarydistro - A single file binary distribution.  May map to CycloneDX File.
- distributedfile - A single file which may have its own version and originator (e.g. a properties file or a script).  May map to CycloneDX File.

nishakm:
Fixes #621

Signed-off-by: Nisha K <nishak@vmware.com>

### DIFF
--- a/chapters/package-information.md
+++ b/chapters/package-information.md
@@ -1417,6 +1417,55 @@ EXAMPLE 2 RDF: Property `spdx:attributionText` in class `spdx:Package`
 </Package>
 ```
 
+## 7.24 Package Purpose field <a name="7.24"></a>
+
+### 7.24.1 Description
+
+This field provides information about the purpose of the identified package. Package Purpose is intrinsic to how the package is being used rather than the content of the package. The options to populated this field are limited to:
+
+`APPLICATION` if the package is a software application;
+`FRAMEWORK` if the package is a software framework;
+`LIBRARY` if the package is a software library;
+`CONTAINER` if the package refers to a container image which can be used by a container runtime application;
+`OPERATING-SYSTEM` if the package refers to an operating system;
+`DEVICE` if the package refers to a chipset, processor, or electronic board;
+`FIRMWARE` if the package provides low level control over a device's hardware;
+`SOURCE` if the package is a collection of source files;
+`ARCHIVE` if the package refers to an archived collection of files (.tar, .zip, etc);
+`FILE` if the package is a single file which can be independently distributed (configuration file, statically linked binary, Kubernetes deployment, etc);
+`INSTALL` if the package is used to install software on disk;
+`OTHER` if the package doesn't fit into the above categories.
+
+The metadata for the Package Purpose field is shown in Table 36.
+
+**Table 36 — Metadata for the package purpose field**
+
+| Attribute | Value |
+| --------- | ----- |
+| Required | No |
+| Cardinality | 0..* |
+| Format | `APPLICATION` \| `FRAMEWORK` \| `LIBRARY` \| `CONTAINER` \| `OPERATING-SYSTEM` \| `DEVICE` \| `FIRMWARE` \| `SOURCE` \| `ARCHIVE` \| `FILE` \| `INSTALL \| `OTHER` \|
+
+### 7.24.2 Intent
+
+This field is a reasonable estimate of the package usage from the producer and consumer perspective from which both parties can draw conclusions about the context in which the package exists.
+
+### 7.24.3 Examples
+
+EXAMPLE 1 Tag: `PackagePurpose:`
+
+```text
+PackagePurpose: FRAMEWORK
+```
+
+EXAMPLE 2 RDF: Property `spdx:purpose` in class `spdx:Package`
+
+```text
+<Package rdf:about="cluster-api">
+    <packagePurpose rdf:resource="packagePurpose_container" />
+</Package>
+```
+
 [Bazaar]: http://bazaar.canonical.com/
 [FSF]: http://www.fsf.org/
 [Git]: https://git-scm.com/


nishakm:
@goneall, @iamwillbar This is a first-pass PR. A review is appreciated :)

nishakm:
cc @puerco

swinslow:
Thanks for this @nishakm! I haven't been closely involved with the discussions around adding a Package Purpose field, but here are a few random thoughts anyway  :)

* For the PR, should this maybe be merging into a `development/v2.3` branch rather than v2.2.2? I think 2.2.2 would be a point release, just for bugfixes but not to add new fields. @goneall @kestewart do we need / want to create a separate `development/v2.3` branch?
* I tend to think that it should be optional and the cardinality should be `0..*` rather than `1..1`:
  * 0 rather than 1 on the left, because I don't think it should be a mandatory field. Both because (A) not all use cases for SPDX Documents are going to be appropriate (or possible) to define a Package Purpose for every Package; and also (B) because it would make existing 2.1, 2.2, etc. SPDX Documents no longer valid, which would be a breaking change for a minor release.
  * `*` rather than 1 on the right, because there may be cases where a Package has multiple "Purposes". E.g., a Package could contain the source code for both an application and a library, and therefore could have 3 PackagePurposes. I'm also thinking by comparison the [File Type field](https://spdx.github.io/spdx-spec/file-information/#83-file-type-field), which has similar purpose and values and is `0..*`
* This is a more minor point, but it would also probably be useful to clarify what is meant by some of the different values, e.g. the distinction between APPLICATION vs. FRAMEWORK vs. LIBRARY.
  * The existing [File Type field](https://spdx.github.io/spdx-spec/file-information/#83-file-type-field) says that APPLICATION there means a specific category of MIME types; does that apply here also?
  * I'm not sure if there's an obvious distinction of FRAMEWORK vs. LIBRARY -- there probably is, but it might be helpful to clarify that.

Hope this helps -- and apologies if any of this is contrary to things the tech team has discussed separately!

nishakm:
Should I be turning this into a Draft PR?

goneall:
Suggest adding one more enumeration:

- FILE - A single distributed file that can be independently distributed (e.g. a properties file or a configuration file)

This would allow mapping to the CycloneDX file type and also solve some of the issues @rjb4standards has raised.

goneall:
> I tend to think that it should be optional and the cardinality should be 0..* rather than 1..1

I'm leaning toward 0..1 and changing the semantics slightly to be the "primary" purpose of the package.  This would making the mappings to CycloneDX easier.

That being said, there is value to having the multiple purposes represented in that it may help identify expected properties to be present.  However, this scenario may be better addressed through Profiles rather than having multiple purposes.

puerco:
Thanks for filing this @nishakm!

I agree with @swinslow and @goneall , I would not make it mandatory as it would break compatibility with older docs.

I'm wondering if _purpose_ really captures the notion we want. Talking about _purpose_ sets the origin of the subject matter within the SBOM. In other words, it sounds as if the SBOM were dictating what the software should be used for while we are really trying to indicate what kind of bundle the package is describing. The said bundle has its inherent properties which the SBOM can only describe but not set.

I'm sorry, I know it is ontological pedantry, I just wanted to note the misalignment. I think I would add an enumerated `PackageKind` or `PackageType` mimicking `FileType`, but if naming it `Purpose` makes it compatible with 3.0, then there's no point to discuss. :)

rjb4standards:
I believe the 2.3 effort is considering the addition of a Classification element; I'm wondering if PackageType  would be more appropriate and in harmony with 3.0, if @puerco suggestion passes muster.

kestewart (changes requested):
Thanks for getting this started @nishakm!   Nice work. 

Agree with Gary - this should be optional, so Cardinality 0..1

Of the Purposes (or Classifications)  we've got listed, some are beyond software, so we should discuss further.    

We should probably discuss if the Examples should include JSON as well at this point?   Topic for the next tech call I guess.

kestewart:
@rjb4standards - FileTypes is referencing MIME types today, so as this has a different semantics, we should use a different term to describe it.   Classification, Purpose, PrimaryPurpose, ???  

yeah let's see if we can come to agreement either in this thread or at one of the meetings.

rjb4standards:
I defer to the group consensus on the name to use. I'll be happy if I can report a version on a file that is part of an installation package. This fills the NTIA minimum check in SAG-PM that is kicking out SPDX SBOM for non-conformance with NTIA minimum standards. That solves a problem for me.

Also, we are currently coding in support for CycloneDX VEX for vulnerability reporting and look forward to doing the same for whatever the SPDX defects team produces to answer the question: 
What is the vulnerability status of product P, version V from Supplier S at time(t) at the SBOM component level?

Thanks to all for their technical contributions in helping to solve these problems.

seabass-labrax:
> Suggest adding one more enumeration:
> 
>     * FILE - A single distributed file that can be independently distributed (e.g. a properties file or a configuration file)

@goneall, wouldn't a `FILE` always be for a specific purpose? For instance it could be a `PATCH` or `CONFIGURATION`.

rjb4standards:
+1 for FILE as a catchall for objects that have a version identifier that don't fit comfortably into another calss.

goneall:
> wouldn't a FILE always be for a specific purpose? For instance it could be a PATCH or CONFIGURATION.

@seabass-labrax I agree a FILE will likely have a specific purpose, however, it may not be easy for the document creator to know the purpose of the file and it may not always be important to the use case.

I would suggest having a generic FILE purpose and if there are specific use cases where there is value in knowing the purpose having those as additional purposes.

rjb4standards:
Quick question, is PURPOSE the current element name to represent CLASSIFICATION info, i.e. File, Library, installPackage, etc. in V 2.3 or are these two different concepts?
Thanks...

rjb4standards:
Thanks, Gary. I can live with a minimum of 2 classifications:
Installation
File
But will be happy to support whatever the team decides for classifications.
The other important item for v 2.3 is the ability to identify an SBOM with a vulnerability report, that updates independently of the SBOM, and is always available for download to check the vulnerability status of the SBOM over time.

nishakm:
To summarize everyone's feedback thus far:

1. Cardinality should be `0..1`
2. Add File enumeration: `FILE - A single file that can be independently distributed (e.g. a properties file or a configuration file)`

Do I understand correctly that we are unclear on whether this field should be "PackagePurpose" or "PackageType"?

goneall:
> Do I understand correctly that we are unclear on whether this field should be "PackagePurpose" or "PackageType"?

@iamwillbar - Did we already discuss and agree on the name for this field?  If so, we should go with the previously chosen field name.

iamwillbar:
@goneall For SPDX 3.0 there was agreement on PackagePurpose. Type was deemed to be ambiguous.

nishakm:
On Cardinality: `0..*` is preferred.

nishakm:
From Kate: `FILE` is confusing. Need another term. `FIRMWARE` could also be a file.

nishakm:
@kestewart @goneall Will a 2.3 branch be created soon? :)

goneall:
> @kestewart @goneall Will a 2.3 branch be created soon? :)

@kestewart @tsteenbe were you going to set this up?

chapters/package-information.md
```diff
@@ -1417,6 +1417,54 @@ EXAMPLE 2 RDF: Property `spdx:attributionText` in class `spdx:Package`
...
+`OTHER` if the package doesn't fit into the above categories.
+
+The metadata for the Package Purpose field is shown in Table 36.
+
+**Table 36 — Metadata for the package name field**
+
+| Attribute | Value |
+| --------- | ----- |
+| Required | Yes |
+| Cardinality | 0..* |
```
swinslow, line 1445: If Cardinality will be `0..*`, then I think Required should be "No" rather than "Yes"?

chapters/package-information.md
```diff
@@ -1417,6 +1417,54 @@ EXAMPLE 2 RDF: Property `spdx:attributionText` in class `spdx:Package`
...
+## 7.24 Package Purpose field <a name="7.24"></a>
+
+### 7.24.1 Description
+
+This field provides information about the purpose of the identified package. Package Purpose is intrinsic to how the package is being used rather than the content of the package. The options to populated this field are limited to:
+
+`APPLICATION` if the package is a software application;
+`FRAMEWORK` if the package is a software framework;
+`LIBRARY` if the package is a software library;
+`CONTAINER` if the package refers to a container image which can be used by a container runtime application;
```
swinslow, line 1429: Here and below, maybe consider whether "refers to" is the intended language to use here?

In the first few entries, it says "if the package **is** a...". Is it meant to mean something different where it says "refers to" instead of "is"?

chapters/package-information.md
```diff
@@ -1417,6 +1417,54 @@ EXAMPLE 2 RDF: Property `spdx:attributionText` in class `spdx:Package`
...
+`OPERATING-SYSTEM` if the package refers to an operating system;
+`DEVICE` if the package refers to a chipset, processor, or electronic board;
```
swinslow, line 1431: I'm not sure I understand the meaning of `DEVICE` as defined here. Is the intention that this would be used for:

1. files which are the actual embodiment of a chipset, etc. (such as PCB images, or VHDL files, etc.); or
2. files which are specific to a particular device (such as architecture-specific files in the Linux kernel sources)

I'm guessing the intention is 1 but not 2, but the current definition seems to potentially apply to either one... maybe consider clarifying whether that's the intended meaning.

chapters/package-information.md
```diff
@@ -1417,6 +1417,54 @@ EXAMPLE 2 RDF: Property `spdx:attributionText` in class `spdx:Package`
...
+`FIRMWARE` if the package provides low level control over a device's hardware;
+`SOURCE` if the package is a collection of source files;
+`ARCHIVE` if the package refers to an archived collection of files (.tar, .zip, etc);
+`FILE` if the package is a single file which can be independently distributed (configuration file, statically linked binary, Kubernetes deployment, etc);
```
swinslow, line 1435: "Kubernetes deployment" seems a bit odd to call out here. Is Kubernetes always or typically deployed as a single file? Apologies if it is, I genuinely don't know  :)

chapters/package-information.md
```diff
@@ -1417,6 +1417,54 @@ EXAMPLE 2 RDF: Property `spdx:attributionText` in class `spdx:Package`
...
+`OTHER` if the package doesn't fit into the above categories.
+
+The metadata for the Package Purpose field is shown in Table 36.
+
+**Table 36 — Metadata for the package name field**
```
swinslow, line 1440: The caption here says "the package name field", should that be "the package **purpose** field"?

swinslow (commented):
I have a few specific comments / suggested edits for particular parts of this, which I'm including inline.

I'd also asked earlier in the conversation thread about whether some of the definitions can be clarified, in particular FRAMEWORK vs. LIBRARY. But I know I haven't been on tech team calls for a long time so you may have discussed or resolved any of these comments already. Just sharing in case they're helpful. Thanks!

chapters/package-information.md
rjb4standards: @nishakm does this proposal require the use of "CAPITAL" letters for the enumerated set of 'PackageProposal" content model members? Thanks.

nishakm: Typically, enumerations are written with CAPITAL letters, at least in code. To be honest, I was just following the pattern in the model document :).
https://github.com/spdx/meetings/blob/main/tech/Model.png

nishakm:
Some questions have come up that I would like clarification on:
- Alternative to "FILE" as the enumeration of singleton file distribution
- Addition of "INSTALL" enumeration
- Clarification of "DEVICE" enumeration
- How to describe "CONTAINER" enumeration
- What to use for kubernetes distribution

rjb4standards:
Suggestion:
Alternative to "FILE" as the enumeration of singleton file ### object

kestewart (approved):

kestewart:
From discussion on call.

Agreement to add: 
INSTALL - purpose to identify as installable image (ie. RPM, Deb, archive file, etc.)   

Next steps:  Nisha to revise PR, then submit for re-review.

rjb4standards:
> 

@nishakm , please include the proposed language for section 5.2  in your edits. Thanks.. https://github.com/spdx/spdx-spec/issues/628

nishakm:
@rjb4standards Not sure if I added the information you requested. Please review.

chapters/package-information.md
```diff
@@ -1,4 +1,25 @@
...
+a collection of one or more sub-packages
+a Git repository snapshot from a particular point in time
+a open-source library, i.e. OpenSSL
+a javascript file
+a customer script file
+a index,html file
+a LICENSE file
+a CHANGE log 
+a properties file
+```
```
rjb4standards, line 22: @nishakm  this looks good. Will need to sync this material with section 5.2.2 which needs to be updated accordingly
https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/composition-of-an-SPDX-document.md
:
5.2.2 Package information section

If SPDX information is being used to describe packages, then one instance of the Package Information per package being described shall exist. It provides important meta information about the package as a whole. Packages are an abstract concept that can be used to refer to any distribution of software, typically consisting of one or more files and capable of containing sub-packages. Starting with SPDX 2.0, it is not necessary to have a package wrapping a set of files.

A Package refers to any unit of content that can be associated with a distribution of software. Typically, a Package is composed of one or more files. An SPDX document may, but is not required to, provide details about the individual files comprising a Package (see [Clause 8](https://spdx.github.io/spdx-spec/file-information/)).

Any of the following non-limiting examples may be (but are not required to be) represented in SPDX as a Package:

rjb4standards:
Nisha,

 

I added a comment regarding the need to sync up section 5.2.2 with section 7 concepts.

 

Thanks,

 

Dick Brooks

  

Active Member of the CISA Critical Manufacturing Sector, 

Sector Coordinating Council – A Public-Private Partnership

 

 <https://reliableenergyanalytics.com/products> Never trust software, always verify and report! ™

 <http://www.reliableenergyanalytics.com/> http://www.reliableenergyanalytics.com

Email:  ***@***.***> ***@***.***

Tel: +1 978-696-1788

chapters/package-information.md
nishakm: Oh I see! I will update accordingly.

nishakm:
@rjb4standards I just realized what you were asking as a request to sync the package information section. There is a lot to synchronize, so if you don't mind, I would like to submit a separate PR for that bit once this PR is merged. I think you filed an issue for it so I will refer to that.

rjb4standards:
No problem at all Nisha - I defer to your better judgement as to how to accomplish this matter. Thanks for the update.

seabass-labrax (approved):

seabass-labrax (approved):

PR #659, opened 2022-04-19
JSON Schema Fixes Add AMENDS and required array properties

goneall:
Resolves issue #656

Signed-off-by: Gary O'Neall <gary@sourceauditor.com>

PR #660, opened 2022-04-20
Make creators attribute mandatory in JSON schema

dmholtz:
In the JSON schema, the `creators` field of `createrInfo` was not listed as a `required` although the latest specification denotes the creator as mandatory (cp. 2.8.3 Cardinality, p. 19, [SPEC](https://spdx.dev/wp-content/uploads/sites/41/2020/08/SPDX-specification-2-2.pdf)).

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -63,7 +63,7 @@
           "type" : "string"
         }
       },
-      "required" : [ "created" ],
+      "required" : [ "created, creators" ],
       "additionalProperties" : false,
       "description" : "One instance is required for each SPDX file produced. It provides the necessary information for forward and backward compatibility for processing tools."
     },


goneall:
I was assuming the `"minItems" : 1` in the creators property provided the same validation of required.  

After thinking about this, I am likely wrong in this assumption and will need to add any list with a minimum number of elements >=1 to the required list of properties.

@dmholtz let me know if agree my old assumption was wrong.

dmholtz:
Thank you very much for the quick response!

In JSON schema, setting `"minItems" : 1` has slightly different semantics: It means, that we must not provide an empty array for the `creators` field. However, omitting the `creators` property is perfectly fine with this schema. Example:

```json
"creationInfo" : {
    "created" : "2010-01-29T18:30:22Z"
}
```

However, an empty creators array is not allowed:
```json
"creationInfo" : {
    "created" : "2010-01-29T18:30:22Z"
    "creators": []
}
```

I think the idea of the specification is that a creator is always listed, and my PR should fix this. Nonetheless, we still need `"minitems:" 1` to avoid empty lists.

goneall:
@dmholtz Thanks for the detailed answer.

The JSON file is generated from the RDF OWL document.  I'll fix the [generator](https://github.com/spdx/tools-java/blob/master/src/main/java/org/spdx/tools/schema/OwlToJsonSchema.java) code to add the required parameter.  This will probably result in a few more changes.

I'll probably just amend PR #659 to include this fix as well once it is regenerated.

Let's leave this PR open to track the issue until it is resolved in the other PR.

dmholtz:
@goneall Thanks, I'll leave the PR open. Of course, combining both fixes is more efficient. From skimming the project, I didn't notice that schema has been generated :)

PR #669, opened 2022-05-03
Align package information with doc composition

nishakm:


kestewart:
@nishakm - can you please add in DCO and description.

swinslow (commented):
Thanks @nishakm! One comment on formatting, otherwise +1 from me.

PR #671, opened 2022-05-06
"Usage Profile" to describe the intended usage by package supplier

yoshi-i:
"Usage Profile", the intended usage and license conditions assumed by the package supplier are appended to the generated package in a manner that supplements those conditions contained in the source code.

Signed-off-by: Yoshiyuki Ito <yoshiyuki.ito.ub@renesas.com>

### DIFF
--- a/chapters/usage-profile.md
+++ b/chapters/usage-profile.md
@@ -28,7 +28,7 @@ To describe detail intended usage of the package assumed by the package supplier
 EXAMPLE 1 Tag: `UsageScope:`
 
 ```text
-UsageScope: SPDXRef-a_specific_binary_verNNN: shipped for consumer product/ medical system / automotive system
+UsageScope: SPDXRef-a-specific-binary-verNNN: shipped for consumer product/ medical system / automotive system
 ```
 
 EXAMPLE 2 RDF: Property `spdx:UsageScope` in any `spdx:SpdxDocument`, `spdx:Package` or `spdx:File`
@@ -38,7 +38,7 @@ EXAMPLE 2 RDF: Property `spdx:UsageScope` in any `spdx:SpdxDocument`, `spdx:Pack
       <UsageScopeElement>
         <File rdf:about="http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82..."
       </UsageScopeElement>
-      <Usage>SPDXRef-a_specific_binary_verNNN: shipped for consumer product/ medical system / automotive system
+      <Usage>SPDXRef-a-specific-binary-verNNN: shipped for consumer product/ medical system / automotive system
       </Usage>
     </UsageScope>
   ...
@@ -69,15 +69,15 @@ To provide additional context. For example, describing prerequisite assumptions 
 EXAMPLE 1 Tag: `UsageScopeComment:`
 
 ```text
-UsageScope: SPDXRef-a_specific_binary_verNNN: shipped for consumer product/ medical system / automotive system
-UsageScopeComment: A_specific_binary_verNNN is compiled and velified as production level quality
+UsageScope: SPDXRef-a-specific-binary-verNNN: shipped for consumer product/ medical system / automotive system
+UsageScopeComment: A-specific-binary-verNNN is compiled and velified as production level quality
 ```
 
 EXAMPLE 2 RDF: Property `rdfs:comment` in class `spdx:UsageScope`
 
 ```text
 <UsageScope>
-         <rdfs:comment> A_specific_binary_verNNN is compiled and velified as production level quality</rdfs:comment>
+         <rdfs:comment> A-specific-binary-verNNN is compiled and velified as production level quality</rdfs:comment>
 </UsageScope>
 ```
 
--- a/chapters/usage-profile.md
+++ b/chapters/usage-profile.md
@@ -1,0 +1,137 @@
+# 31 Usage Profile section
+
+## 31.1 Usage Scope field <a name="31.1"></a>
+
+### 31.1.1 Description
+
+Software packages are sometimes generated with specific Build conditions based on package supplier's license conditions and intended usage in addition to original software source code. In those cases, it prefer to package supplier clearly state into the SPDX document that the intended usage of the package as envisioned when setting the Build conditions by the package supplier.
+In other cases, some specific licenses out of dual or multiple licenses of original software source code may be applied to the package according to the Build conditions. Or, the inspection conditions for packages may also have to be clearly stated for each package.
+
+With this "Usage Scope" field, the intended usage and license conditions assumed by the package supplier are appended to the generated package in a manner that supplements those conditions contained in the source code.
+
+
+Table 31.1 — Metadata for the usage scope field
+
+| Attribute | Value |
+| --------- | ----- |
+| Required | No |
+| Cardinality | 0..* |
+| Format | ["DocumentRef-"[idstring]":"]SPDXID \<single line of text\> \| `NONE` \| `NOASSERTION`<br>where "DocumentRef-"`[idstring]`":" is an optional reference to an external SPDX document as described in [6.6](document-creation-information.md#6.6)<br>where `SPDXID` is a string containing letters, numbers, `.` and/or `-`. to designate SPDX document/package/file as described in [6.3](document-creation-information.md#6.3), [7.2](package-information.md#7.2), [8.2](file-information.md#8.2).<br>where `<single line of text>` is the intended usage and license conditions assumed by the package supplier.<br>where `NONE` can be used to explicitly indicate there are NO other "usage scope".<br>where `NOASSERTION` can be used to explicitly indicate it is not clear if there are "usage scope" that may apply or not. |
+
+<br>
+
+### 31.1.2 Intent
+
+To describe detail intended usage and/or license conditions assumed by the package supplier in a manner that supplements those conditions contained in the software.
+
+### 31.1.3 Examples
+
+EXAMPLE 1 Tag: `UsageScope:`
+
+```text
+UsageScope: SPDXRef-a_specific_binary_verNNN: shipped for consumer product/ medical system / automotive system
+```
+
+EXAMPLE 2 RDF: Property `spdx:UsageScope` in any `spdx:SpdxDocument`, `spdx:Package` or `spdx:File`
+
+```text
+    <UsageScope>
+      <UsageScopeElement>
+        <File rdf:about="http://spdx.org/spdxdocs/spdx-tools-v1.2-3F2504E0-4F89-41D3-9A0C-0305E82..."
+      </UsageScopeElement>
+      <Usage>SPDXRef-a_specific_binary_verNNN: shipped for consumer product/ medical system / automotive system
+      </Usage>
+    </UsageScope>
+  ...
+
+```
+
+## 31.2 Usage Scope Comment field <a name="31.2"></a>
+
+### 31.2.1 Description
+
+This field provides a place for the SPDX document creator to record any general comments about the "usage scope".
+
+Table 31.2 — Metadata for the usage scope comment field
+
+| Attribute | Value |
+| --------- | ----- |
+| Required | No |
+| Cardinality | 0..1 |
+| Format | Free form text that may span multiple lines, refers only to the immediately preceding usagescope.  |
+
+
+### 31.2.2 Intent
+
+To describe prerequisite assumptions of usage scope by the package supplyer.
+
+### 31.2.3 Examples
+
+EXAMPLE 1 Tag: `UsageScopeComment:`
+
+```text
+UsageScope: SPDXRef-a_specific_binary_verNNN: shipped for consumer product/ medical system / automotive system
+UsageScopeComment: A_specific_binary_verNNN is compiled and velified as production level quality
+```
+
+EXAMPLE 2 RDF: Property `rdfs:comment` in class `spdx:UsageScope`
+
+```text
+<UsageScope>
+         <rdfs:comment> A_specific_binary_verNNN is compiled and velified as production level quality</rdfs:comment>
+</UsageScope>
+```
+
+
+## 31.3 UsageScope SPDX identifier field <a name="31.3"></a>
+
+### 31.3.1 Description
+
+Uniquely identify any element in an SPDX document which may be referenced by other elements. These may be referenced internally and externally with the addition of the SPDX document identifier. The metadata for the SPDX identifier field is shown in Table NN.
+
+**Table 31.3 — Metadata for the UsageScope SPDX identifier field**
+
+| Attribute | Value |
+| --------- | ----- |
+| Required | Yes |
+| Cardinality | 1..1 |
+| Format | "SPDXRef-"`[idstring]` <br> where `[idstring]` is a unique string containing letters, numbers, `.`, and/or `-`. |
+
+### 31.3.2 Intent
+
+There may be several usecase of the same package within an SPDX document. Each element needs to be able to be referred to uniquely so that relationships between elements can be clearly articulated.
+
+### 31.3.3 Examples
+
+EXAMPLE 1 Tag: `UsageScopeSPDXID:`
+
+```text
+UsageScopeSPDXID: SPDXRef-1
+```
+
+EXAMPLE 2 RDF: The URI for the element will follow the form:
+
+```text
+[SPDX document namespace]#[SPDX identifier]
+```
+
+See [6.5](document-creation-information.md#6.5) for the definition of the SPDX document namespace and [6.3](document-creation-information.md#6.3) for the definition of the SPDX identifier
+
+Using `xml:base`:
+
+```text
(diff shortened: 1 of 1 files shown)


chapters/usage-profile.md
```diff
@@ -0,0 +1,244 @@
+# 31 Usage Profile section
+
+## 31.1 Usage Scope field <a name="31.1"></a>
+
+### 31.1.1 Description
+
+Software packages are sometimes generated with specific Build conditions based on package supplier's license conditions and intended usage in addition to original software source code. In those cases, it prefer to package supplier clearly state into the SPDX document that the intended usage of the package as envisioned when setting the Build conditions by the package supplier.
+In other cases, some specific licenses out of dual or multiple licenses of original software source code may be applied to the package according to the Build conditions. Or, the inspection conditions for packages may also have to be clearly stated for each package.
```
goneall, line 8: Minor grammar suggestion:
`In those cases, it prefer to package supplier clearly state into the SPDX document that the intended usage of the package as envisioned when setting the Build conditions by the package supplier.`

->

`In those cases, it is preferred for the package supplier to clearly state in the SPDX document the intended usage of the package as envisioned when setting the Build conditions.`

chapters/usage-profile.md
```diff
@@ -0,0 +1,244 @@
...
+| Attribute | Value |
+| --------- | ----- |
+| Required | No |
+| Cardinality | 0..1 |
+| Format | Free form text that may span multiple lines, refers only to the immediately preceding usagescope.  |
+
+
+### 31.2.2 Intent
+
+To describe prerequisite assumptions of usage scope by the package supplyer.
```
goneall, line 66: Is the intent more general than this?

Perhaps we could generalize the intent something like:

`To provide additional context.  For example, describing prerequisite assumptions of usage scope by the package supplier.`

Also note spelling of `supplier`.

chapters/usage-profile.md
```diff
@@ -0,0 +1,244 @@
...
+| Attribute | Value |
+| --------- | ----- |
+| Required | Yes |
+| Cardinality | 1..1 |
+| Format | "SPDXRef-"`[idstring]` <br> where `[idstring]` is a unique string containing letters, numbers, `.`, and/or `-`. |
+
+### 31.3.2 Intent
+
+There may be several usecase of the same package within an SPDX document. Each element needs to be able to be referred to uniquely so that relationships between elements can be clearly articulated.
+
```
goneall, line 103: suggest changing usecase -> cases

chapters/usage-profile.md
```diff
@@ -0,0 +1,244 @@
...
+| --------- | ----- |
+| Required | No |
+| Cardinality | 0..1 |
+| Format | `YYYY-MM-DDThh:mm:ssZ`<br>where:<br><ul><li>`YYYY` is year</li><li>`MM` is month with leading zero</li><li>`DD` is day with leading zero</li><li>`T` is delimiter for time</li><li>`hh` is hours with leading zero in 24 hour time</li><li>`mm` is minutes with leading zero</li><li>`ss` is seconds with leading zero</li><li>`Z` is universal time indicator</li></ul> |
+
+
+### 31.5.2 Intent
+
+Description of the built date of the package for strict identification of the prerequisite assumptions of usage scope by the package supplyer.
+It's recoreded from build system tools or date field of the package file.
```
goneall, line 192: Note spelling of supplier.

I'm not sure we want to add the "how" in the intent.  Could remove the last line or change it to more of a suggestion or "should".  ie: `It should recorded from build system tools or date field of the package file.`

Also - note spelling of recorded

goneall (changes requested):
A few suggestions on wording.  The RDF looked fine on my initial review.

goneall:
From the SPDX tech call on 10 May 2022: Question on whether this should only be applied to Packages since Package is defined to be a unit of software distributed (or released) rather than to all elements.

Are there known use cases where this would apply to Snippets or Files?

swinslow:
Thank you for this draft @yoshi-i! 

I will take a closer look, but one preliminary comment I have is that I would suggest limiting it to describing the supplier's "intended usage". I would suggest omitting references to licensing and license conditions.

If use of a package has license conditions, those would most properly be reflected either (1) through use of an SPDX License Identifier from the [SPDX License List](https://spdx.org/licenses), or (2) through defining a customer `LicenseRef-` identifier via an [Other Licensing Information](https://spdx.github.io/spdx-spec/other-licensing-information-detected/) section. I'm guessing that for the intended use case here, option 2 will be the more common scenario.

In those cases, the existing license fields would apply. I would _not_ be in favor of adding new fields in a different profile that have overlapping or different meanings with regards to licensing matters.

But I don't expect that's needed anyway, to express the sorts of things it looks like you're wanting to convey here -- e.g. intended scope of usage, release date, etc.

yoshi-i:
@goneall  -san, thank you for suggestions. I've corrected spellings and I agree that the modifications you have pointed out are more appropriate in line with the wordings. 

> Are there known use cases where this would apply to Snippets or Files?
I thought that binary blobs (and other set-up scripts) are delivered as the "Delivery in accordance with the development contract" in the supply chain, not in any specific package format but are raw library files which will be used in target product, occasionally.
And in some case, raw snippets instead of patch files for set-up scripts are delivered later as a rapid fix-up.
So, I prefer to enable this usage scope field not only for the package but also snippets and raw files.

chapters/usage-profile.md
```diff
@@ -0,0 +1,243 @@
...
+
+### 31.3.1 Description
+
+Uniquely identify any element in an SPDX document which may be referenced by other elements. These may be referenced internally and externally with the addition of the SPDX document identifier. The metadata for the SPDX identifier field is shown in Table NN.
+
+**Table 31.3 — Metadata for the UsageScope SPDX identifier field**
+
+| Attribute | Value |
+| --------- | ----- |
+| Required | Yes |
```
zvr, line 95: Adding new _required_ attributes is an incompatible change and should not be in a 2.x specification version

yoshi-i: Thank you for pointing out it. I've changed Required attribute of UsageProfile SPDX ID to "No".

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
+# 31 Usage Profile section
```
swinslow, line 1: Note that SPDX 2.3 doesn't yet have a "profile" concept, and if I'm understanding this correctly, these are all fields that would be included on a Package rather than being in their own section. If that's correct, then I assume this would all go into the [Package Information section](https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/package-information.md) as optional fields rather than in a separate standalone chapter?

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+
+## 31.1 Usage Scope field <a name="31.1"></a>
+
+### 31.1.1 Description
+
+Software packages are sometimes generated with specific Build conditions based on package supplier's license conditions and intended usage in addition to original software source code. In those cases, it is preferred for the package supplier to clearly state in the SPDX document the intended usage of the package as envisioned when setting the Build conditions, and/or the inspection conditions for packages may also have to be clearly stated for each package.
```
swinslow, line 7: With apologies, I'm not sure I really understand what is meant here by "setting the Build conditions, and/or the inspection conditions". I'm not sure a reader of the spec would really understand the purpose of this field or the other "Usage Scope" fields below with how this is currently written. Is it possible to provide more clarity or examples?

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+### 31.1.2 Intent
+
+To describe detail intended usage of the package assumed by the package supplier in a manner that supplements those conditions contained in the software.
+
+### 31.1.3 Examples
+
+EXAMPLE 1 Tag: `UsageScope:`
+
+```text
+UsageScope: SPDXRef-a_specific_binary_verNNN: shipped for consumer product/ medical system / automotive system
```
swinslow, line 31: Note that underscores are not valid characters in SPDX IDs. So here and in other examples below, I assume you'll want to replace those with hyphens / dashes.

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+With this "Usage Scope" field, the intended usage assumed by the package supplier is appended to the generated package in a manner that supplements those conditions contained in the source code.
+
+
+Table 31.1 — Metadata for the usage scope field
+
+| Attribute | Value |
+| --------- | ----- |
+| Required | No |
+| Cardinality | 0..* |
+| Format | ["DocumentRef-"[idstring]":"]SPDXID \<single line of text\> \| `NONE` \| `NOASSERTION`<br>where "DocumentRef-"`[idstring]`":" is an optional reference to an external SPDX document as described in [6.6](document-creation-information.md#6.6)<br>where `SPDXID` is a string containing letters, numbers, `.` and/or `-`. to designate SPDX document/package/file as described in [6.3](document-creation-information.md#6.3), [7.2](package-information.md#7.2), [8.2](file-information.md#8.2).<br>where `<single line of text>` is the intended usage of the package assumed by the package supplier.<br>where `NONE` can be used to explicitly indicate there are NO other "usage scope".<br>where `NOASSERTION` can be used to explicitly indicate it is not clear if there are "usage scope" that may apply or not. |
```
swinslow, line 18: Perhaps this is different from what I was thinking in my comment above, about these being fields on packages. This appears to be set up where the "Usage Scope" field is a reference to an SPDX element that is defined somewhere else -- maybe even in a different SPDX document. Is that the intention?

If so, would this whole "usage section" work better instead as another subcategory of [Annotations](https://spdx.github.io/spdx-spec/annotations/), since that already exists as a way to add additional information on a separately-defined SPDX element? This appears to be partially copying some of the concepts from Annotations, but in a fully separate section.

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+| Attribute | Value |
+| --------- | ----- |
+| Required | No |
+| Cardinality | 0..1 |
+| Format | Free form text that may span multiple lines, refers only to the immediately preceding usagescope.  |
+
+
+### 31.2.2 Intent
+
+To provide additional context. For example, describing prerequisite assumptions of usage scope by the package supplier.
```
swinslow, line 65: What's the difference between the text that is supposed to go into the `UsageScope` field vs. the `UsageScopeComment` field? Given that they are both just unstructured free text fields, would it be easier to just have text in the `UsageScopeComment` field and to have the `UsageScope` contain just the relevant SPDX ID for this section?

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+         <rdfs:comment> A_specific_binary_verNNN is compiled and velified as production level quality</rdfs:comment>
+</UsageScope>
+```
+
+
+## 31.3 UsageScope SPDX identifier field <a name="31.3"></a>
+
+### 31.3.1 Description
+
+Uniquely identify any element in an SPDX document which may be referenced by other elements. These may be referenced internally and externally with the addition of the SPDX document identifier. The metadata for the SPDX identifier field is shown in Table NN.
```
swinslow, line 89: I'm not sure I understand the purpose of this being a separate field -- isn't this duplicative of having the SPDX ID in the `UsageScope` field above?

Also, if this is referring to other SPDX elements, should it include the `DocumentRef-` optional prefix also so that it can refer to SPDX elements defined in other Documents?

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+
+Using document URI:
+
+```text
+<UsageScope rdf:about="http://acme.com/spdxdocs/acmeproj/v1.2/1BE2A4FF-5F1A-48D3-8483-28A9B0349A1B#SPDXRef-1">
+    ...
+</UsageScope>
+```
+
+## 31.4 Package Release Date field <a name="31.4"></a>
```
swinslow, line 138: Would it be appropriate for this field to be moved into the [Package Information section](https://spdx.github.io/spdx-spec/package-information/), rather than kept here?

This seems to be metadata about the package, and even if it's useful for a Usage Information section, it seems like it would belong as an optional field in the Package Information section along with the other Package-specific data.

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+
+EXAMPLE 2 RDF: Property `spdx:packageReleseDate` in class `spdx:UsageScope`
+
+```text
+<UsageScope rdf:about="...">
+    <packageReleaseDate> 2010-01-29T18:30:22Z </packageReleaseDate>
+</UsageScope>
+```
+
+## 31.5 Package Built Date field <a name="31.5"></a>
```
swinslow, line 173: Same comment here as for the Package Release Date field above. This appears to be recording objective metadata about the Package, and seems like it should go in the Package Information section.

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+<UsageScope rdf:about="...">
+    <packageBuiltDate> 2010-01-29T18:30:22Z </packageBuiltDate>
+</UsageScope>
+```
+
+## 31.6 Valid Until Date field <a name="31.6"></a>
+
+### 31.6.1 Description
+
+Identify expiration date of designated "usage scope" descriptions which declared by supplier of the package.
```
swinslow, line 213: I'm not sure I understand what an "expiration date" or "valid until" means in this context.

Is this meant to say that the actual item being described by the SPDX element (e.g. the actual software package) is "expired" or "invalid" after this date? Or is this saying that the SPDX metadata itself (e.g. the metadata in the SPDX document) is "expired" or "invalid" after this date?

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+
+EXAMPLE 2 RDF: Property `spdx:validUntilDate` in class `spdx:UsageScope`
+
+```text
+<UsageScope rdf:about="...">
+    <validUntilDate> 2030-01-29T18:30:22Z </validUntilDate>
+</UsageScope>
+```
+
+## 31.7 Valid Until Event field <a name="31.7"></a>
```
swinslow, line 244: Same question here as for the "Valid Until Date" field above. What does it mean for this to be an "expiration condition," and what it mean for something here to be expiring or become invalid?

chapters/usage-profile.md
```diff
@@ -0,0 +1,326 @@
...
+
+```text
+<UsageScope rdf:about="...">
+    <validUntilEvent>
+         <rdfs:comment> Updated descriptions are delivered</rdfs:comment>
+    </validUntilEvent>
+</UsageScope>
+```
+
+## 31.8 ValidUntil SPDX identifier field <a name="31.8"></a>
```
swinslow, line 284: Same question here as for UsageScopeSPDXID above -- I'm not sure I understand the purpose of this being a separate field; isn't this duplicative of having the SPDX ID in the UsageScope field above?

swinslow (changes requested):

chapters/usage-profile.md
yoshi-i: Following examples are considered as a premise for describing this fields as a intention of package supplier, to a consumer downstream in the supply chain can mechanically process and retrieve the upstream provider's intent from the SBOM file included in the package.

- "Binary file A contained in this package was built with compile-time option X to use on product Y" 
- "This Package is initial Alpha release, limited verification are applied, do not use in the final product"
- "file A is library file for product X and Y", "script file B is configuration file for library A on product X" and "script file C is configuration file for library A on product Y".

swinslow:
Hi @yoshi-i, thanks very much for the responses to my earlier comments.

As I mentioned on the spdx-tech mailing list, my main objective for this was to make sure that the licensing-related content was removed. I appreciate you having made those changes.

My other comments here were intended to help with clarifying or improving the remainder. I admit that I don't fully understand some of the practical details of how these fields would be used.

But in full honesty, I'm not likely to be a regular user or consumer of these fields. So I'll withdraw any of my remaining comments, and I'll let the other participants weigh in on whether they would like further changes based on my earlier comments or whether they are comfortable with the PR as-is.

Thank you again for the changes you've made to address my feedback!

yoshi-i:
@swinslow -san, I appreciate your quite important comments. 
I would like to encourage future discussions to consider making the field more user-friendly for a wider range of uses, not just some sort of specific business uses.

chapters/usage-profile.md
kestewart: It is referring to the period of support.   Example is the kernel LTS is supported by the community until a specific date.   Automated checks can then monitor when these dates are known.

kestewart:
@yoshi-i - We're going to need to wait until 3.0 to add the full usage profile, but some of the fields described in this PR make sense to pull into 2.3 and let folks start using right now.    I've created https://github.com/spdx/spdx-spec/pull/709 to include these fields.   Can you please review that PR and comment on it.   Note:  I removed the explicit 
"Package" in some field names, as the AI BOM profile will want to use some of them as well at the "File" level in 3.0.

kestewart:
Changing the milestone on this to 3.0.    Note 3 of the fields have been included in 2.3 as part of https://github.com/spdx/spdx-spec/pull/709

kestewart:
Usage profile and Operations need to be reconciled.   Moving this to 3.1

kestewart:
This is stale at this point.   I'm going to close it for now.  @yoshi-i - if you disagree, please feel free to reopen.

Issue #605, opened 2021-11-11
Clause 10: which is better example, LicenseRef-Whiskeyware or LicenseRef-Beerware-4.2


fu7mu4:
In 10.1.3 Examples of current SPDX, the EXAMPLE LicenseID: is  LicenseID: `LicenseRef-Beerware-4.2`
But 10.2.3 Examples the EXAMPLE 2 RDF is  `LicenseRef-Whiskeyware` :

```xml
<ExtractedLicensingInfo rdf:about="licenseRef-Whiskeyware">
    <licenseId>LicenseRef-Whiskeyware</licenseId>
    <extractedText>This software is licensed under the WHISKEY-WARE LICENSE.</extractedText>
</ExtractedLicensingInfo>
```

I think `LicenseRef-Whiskeyware` is better than `LicenseRef-Beerware-4.2` for the example LicenseID in 10.1.3 because the `Beerware License` has already its identifier , see https://spdx.org/licenses/Beerware.html

see https://spdx.github.io/spdx-spec/other-licensing-information-detected/

seabass-labrax:
Good point, @fu7mu4! I'd agree that it makes sense to choose a license that isn't in the License List when demonstrating the use of `LicenseRef-`.

PR #674, opened 2022-05-10
Use 'Whiskeyware' license for LicenseRef example

seabass-labrax:
Resolves issue https://github.com/spdx/spdx-spec/issues/605.

Signed-off-by: Sebastian Crane <seabass-labrax@gmx.com>


kestewart (approved):

seabass-labrax:
Note: there are still some example SPDX documents in the `examples/` directory which use the 'beerware' license.

tsteenbe (changes requested):
Suggest we use any of the LicenseRef-scancode-  SPDX on https://scancode-licensedb.aboutcode.org/ instead of a fake license

kestewart:
From call,  bias to just go with whiskeyware for now.    Add in something on top to indicate proprietary. 

Motivation - show how to do a commercial license?   See scancode.   Can put any license.

PR #675, opened 2022-05-11
Fix gh pages

NorioKobota:


NorioKobota:
So sorry, this is a mistake to send a PR..

Issue #661, opened 2022-04-21
Mkdocs are failing in the CI


goneall:
We're getting the following error when the CI runs (see PR #659 for an example):

```
Traceback (most recent call last):
  File "/usr/local/bin/mkdocs", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1[6](https://github.com/spdx/spdx-spec/runs/6116293939?check_suite_focus=true#step:5:6)5[7](https://github.com/spdx/spdx-spec/runs/6116293939?check_suite_focus=true#step:5:7), in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/mkdocs/__main__.py", line 1[8](https://github.com/spdx/spdx-spec/runs/6116293939?check_suite_focus=true#step:5:8)7, in build_command
    build.build(config.load_config(**kwargs), dirty=not clean)
  File "/usr/local/lib/python3.[10](https://github.com/spdx/spdx-spec/runs/6116293939?check_suite_focus=true#step:5:10)/site-packages/mkdocs/config/base.py", line 216, in load_config
    from mkdocs.config.defaults import get_schema
  File "/usr/local/lib/python3.10/site-packages/mkdocs/config/defaults.py", line 1, in <module>
    from mkdocs.config import config_options
  File "/usr/local/lib/python3.10/site-packages/mkdocs/config/config_options.py", line 8, in <module>
    from mkdocs import utils, theme, plugins
  File "/usr/local/lib/python3.10/site-packages/mkdocs/theme.py", line 6, in <module>
    from mkdocs.utils import filters
  File "/usr/local/lib/python3.10/site-packages/mkdocs/utils/filters.py", line [13](https://github.com/spdx/spdx-spec/runs/6116293939?check_suite_focus=true#step:5:13), in <module>
[...]

PR #676, opened 2022-05-11
Fix gh pages v2.3

NorioKobota:
Relates to issue #661.
Please refer to this site: https://github.com/mkdocs/mkdocs/releases/tag/1.2.4
Due to the breaking chage in jinja2, you need to upgrade your mkdocs version to 1.2.4 or higher.
Note:
Depending on the `uses: actions/checkout@v2` specification, you may also need to update the requirements.txt that exists in v3.0 and master branch.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,6 +1,6 @@
 on: 
   push:
-    branches: [master, development/v2.2.2, development/v3.0]
+    branches: [master, development/v2.3, development/v3.0]
 
 jobs:
   build_current:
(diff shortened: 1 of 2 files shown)


goneall:
Thanks @NorioKobota 

@tsteenbe @iamwillbar Looks like this resolves the issue - at least the CI runs without complaining.  I'll merge tomorrow unless you see any issues with the fix.

goneall (approved):

NorioKobota:
@goneall Thank you for accepting the PR.
As you can see the log here: https://github.com/spdx/spdx-spec/actions/runs/2315171762, you also have to update mkdocs version in requirements.txt that exists in v3.0 and master(main) branch.
I do not know how each branch, especially the master branch, is operated, so I would appreciate it if you could fix it yourself.
Best,
 -- kobota

PR #629, opened 2022-03-03
SPDX Lite: Add Package Supplier field

hfukuchi:
Added/inserted Package Supplier field to SPDX Lite as L2.5

PR #678, opened 2022-05-12
Retry of Added Package Supplier field to SPDX Lite (#629)

hfukuchi:
Added/inserted Package Supplier field to SPDX Lite as L2.5
Added signed off to the previous pull request and re-push it.

#629 will be cancelled.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,6 +1,6 @@
 on: 
   push:
-    branches: [master, development/v2.2.2, development/v3.0]
+    branches: [master, development/v2.3, development/v3.0]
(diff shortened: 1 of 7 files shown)


kestewart:
Reviewed the 2.3 sources, and the changes for SPDX-Lite profile(https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/SPDX-Lite.md), as well as the fixes for the document creation (https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/document-creation-information.md)  are included as well.   This PR is against 2.2.2 which is already released.   Closing.

PR #686, opened 2022-05-18
Same text in tag:value and RDF examples

vargenau:
In `other-licensing-information-detected.md`, add a full stop at the end of the sentence, so that the text is the same in tag:value and RDF.

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>


goneall (approved):
@vargenau Good catch.

kestewart (approved):
Thanks for catching this!

PR #700, opened 2022-05-25
Update the gitbom section

AevaOnline:
- change section identifier to F.4.2
- add second example
- add link to IANA format reference
- expand Documentation string


chapters/external-repository-identifiers.md
```diff
@@ -238,15 +238,19 @@ Examples:
 
 External Reference Site: [Software Heritage persistent IDentifiers](https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html)
 
-### F.4.1 gitbom <a name="gitbom"></a>
+### F.4.2 gitbom <a name="gitbom"></a>
 
-Locator Format: gitoid":"\<git object type>":"\<hash algorithm>":"\<hash value>
-	
-Contextual Example: gitoid:blob:sha1:261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64
+Locator Format: `gitoid":"\<git object type>":"\<hash algorithm>":"\<hash value>`
```
zvr, line 243: Should this start with `"gitoid:"` ? I mean, the `gitoid` part is standard text, it should be in quotes.

chapters/external-repository-identifiers.md
```diff
@@ -238,15 +238,19 @@ Examples:
...
+
+Locator Format Reference: <https://www.iana.org/assignments/uri-schemes/prov/gitoid>
```
zvr, line 245: While an external reference is useful, we should have the details in this document:

- Valid values for <git object type> are `blob`, `tree`, `commit`, and `tag`.
- Valid values for <hash algorithm> are 'sha1' and 'sha256'
- <hash value> should be expressed as a hexadecimal string in lower case.

chapters/external-repository-identifiers.md
AevaOnline: Thanks. For this line, I was merely copying what was present, and you're correct, "gitoid:" is a literal string.

chapters/external-repository-identifiers.md
AevaOnline: That's a good point. I'll update.

jeff-schutt:
Thanks for the review and updates @AevaOnline and @zvr! Definite improvements.

@tsteenbe @kestewart can you merge this into 2.3 before release?

chapters/external-repository-identifiers.md
```diff
@@ -238,15 +238,24 @@ Examples:
...
+
+Current valid values are:
+* `<git object type>`: `blob`, `tree`, `commit`, `tag`
+* `<hash algorithm>`: `sha1`, `sha256`
+* `<hash value>` should be expressed as a hexadecimal string in lower case
+
+Contextual Examples:
+* `gitoid:blob:sha1:261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64` is the gitbom identifier of a software artifact using the SHA1 algorithm
+* `gitoid:bom:sha256:3557f7eb43c621c71483743d4b37059bb80933e7f71277c0c3b3846159d1f61c` is the gitbom identifier of an artifact dependency graph using the SHA256 algorithm
```
goneall, line 254: This example has an object type of `bom` which is not one of the `<git object type>` listed above (`blob`, `tree`, `commit`, `tag`)

goneall (changes requested):
I noticed one inconsistency in the example used below.

chapters/external-repository-identifiers.md
edwarnicke: One small note: Locators are not identifiers.  Shouldn't this be 'Identifier Format' ?

chapters/external-repository-identifiers.md
edwarnicke: Thoughts on :

`gitbom:gitoid:blob:sha256:3557f7eb43c621c71483743d4b37059bb80933e7f71277c0c3b3846159d1f61c`

chapters/external-repository-identifiers.md
```diff
@@ -238,15 +238,41 @@ Examples:
...
+Contextual Examples:
+* `gitoid:blob:sha1:261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64` is the git identifier of a software artifact using the SHA1 algorithm
+* `gitoid:blob:sha256:3557f7eb43c621c71483743d4b37059bb80933e7f71277c0c3b3846159d1f61c` is the git identifier of a software artifact using the SHA256 algorithm
+
+Documentation: `gitoid` stands for Git Object ID.  Git relies on a Merkle Tree to index stored objects. See <https://git-scm.com/book/en/v2/Git-Internals-Git-Objects>.
+
+### F.4.3 gitbom <a name="gitbom"></a>
+
+Locator Format: 
+`"gitbom:"\<git object type>":"\<hash algorithm>":"\<hash value>`
```
edwarnicke, line 261: Would it make sense to make this

`"gitbom:"\<gitoid uri>`

where `\<gitoid uri>` is defined above?

chapters/external-repository-identifiers.md
```diff
@@ -238,15 +238,41 @@ Examples:
...
 External Reference Site: [Software Heritage persistent IDentifiers](https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html)
 
-### F.4.1 gitbom <a name="gitbom"></a>
+### F.4.2 gitoid <a name="gitoid"></a>
 
-Locator Format: gitoid":"\<git object type>":"\<hash algorithm>":"\<hash value>
-	
-Contextual Example: gitoid:blob:sha1:261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64
+Locator Format: `"gitoid:"\<git object type>":"\<hash algorithm>":"\<hash value>`
+
```
edwarnicke, line 244: You may want to look at this in preview, as it appears that the `<` does not require escaping in a pre block.

zvr:
I am confused...

This seems to define two formats, `gitoid` and `gitbom` which are exactly the same -- to the point that the GitBOM section points to the Gitoid IANA scheme (which uses `gitoid`).

Is this a copy-paste error?

edwarnicke:
> I am confused...
> 
> This seems to define two formats, `gitoid` and `gitbom` which are exactly the same -- to the point that the GitBOM section points to the Gitoid IANA scheme (which uses `gitoid`).
> 
> Is this a copy-paste error?

@zvr Let me try and see if I can clarify :)

It comes down to the disambiguating of 'identifier' from 'kind'.

A gitoid *identifies* literally any array of bytes ([]byte).   By default in SPDX we are presuming the kind of the thing being identified is 'software artifact'.  So a bare gitoid URI identifies a software artifact.

But there is another interesting 'kind' of thing you may want to identify: the GitBOM doc that expresses the *relationship* between software artifacts as captured by the build tools themselves (ie: canonical).  In order to indicate that kind, one prefixes the gitoid with `gitbom:`

So `gitoid:blob:sha1:261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64` would be interpreted as an identifier of a 'software artifact' and `gitbom:gitoid:blob:sha1:261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64` would be interpreted as an identifier of a 'GitBOM doc'.

This was based on our understanding that the identifiers for use in an ExternalReference that itself, in SPDX 2.3, would not have a means to indicate the 'kind' of the external reference.  Please note: we'd be *delighted* to be corrected if we are mistaken on that point :)

Does that help clarify?

AevaOnline:
Updated PR based on Ed's comments and our conversation this morning.

After lengthy consideration, I do not think that `gitbom:gitoid:<type>:<aglo>:<hash value>` is a good Locator Reference, since `gitbom` is not the Locator Type. Regardless of whether the 'kind' of artifact referenced is a binary artifact or the GitBOM document of that software artifact, the Locator Type is the `gitoid` -- the git-compatible hash of that file's bytes.

I still believe a second type, of `gitbom:<type>:<algo>:<hash value>` would be ideal here, but this is not an IANA-registered type at the moment, and to call this a valid Locator Reference Type, I would also want to have a canonical location to search for `gitbom` documents by their identifier. The community has seen some very early proposals for this, but nothing material yet. 

I hope we can improve this ambiguity in SPDX 3.x

edwarnicke:
@goneall Its been too long! :)

chapters/external-repository-identifiers.md
```diff
@@ -238,15 +238,33 @@ Examples:
...
-### F.4.1 gitbom <a name="gitbom"></a>
+### F.4.2 gitoid <a name="gitoid"></a>
 
-Locator Format: gitoid":"\<git object type>":"\<hash algorithm>":"\<hash value>
-	
-Contextual Example: gitoid:blob:sha1:261eeb9e9f8b2b4b0d119366dda99c6fd7d35c64
+Locator Format:
+
+```
+"gitoid:"<git object type>":"<hash algorithm>":"<hash value>
```
jeff-schutt, line 246: Looks like the scheme syntax was changed with the reformat from inline code to code on a new line. Updating to ensure consistency w/ the scheme in the IANA link
```suggestion
gitoid":"<git object type>":"<hash algorithm>":"<hash value>
```

jeff-schutt (changes requested):
Minor tweak, otherwise looks good. Thanks all :)

chapters/external-repository-identifiers.md
zvr: No, I made a comment before. The literal string is `"gitoid:"`, not just the colon.

edwarnicke: I tend to concur here, the literal string is "gitoid:".

jeff-schutt: Got it. Looks like the gitoid scheme in the IANA URI filing is what should change. Good catch, @zvr

jeff-schutt:
LGTM

goneall (approved):

goneall:
Once @kestewart confirms her review comments were addressed, I think we can merge this

kestewart (approved):
Looks good,  thanks for clarifying!

PR #701, opened 2022-05-31
publish: Only run branch jobs when PR the branch

tsteenbe:
publish: Only run branch jobs when PR the branch


goneall:
@tsteenbe Do we need to merge this into the 2.3 branch as well?

tsteenbe:
@goneall  Yes the same change needs to also done for 2.3

PR #639, opened 2022-03-15
Typo: fix missing hyphen in 7.4.3 RDF Example 2

rnjudge:
This commit fixes a missing hyphen typo in RDF Example 2 in section
7.4.3.

Fixes #604

Signed-off-by: Rose Judge <rjudge@vmware.com>

PR #703, opened 2022-06-01
Document `revision_tag_or_branch`

### LINKED ISSUE #636: Download location should recommmend a fully qualified git ref, not a bare tag or branch name
In git, a bare `<refname>` (e.g. `main`) can refer to a tag or a branch. If the name refers to both, it is ambiguous what it refers to; see [gitrevisions(7)](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitrevisions.html). Some tools use the tag, while others (e.g. `git checkout`) use the branch.

In the SPDX "Package download location field", the field `[@<revision_tag_or_branch>]` is undocumented. For git, most reader would assume it means a bare tag or branch name, e.g. `main` or `v1.2.3`. Unfortunately that is ambiguous as noted above.

In order to make the download location fully unambiguous, I suggest explaining this field and recommending that the full `refs/heads/<branch>` or `refs/tags/<tag>` be used, rather than a bare `<branch>` or `<tag>`. This avoids a problem in the case where someone accidentally (or intentionally!) creates a tag that shadows a branch, or vice versa.

ivanayov:
Using only `<tag>` or `<branch>` for `<revision_tag_or_branch>` is
ambiguous as the gitrevisions manual suggest. This is why it needs
to be documented with appropriate advice.

Fixes #636

Signed-off-by: Ivana Atanasova <iyovcheva@vmware.com>

Applying a markdown copy of the change, as it's not very convenient for review:

[prev_paragraph]<br><br>For git, `<revision_tag_or_branch>` is a `<refname>` or `<sha1>` as described by [gitrevisions(1)](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitrevisions.html). It is RECOMMENDED to use the unambiguous `refs/heads/<branch>` or `refs/tags/<tag>`, rather than only `<branch>` or `<tag>`, in order to prevent issues caused by shadowing.<br><br>[next_paragraph]


goneall:
@MarkLodato could you review the changes and comment if it resolves issue #639?

@iamwillbar - would you mind doing a quick review as well?

Thanks

chapters/package-information.md
```diff
@@ -295,7 +295,7 @@ The metadata for the package download location field is shown in Table 19.
...
...ike `refs/heads/<branch>` or `refs/tags/<tag>`, rather than only `<branch>` or `<tag>`, in order to prevent issues caused by shadowing (see [gitrevisions](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitrevisions.html) manual page). <br><br>This VCS location compact notation (inspired and mostly adopted from [pip][pip-vcs] as of 2015-02-20) supports referencing locations in version control systems such as [Git][], [Mercurial][], [Subversion][] and [Bazaar][], and specifies the type of VCS tool using url prefixes: `git+`, `hg+`, `bzr+`, `svn+` and specific transport schemes such as SSH or HTTPS.<br><br>Specifying sub-paths, branch names, a commit hash, a revision or a tag name is recommended, and supported using the `@` delimiter for commits and the `#` delimiter for sub-paths.<br><br>Using user names and password in the `<host_name>` is not supported and should be considered as an error. User access control to URLs or VCS repositories shall be handled outside of an SPDX document.<br><br>In VCS location compact notations, the trailing slashes in `<host_name>`, `<path_to_repository>` are not significant. Leading and trailing slashes in `<sub_path>` are not significant. |
```
MarkLodato, line 298: Suggestion:

For git, `<revision_tag_or_branch>` is a `<refname>` or `<sha1>` as described by [gitrevisions(1)](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitrevisions.html). It is RECOMMENDED to use the unambiguous `refs/heads/<branch>` or `refs/tags/<tag>`, rather than only `<branch>` or `<tag>`, in order to prevent issues caused by shadowing.

MarkLodato, line 298: Q: Should this allow any other form of `<rev>`? I believe the following are the only ones that make sense:

* `<describeOutput>`, `<rev>~[<n>]`, and/or `<rev>^[<n>]` would allow someone to symbolically reference a commit that does not have a tag or branch pointing directly to it. On the one hand, that might be valuable in some cases, since it is more descriptive than a raw commit. On the other, it might be better to push people toward simplicity.

* `<rev>:<path>` would allow referencing a particular tree or blob within the repository, rather than the full commit. I don't know if this is desirable or not. (We'd have to disallow `<path>` starting with `./` or `../` since those are relative to the current working directory, which is undefined.)

* `<rev>^{<type>}` / `<ref>^{}` would allow referencing a particular object type. This *might* be useful if one wants to say that only the tree mattered, not the full commit, though I'm not sure anyone would ever use this in practice.

I suspect we do NOT want to allow the following:

* `...@...`, `:[<n>:]<path>`: relative to the local copy of the repository (HEAD, reflog, branch configuration, or index)
* `<rev>^{/<text>}`: too complex; can be described by one of the supported values instead.
* `:/<text>`: both problems combined!

MarkLodato (commented):
Thanks! I made a suggestion to make this more explicit.

chapters/package-information.md
ivanayov: Thank you, I've updated with your proposal.

MarkLodato (approved):
(Pending the other open question, but either way is OK by me.)

goneall:
@ivanayov @MarkLodato I have no opinion on the open question.  If both of you are OK with the current state, I can merge this in for the 2.3 release.

@ivanayov do you think this is ready to merge?

ivanayov:
I suggest we merge this and continue the discussion in a follow-up, once we agree on.

chapters/package-information.md
ivanayov: Regarding `<describeOutput>`, `<rev>~[<n>]`, and/or `<rev>^[<n>]`, I think we would definitely like to allow, given that: 
* contents of a branch might change over time
* in case the build is cut-off on such a reference, it's better if the sbom reflects that

`<rev>:<path>` - not sure if this should be encouraged. Do you know any good example? The only reason to allow that IMO is if this is anyway how the build is created and we want that information to be visible. Which in other hand might be a good enough reason to allow?
Same with `<rev>^{<type>}` / `<ref>^{}` - I cannot think of a practical use-case to reference a build to that, but if you know any example, this might make sense.

Absolutely agree on the last 3.

goneall:
I'll merge this in - as @ivanayov suggested above, we can open a new issue or PR to continue the discussion on further improvements.

PR #707, opened 2022-06-06
Fix nondisplayable characaters

vargenau:
Replace characters that cannot be displayed by quotation marks and Copyright symbol in file SPDXTagExample-v2.2.spdx.

Signed-off-by: Marc-Etienne Vargenau <marc-etienne.vargenau@nokia.com>


goneall (approved):
Thanks @vargenau

goneall:
I'm wondering if this same issue is in the other formats (e.g. JSON).  I haven't had a chance to look - but if it is, we can either edit directly or run the converter on the corrected tag/value file.

kestewart (approved):

vargenau:
> I'm wondering if this same issue is in the other formats (e.g. JSON). I haven't had a chance to look - but if it is, we can either edit directly or run the converter on the corrected tag/value file.

I have updated the other formats.

goneall (approved):
Thanks @vargenau - I double checked the encoding and all looks good.

Issue #705, opened 2022-06-03
https://spdx.github.io/spdx-spec/v2-draft/ should point to the v2.3 spec


goneall:
It is currently showing v2.2.2.

This may have been broken with one of the recent PR's:

PR #702 PR #701 PR #699 PR #698 PR #697 PR #693 PR #692 and/or PR #676 

@NorioKobota @tsteenbe - would you mind taking a look and see what went wrong?

NorioKobota:
@goneall @tsteenbe 
Thank you. 

https://github.com/spdx/spdx-spec/blob/development/v2.3/.github/workflows/publish.yml#L8
https://github.com/spdx/spdx-spec/blob/development/v2.3/.github/workflows/publish.yml#L28
Only branches that have changed are now built.

https://github.com/spdx/spdx-spec/blob/development/v2.3/.github/workflows/publish.yml#L49
However, to publish as a website, both master and v2.3 branches must be successfully built.

If we modify it so that it is deployed if either build succeeds, it should work correctly, but I'm not sure a bit without checking.

goneall:
Turns out that this does get updated for some merges, but not updated for others.

Below is a screenshot from the Github action:

![image](https://user-images.githubusercontent.com/378172/175119549-32f423f8-e800-42c4-a559-be5b9bea1d2f.png)

NorioKobota:
As far as I checked, it is not possible to build each branch independently and create a website.
It is difficult to explain in words for me, so I will put pictures.

The following configuration will attempt to create a website independently, but it will cause problems as shown in the attached picture.
https://github.com/NorioKobota/spdx-spec/blob/fix-705-default/.github/workflows/publish.yml

If someone commits to development/v2.3, GithubActions will create a website as follows:

![image](https://user-images.githubusercontent.com/8510920/175471544-978c9bdd-4fc3-4259-a337-69ed9ff35516.png)
about 'destination_dir': https://github.com/peaceiris/actions-gh-pages/issues/324 

Then when someone commits to the main branch, the v2-draft part is removed.

![image](https://user-images.githubusercontent.com/8510920/175473130-5f6bcd62-dd27-423c-a726-672b922912c0.png)

NorioKobota:
Reverting following two commits will solve the problem, but I think there is some intention, so I will leave it to you.
https://github.com/spdx/spdx-spec/commit/7ba7bf571c0c3c3fd6a4bd780914d58f9274adcc
[...]

PR #710, opened 2022-06-12
Change to not to use requirements.txt for fixing #705

NorioKobota:
If requirements.txt exists in a separate file, we have to modify all monitored branches if the dependent libraries in requirements.txt need to be upgraded.
Therefore, I re-documented all dependent library installations in the publish.yml.
With this change, from now on we only need to maintain and update the publish.yml of the default branch for the generation of github pages.
Signed-off-by: Norio Kobota <norio.kobota@sony.com>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,7 +5,6 @@ on:
 jobs:
   build_current:
     name: Build current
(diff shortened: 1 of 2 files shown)


NorioKobota:
requiremets.txt is used by another action.. I'll withdraw this temporarily.

PR #711, opened 2022-06-12
Change to not use requirements.txt

NorioKobota:
Fix issue #705 
If requirements.txt exists in a separate file, we have to modify all monitored branches if the dependent libraries in requirements.txt need to be upgraded. Therefore, I re-documented all dependent library installations in the publish.yml.
With this change, from now on we only need to maintain and update the publish.yml of the default branch for the generation of github pages.
Signed-off-by: Norio Kobota <norio.kobota@sony.com>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,7 +5,6 @@ on:
 jobs:
   build_current:
     name: Build current
-    if: github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     container: python:3
     steps:
@@ -13,7 +12,7 @@ jobs:
       with:
         ref: 'master'
     - name: Install pre-requisites
-      run: pip install -r requirements.txt
+      run: pip install mkdocs==1.3.0 json-schema-for-humans==0.39.5
     - name: Build
       run: mkdocs build -v --clean
     - name: Schema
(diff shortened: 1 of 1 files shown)


goneall:
thanks @NorioKobota 

@tsteenbe  - could you do a quick review / merge?

.github/workflows/publish.yml
```diff
@@ -5,15 +5,14 @@ on:
 jobs:
   build_current:
     name: Build current
-    if: github.ref == 'refs/heads/master'
```
seabass-labrax, line 8: This reverts commit ed94fa4f74eae7a5486892375aa87f13270f9dd8 - is this intentional?

NorioKobota: @seabass-labrax 
Yes. 
https://github.com/spdx/spdx-spec/blob/development/v2.3/.github/workflows/publish.yml#L49
In line 49, It was very difficult for me to write logic to determine whether a build of the master branch succeeded if it was committed/pushed to the master branch, and whether a build of the v 2.3 branch succeeded if it was committed/pushed to the v2.3 branch. Therefore, I took the approach (reverted) that if a modification was committed/pushed to either branch, I would build both and reconfigure the website.
However, since I have listed the dependent pip packages in publish.yml, the only file we need to modify for future library updates is publish.yml in the default branch.

tsteenbe:
@NorioKobota  Althought I understand the motivation for this change, I deliberately used a requirements.txt instead including it in the publish.yml - without requirements.txt people can not build the spec locally.

PR #713, opened 2022-06-13
Update Spec Version to 2.3, fix comma typo

davaya:


goneall (approved):

goneall:
Thanks @davaya - good catch.  Could you add a sign-off in the commit record (see https://github.com/spdx/spdx-spec/runs/6869800622?check_suite_focus=true)

davaya:
If I'm reading this correctly, the only way I can fix the check is to do a dangerous rebase.

Instead, you could close this PR and add the comma yourself.  Much safer.  I don't have any pride of ownership or intellectual property claim on that comma; it's all yours.

goneall:
@tsteenbe Any issue if I override and accept this PR?

goneall:
I found another typo or copy/paste error - so I included this change in PR #717 

Closing this PR.

Issue #606, opened 2021-11-11
Just a rant re: JSON


rjb4standards:
I'm starting to parse JSON spdx and am finding it a little irritating that the tag names don't match the same case as TV tag names. It sure would be nice to have one consistent spdx data model and tag name taxonomy that is applied consistently across all formats. This would really simplify data searches within an SPDX SBOM.

goneall:
@rjb4standards Having implemented some of the translation tools between the different formats, I understand the frustration.

This was a conscious decision to align with the naming conventions of the serialization format rather than keep them consistent between the formats which we believe will lead to better adoption within the users of that serialization format.

Although I support the decision to have differences in case between the serialization formats, I believe these differences should be formalized and documented in the spec.  Something I think will get addressed in 3.0 with our approach of generating the serialization format specifications.

rjb4standards:
Thanks for your reply Gary. I'll just learn to live with the differences. Fortunately I only need a subset of the total taxonomy for my C-SCRM use case. 
Are there any plans to produce a Vulnerability Disclosure Report format that correlates with an SBOM, i.e. reports vulnerability search results at the component level?

goneall:
> Are there any plans to produce a Vulnerability Disclosure Report format that correlates with an SBOM, i.e. reports vulnerability search results at the component level?

There is a security related profile in the works led by @tsteenbe in the works - contributors welcome :)

rjb4standards:
Thanks, Gary. Much appreciate all your help.

@tsteenbe Thomas, can you point me in the right direction for this work. FYI: I created an open-source Vulnerability Disclosure Report (VDR) in order to satisfy the requirements of the SBOM Bill, H.R. 4611 - am interested in exchanging info. Thanks, Dick Brooks, REA

kestewart:
@goneall - need to update JSON & XML schemas, and automatic drawings.

goneall:
Resolved with PR #716

Issue #459, opened 2020-07-10
Update OWL document with 2.2.1 changes


goneall:
Some of the text descriptions have been updated in 2.2 (see PR #455).

These should be reflected in the OWL, JSON-LD ontology, and schema documents.

goneall:
Several of the issues have been Resolved with PR #716.  A new issue has been created to track any remaining items.

Issue #691, opened 2022-05-24
Meta Issue - Update Schemas for 2.3


goneall:
This issue is to track the work needed to update all the schemas for SPDX release 2.3.

- [x] Resolve all SPDX issues and pull requests which impact the schemas (PR #671)
- [x] Diff the 2.2 and 2.3 versions of the spec to identify changes to the schemas
- [x] Update the OWL document based on the diff
- [x] Generate TTL version of OWL along with RDF (replaces PR #501 )
- [x] Generate JSON Schema
- [x] Review OWL and JSON schemas
- [x] Close issue #606
- [x] Close issue #459 
- [x] Close PR #547
- [x] Close PR #501

goneall:
Note that the following 2.3 PR's and issues are assumed to only impact documentation and will not change impact the schema:
- PR #681 
- PR #679 
- PR #671 
- PR #464 
- Issue #657 
- Issue #630 
- Issue #593 
- Issue #587 
- Issue #582 
- Issue #571

goneall:
> Gary, I've requested that the two pending items from PR 670 be addressed and resolved. See details of the two pending items in the meeting minutes

@rjb4standards Let me know kif this is resolved or something we should discuss on Tuesday's tech call

goneall:
All tasks have been completed

PR #716, opened 2022-06-13
Update schemas for version 2.3

goneall:
See the list of changes in [this comment for issue 691](https://github.com/spdx/spdx-spec/issues/691#issuecomment-1154198748).

Note that not all of the items listed in the comment needed updates.

This PR also copies over fixes from PR #547 

### DIFF
--- a/ontology/spdx-ontology.owl.xml
+++ b/ontology/spdx-ontology.owl.xml
@@ -9,14 +9,15 @@
      xmlns:doap="http://usefulinc.com/ns/doap#"
      xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
     <owl:Ontology rdf:about="http://spdx.org/rdf/terms">
-        <owl:versionIRI rdf:resource="http://spdx.org/rdf/terms/2.2"/>
-        <rdfs:comment>This specification describes the SPDX® language, defined as a dictionary of named properties and classes using W3C&apos;s RDF Technology.
+        <owl:versionIRI rdf:resource="http://spdx.org/rdf/terms/2.3"/>
+        <rdfs:comment xml:lang="en">This specification describes the SPDX® language, defined as a dictionary of named properties and classes using W3C&apos;s RDF Technology.
 
-SPDX® is a designed to allow the exchange of data about software packages. This information includes general information about the package, licensing information about the package as a whole, a manifest of files contained in the package and licensing information related to the contained files.
-
-Known issues:</rdfs:comment>
-        <rdfs:label>SPDX 2.2</rdfs:label>
-        <owl:versionInfo rdf:datatype="http://www.w3.org/2001/XMLSchema#decimal">2.2</owl:versionInfo>
+SPDX® is an open standard for communicating software bill of material information, including components, licenses, copyrights, and security references. SPDX reduces redundant work by providing a common format for companies and communities to share important data, thereby streamlining and improving compliance.
+.
+Known issues:
+- rdfs:comment and rdfs:seeAlso are used within the SPDX classes and causes a redefinition of these standard terms</rdfs:comment>
+        <rdfs:label>SPDX 2.3</rdfs:label>
+        <owl:versionInfo rdf:datatype="http://www.w3.org/2001/XMLSchema#decimal">2.3</owl:versionInfo>
     </owl:Ontology>
     
 
@@ -40,18 +41,6 @@ Known issues:</rdfs:comment>
     
 
 
-    <!-- http://www.w3.org/2002/07/owl#deprecatedClass -->
-
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2002/07/owl#deprecatedClass"/>
-    
-
-
-    <!-- http://www.w3.org/2002/07/owl#deprecatedProperty -->
-
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2002/07/owl#deprecatedProperty"/>
-    
-
-
     <!-- http://www.w3.org/2002/07/owl#qualifiedCardinality -->
 
     <owl:AnnotationProperty rdf:about="http://www.w3.org/2002/07/owl#qualifiedCardinality"/>
@@ -60,10 +49,24 @@ Known issues:</rdfs:comment>
 
     <!-- http://www.w3.org/2003/06/sw-vocab-status/ns#term_status -->
 
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2003/06/sw-vocab-status/ns#term_status">
-        <rdfs:isDefinedBy>http://spdx.org/rdf/terms#</rdfs:isDefinedBy>
-        <ns:term_status>stable</ns:term_status>
-    </owl:AnnotationProperty>
+    <owl:AnnotationProperty rdf:about="http://www.w3.org/2003/06/sw-vocab-status/ns#term_status"/>
+    
+
+
+    <!-- 
+    ///////////////////////////////////////////////////////////////////////////////////////
+    //
+    // Datatypes
+    //
+    ///////////////////////////////////////////////////////////////////////////////////////
+     -->
+
+    
+
+
+    <!-- http://www.w3.org/2001/XMLSchema#date -->
+
+    <rdfs:Datatype rdf:about="http://www.w3.org/2001/XMLSchema#date"/>
     
 
 
@@ -128,9 +131,10 @@ Known issues:</rdfs:comment>
     <owl:ObjectProperty rdf:about="http://spdx.org/rdf/terms#artifactOf">
         <rdfs:domain rdf:resource="http://spdx.org/rdf/terms#SpdxElement"/>
         <rdfs:range rdf:resource="http://usefulinc.com/ns/doap#Project"/>
+        <rdfs:comment xml:lang="en">Deprecated as of version 2.1</rdfs:comment>
         <rdfs:comment xml:lang="en">Indicates the project in which the SpdxElement originated. Tools must preserve doap:homepage and doap:name properties and the URI (if one is known) of doap:Project resources that are values of this property. All other properties of doap:Projects are not directly supported by SPDX and may be dropped when translating to or from some SPDX formats.</rdfs:comment>
-        <owl:deprecatedProperty xml:lang="en">Deprecated as of version 2.1</owl:deprecatedProperty>
-        <ns:term_status>stable</ns:term_status>
+        <owl:deprecated rdf:datatype="http://www.w3.org/2001/XMLSchema#boolean">true</owl:deprecated>
+        <ns:term_status xml:lang="en">deprecated</ns:term_status>
     </owl:ObjectProperty>
     
 
@@ -196,7 +200,6 @@ Known issues:</rdfs:comment>
         <rdfs:domain rdf:resource="http://spdx.org/rdf/terms#SpdxDocument"/>
         <rdfs:range rdf:resource="http://spdx.org/rdf/terms#Package"/>
         <rdfs:comment xml:lang="en">The describesPackage property relates an SpdxDocument to the package which it describes.</rdfs:comment>
-        <owl:deprecatedProperty></owl:deprecatedProperty>
         <ns:term_status xml:lang="en">stable</ns:term_status>
     </owl:ObjectProperty>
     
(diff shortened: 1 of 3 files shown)


goneall:
Going back through my checklist I found some text for 2.2.1 that was not included in this update.

Changing to dreaft.

goneall:
This now include updates to fix #459

goneall:
Fixes #606

davaya:
JSON schema $id should identify the schema (as a purl identifies a package).

From: https://json-schema.org/understanding-json-schema/structuring.html:
> Schema documents are identified by non-relative URIs.

Line 2 of schemas/spdx-schema.json should be changed from a constant to a unique namespace identifier, something like:

`"$id" : "http://spdx.org/rdf/terms"`  
`"$id" : "http://spdx.org/spdx/v2.3",`

goneall:
@fviernau - Can you take a look and see if we missed anything?

ontology/spdx-ontology.owl.xml
```diff
@@ -9,14 +9,15 @@
...
     <owl:Ontology rdf:about="http://spdx.org/rdf/terms">
-        <owl:versionIRI rdf:resource="http://spdx.org/rdf/terms/2.2"/>
-        <rdfs:comment>This specification describes the SPDX® language, defined as a dictionary of named properties and classes using W3C&apos;s RDF Technology.
-
-SPDX® is a designed to allow the exchange of data about software packages. This information includes general information about the package, licensing information about the package as a whole, a manifest of files contained in the package and licensing information related to the contained files.
-
-Known issues:</rdfs:comment>
-        <rdfs:label>SPDX 2.2</rdfs:label>
-        <owl:versionInfo rdf:datatype="http://www.w3.org/2001/XMLSchema#decimal">2.2</owl:versionInfo>
+        <owl:versionIRI rdf:resource="http://spdx.org/rdf/terms/2.3"/>
```
zvr, line 12: I'd use the simpler `http://spdx.org/rdf/2.3`

ontology/spdx-ontology.owl.xml
```diff
@@ -9,14 +9,15 @@
...
+        <rdfs:comment xml:lang="en">This specification describes the SPDX® language, defined as a dictionary of named properties and classes using W3C&apos;s RDF Technology.
+
+SPDX® is an open standard for communicating software bill of material information, including components, licenses, copyrights, and security references. SPDX reduces redundant work by providing a common format for companies and communities to share important data, thereby streamlining and improving compliance.
+.
+Known issues:
+- rdfs:comment and rdfs:seeAlso are used within the SPDX classes and causes a redefinition of these standard terms</rdfs:comment>
+        <rdfs:label>SPDX 2.3</rdfs:label>
+        <owl:versionInfo rdf:datatype="http://www.w3.org/2001/XMLSchema#decimal">2.3</owl:versionInfo>
```
zvr, line 20: Why is the version a decimal? I'd suggest the datatype is deleted, by default it's a string.

ontology/spdx-ontology.owl.xml
```diff
@@ -307,7 +312,9 @@ Known issues:</rdfs:comment>
                 </owl:unionOf>
             </owl:Class>
         </rdfs:range>
-        <rdfs:comment xml:lang="en">The licensing that the preparer of this SPDX document has concluded, based on the evidence, actually applies to the package.</rdfs:comment>
+        <rdfs:comment xml:lang="en">The licensing that the preparer of this SPDX document has concluded, based on the evidence, actually applies to the package.
```
zvr, line 315: `licenseConcluded` is not only for packages, but also for files and snippets

ontology/spdx-ontology.owl.xml
```diff
@@ -307,7 +312,9 @@ Known issues:</rdfs:comment>
...
+
+If the Concluded License field is not present for an SPDX Item, it implies an equivalent meaning to NOASSERTION.</rdfs:comment>
```
zvr, line 317: Comments refer to the RDF terms, so use `licenseConcluded` instead of `Concluded License`

ontology/spdx-ontology.owl.xml
```diff
@@ -354,7 +361,9 @@ Known issues:</rdfs:comment>
                 </owl:unionOf>
             </owl:Class>
         </rdfs:range>
-        <rdfs:comment xml:lang="en">The licensing information that was discovered directly within the package. There will be an instance of this property for each distinct value of alllicenseInfoInFile properties of all files contained in the package.</rdfs:comment>
+        <rdfs:comment xml:lang="en">The licensing information that was discovered directly within the package. There will be an instance of this property for each distinct value of alllicenseInfoInFile properties of all files contained in the package.
+
+If the All Licenses Information from Files field is not present for a package and FilesAnalyzed property for that same pacakge is true or omitted, it implies an equivalent meaning to NOASSERTION.</rdfs:comment>
```
zvr, line 366: Comments refer to the RDF terms, so use `licenseInfoFromFiles` instead of `All Licenses Information from Files` (`FilesAnalyzed` is correct)

ontology/spdx-ontology.owl.xml
```diff
@@ -380,8 +389,10 @@ Known issues:</rdfs:comment>
                 </owl:unionOf>
             </owl:Class>
         </rdfs:range>
-        <rdfs:comment xml:lang="en">Licensing information that was discovered directly in the subject file. This is also considered a declared license for the file.</rdfs:comment>
-        <ns:term_status>deprecated</ns:term_status>
+        <rdfs:comment xml:lang="en">Licensing information that was discovered directly in the subject file. This is also considered a declared license for the file.
+
+If the License Information in File field is not present for a file, it implies an equivalent meaning to NOASSERTION.</rdfs:comment>
```
zvr, line 394: Comments refer to the RDF terms, so use `licenseInfoInFile` instead of `License Information in File`

ontology/spdx-ontology.owl.xml
```diff
@@ -406,7 +417,9 @@ Known issues:</rdfs:comment>
                 </owl:unionOf>
             </owl:Class>
         </rdfs:range>
-        <rdfs:comment xml:lang="en">Licensing information that was discovered directly in the subject snippet. This is also considered a declared license for the snippet.</rdfs:comment>
+        <rdfs:comment xml:lang="en">Licensing information that was discovered directly in the subject snippet. This is also considered a declared license for the snippet.
+
+If the License Information in Snippet field is not present for a snippet, it implies an equivalent meaning to NOASSERTION.</rdfs:comment>
```
zvr, line 422: Comments refer to the RDF terms, so use `licenseInfoInSnippet` instead of `License Information in Snippet`

ontology/spdx-ontology.owl.xml
```diff
@@ -786,7 +824,9 @@ Known issues:</rdfs:comment>
         <rdfs:domain rdf:resource="http://spdx.org/rdf/terms#SpdxItem"/>
         <rdfs:range rdf:resource="http://www.w3.org/2000/01/rdf-schema#Literal"/>
         <rdfs:range rdf:resource="http://www.w3.org/2001/XMLSchema#string"/>
-        <rdfs:comment xml:lang="en">The text of copyright declarations recited in the Package or File.</rdfs:comment>
+        <rdfs:comment xml:lang="en">The text of copyright declarations recited in the Package, File or snippet.
```
zvr, line 827: For consistency, use `package, file` instead of `Package, File`. There are also more instances of incorrect capitalization.

ontology/spdx-ontology.owl.xml
```diff
@@ -786,7 +824,9 @@ Known issues:</rdfs:comment>
...
+
+If the License Information in File field is not present, it implies an equivalent meaning to NOASSERTION.</rdfs:comment>
```
zvr, line 829: Definitely use `copyrightText` instead of `License Information in File`

zvr (changes requested):
I only looked and commented on a single ontology file, the XML one. I assume all are somehow automatically cross-generated.

It appears that on GitHub I can only comment on changed lines, so here's a couple of comments on non-changed lines:

line 3, default and base namespaces: 
Shall we move this to `http://spdx.org/rdf/2.3/terms`? At some point we should, definitely for 3.0.

line 1124, `licenseId`:
According to Annex D (SPDX license expressions), we actually allow `["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring)`. The description should be amended to reflect this.

schemas/spdx-schema.json
```diff
@@ -380,6 +385,14 @@
             "description" : "The base name of the package file name. For example, zlib-1.2.5.tar.gz.",
             "type" : "string"
           },
+          "packagePurposes" : {
```
rjb4standards, line 388: I'm concerned that making PackagePurpose an array will impact fidelity when translating to other formats, plus it raises ambiguity as to the "primary purpose" of a package.

goneall: Since this would change the main body of the spec, I opened issue #720 to track this

ontology/spdx-ontology.owl.xml
goneall: > I'd use the simpler http://spdx.org/rdf/2.3

Previous versions have always used the `rdf/terms` as part of the URI - so to keep it consistent, I would like to leave this as is.

ontology/spdx-ontology.owl.xml
goneall: @zvr I agree - however, whenever I try to change this in Protege tool, it keeps changing it back to a decimal type.  I looked up the W3C OWL spec and there is no requirement for this to be a decimal datatype.

Let me know if you find a way to fix this (other than manually patching it after editing in Protege).

ontology/spdx-ontology.owl.xml
goneall: These should just be copy/paste from the spec.  I'll fix this one, but I probably won't have time to review the entire set of comments for capitalization.

goneall:
Thanks @zvr for the review - I updated the schemas to resolve most of the identified issues.  Comments for those issues I didn't resolve.

zvr:
@goneall 

Any response to my comment above:

> ` licenseId`:
> According to Annex D (SPDX license expressions), we actually allow ["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring). The description should be amended to reflect this.

goneall:
> licenseId:
> According to Annex D (SPDX license expressions), we actually allow ["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring). The description should be amended to reflect this.

@zvr - oops - missed that one.

I could not find any suitable definition in the spec that explains the semantics for the `["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring)` to copy into the schema.

In thinking about this, it is a bit more complicated how to handle it in the various schemas.

Currently, the LicenseID is only used by itself as a field in the `SimpleLicenseInfo` class (parent class to the listed and extracted license classes).

In non-RDF serializations, the license expression is used to reference other licenses.  

In RDF, the URI is used to reference.

I'm thinking we only document the `ExternalDocumentRef` in the non-RDF schemas to document the license expression syntax.

The tool that generates the non-RDF schemas currently prepends the following text to all fields that use license expressions:

`License expression for [field]`.

We could expand this to something like `'License expression for [field].  See SPDX Appendix D for the license expression syntax`.

Let me know what you think.

zvr:
@goneall I only mentioned it because the comment of the `licenseId` field currently mentions:

> A human readable short form license identifier for a license. The license ID is either on the standard license list or the form "LicenseRef-"[idString] where [idString] is a unique string containing letters, numbers, ".", "-" or "+".

goneall:
@zvr How about adding the sentence:

"When used within a license expression, the license ID can optionally include a reference to an external document in the form `["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring)`."

goneall:
@zvr I will update the license ID text to the following:

```
A human readable short form license identifier for a license. The license ID is either on the standard license list or the form "LicenseRef-[idString]" where [idString] is a unique string containing letters, numbers, ".", "-" or "+".  When used within a license expression, the license ID can optionally include a reference to an external document in the form "DocumentRef-[docrefIdString]:LicenseRef-[idString]" where docRefIdString is an ID for an external document reference.
```

Let me know if you see any issues with the above text.

goneall:
I also added the following to the JSON schema license expression field descriptions:

> See SPDX Annex D for the license expression syntax.

goneall:
> One thing -- unless I'm misunderstanding or missing the context, I don't think [idString] can include "+"? I think that the plus sign can be part of a license expression, but not part of the idString in an identifier itself.

Good catch @swinslow  - I updated the description to remove the `+`.  Updated text below.

> A human readable short form license identifier for a license. The license ID is either on the standard license list or the form "LicenseRef-[idString]" where [idString] is a unique string containing letters, numbers, "." or "-".  When used within a license expression, the license ID can optionally include a reference to an external document in the form "DocumentRef-[docrefIdString]:LicenseRef-[idString]" where docRefIdString is an ID for an external document reference.

goneall:
I believe all review comments have been taken care of - merging this in.

PR #731, opened 2022-06-25
Add Checksum algorithms to package chapter and add missing linkes

### LINKED ISSUE #730: Checksum Documentation Inconsistency
The new checksums added for 2.3 need to be added to the[ Package Information](https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/package-information.md).  Suggest we just make a reference to the file section similar to what we did in the [external document reference checksum section](https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/document-creation-information.md#66-external-document-references-field-).

I also noticed that the external links are missing.

goneall:
Fixes #730


kestewart (approved):
LGTM - Agreed on changes.

seabass-labrax:
Also, we don't have any reference URLs for the File Information chapter. We might want to add them there for consistency :smiley:

PR #733, opened 2022-06-29
Add example information for NIST EO 14028 to Appendix G

rnjudge:
This commit adds an example to Appendix G.1 describing how users can
reference vulnerability reports using the ExternalRefs `SECURITY`
category in accordance with Executive Order 14028.

This example and language was agreed upon at the SPDX Defects WG meeting
on June 29th.

### DIFF
--- a/chapters/how-to-use.md
+++ b/chapters/how-to-use.md
@@ -179,6 +179,23 @@ One can also use it to refer to guidance related to a vulnerability such as CISA
 } ]
 ```
 
+### G.1.9 Linking to an SBOM vulnerability report for a Software Product (per NIST Executive Order 14028)
+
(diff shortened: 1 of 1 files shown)


rnjudge:
@zvr I'm not sure the two are incompatible. This is an example for those looking for guidance on how to correlate vulnerability and SBOM information for a software product at the component level in order to satisfy EO 14028. Per ISO 29147:2018, the advisory may contain an impact statement whether a package (e.g. a product) is or is not affected by vulnerabilities, which the example file does. Technically, even a JSON file stored on GitHub is "published". I agree that there could be a better example document, but the one there now is consistent with the guidance we have provided.

kestewart (approved):
Been following the discussion, and this matches.   Looks good.

PR #734, opened 2022-07-01
Add SHA1-DC algorithm to recognized set.

kestewart:
From information provided by David A. Wheeler:  Git uses SHA-1DC, not SHA-1, as its default hash algorithm. This algorithm normally produces the same result as sha-1 but it tries to detect collision attacks and provides a different result then.  It's been in use since 2017, so we should add it as a valid option. 

Implementation and link to paper about it: https://github.com/cr-marcstevens/sha1collisiondetection
News article about git and hash algorithms: https://lwn.net/Articles/898522/

Also included is adding back SHA1 reference, which appears to have been accidently removed when the new algorithms were added. 

### DIFF
--- a/chapters/file-information.md
+++ b/chapters/file-information.md
@@ -171,8 +171,8 @@ The metadata for the file checksum field is shown in Table 39.
 | Attribute | Value |
 | --------- | ----- |
 | Required | Yes |
-| Cardinality | 1..1 for the [`SHA1`][SHA-1] algorithm, 0..* for all other algorithms |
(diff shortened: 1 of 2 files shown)


goneall:
In reading the background on this, I'm not sure this is really a different hash than SHA1.  It the definition of this a SHA1 which has passed the collision detection from the referred to github repo?  If so, I think we need to define that in the spec - otherwise it may be confusing.  It would be nice if there was a website with a description we could refer to that provide more of a definition - the repo as a reference is a bit confusing.

chapters/file-information.md
```diff
@@ -795,7 +795,8 @@ EXAMPLE 2 RDF: Property `spdx:fileDependency` in class `spdx:File`
 [pip-vcs]: https://pip.pypa.io/en/latest/reference/pip_install.html#vcs-support
 [Red Hat]: https://www.redhat.com/
 [rfc3986]: https://tools.ietf.org/html/rfc3986
-[SHA-1]: https://tools.ietf.org/html/rfc3174
+[SHA1]: https://tools.ietf.org/html/rfc3174
+[SHA1-DC]: https://github.com/cr-marcstevens/sha1collisiondetection
```
goneall, line 799: Is the intent that the SHA1-DC is a SHA1 that the SPDX creator states has passed the collision detection?  If so, I'm not sure this is really a different HASH algorithm from SHA1.

goneall (changes requested):
Can we find a way to clarify this - I found it confusing as to what exactly SHA1-DC represents.

kestewart:
Based on comments from meeting on 6/5, we're going to close this for now, and revisit as an algorithm for 3.0.

PR #735, opened 2022-07-05
fix missing words in sentence in H.3

silverhook:
Found a small lapsus calami :]

PR #736, opened 2022-07-06
fix missing words in sentence in H.3

silverhook:
proper PR instead of my mess in https://github.com/spdx/spdx-spec/pull/735


goneall:
Thanks @silverhook

PR #742, opened 2022-07-11
Fix RDF date example

HansBusch:
The RDF example includes spaces which are not allowed according to the definition above.


zvr (approved):
Great catch!

HansBusch:
Signed-off-by: HansBusch <47081128+HansBusch@users.noreply.github.com>

HansBusch:
Sorry, was not aware to add the sign-off line with the original commit. Hope it is ok now for removing two blanks.

PR #747, opened 2022-07-18
Remove range from FileType in Ontology

### LINKED ISSUE #745: Ontology: mismatch in individuals lists
In the RDF ontology, we have:

- in the [definition of filetype](https://github.com/spdx/spdx-spec/blob/development/v2.3/ontology/spdx-ontology.owl.xml#L241-L269), it is specified that it can be one of 4 values; but
- there are 11 [discrete filetypes defined](https://github.com/spdx/spdx-spec/blob/development/v2.3/ontology/spdx-ontology.owl.xml#L2941-L3049)

I assume the union restriction should be expanded to include all types.

goneall:
The enumeration values of file type is defined by the type associated with each individual value, so there is no need for a range.

Fixes #745 

### DIFF
--- a/ontology/spdx-ontology.owl.ttl
+++ b/ontology/spdx-ontology.owl.ttl
@@ -164,25 +164,6 @@ xsd:date rdf:type rdfs:Datatype .
 ###  http://spdx.org/rdf/terms#fileType
 :fileType rdf:type owl:ObjectProperty ;
           rdfs:domain :File ;
-          rdfs:range [ rdf:type owl:Class ;
-                       owl:unionOf ( [ rdf:type owl:Restriction ;
-                                       owl:onProperty :fileType ;
-                                       owl:hasValue :fileType_archive
-                                     ]
-                                     [ rdf:type owl:Restriction ;
-                                       owl:onProperty :fileType ;
-                                       owl:hasValue :fileType_binary
-                                     ]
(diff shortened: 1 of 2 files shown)


zvr:
Correct me if I"m wrong but these are two discrete statements in the ontology:

1. `fileType_archive` (and all others) are values for `fileType`
2. `fileType` can only take one of these values

What this change does is remove the second.

It also seems we don't provide it in other enumerations (any more?) like `algorithm` or `purpose`, etc.

But shouldn't we? Defining some values in the ontology (i.e., creating individuals of a type) does not introduce a restriction that only _these_ values are the valid ones.

goneall:
@zvr You are correct.  I removed the range restrictions on the other enumerations (such as checksum algorithm) a couple releases ago since they quickly became out of date and out of sync with the individual values.  This makes the file types consistent with the approach used for the other enumeration values.

We could add the ranges back in, but we should do it for all the enumerations and we will need to maintain them - which turns out to be quite a bit of manual error prone work today.

I would suggest we leave the range off for now and add ranges back in in 3.0 when we have some automation.

zvr (approved):
As long as we understand what we're doing, I'm fine.

goneall:
Approved by @zvr - I'll go ahead and merge

PR #751, opened 2022-07-20
Add how to use Annex in the Table of Contents

goneall:
Rename the annex to be at the end of the annexes


kestewart (approved):
Looks good to resolve the Annex naming collision.    There may be other references to this annex in the text that exist and it should be double checked, but that can be handled by a separate PR.

Issue #757, opened 2022-07-21
Ontology issues (collected)


zvr:
While generating the diagrams and documentation for the ontology, I went at it with a very fine comb...

@goneall these are issues that I found. Most of them are minor, but it wouldn't hurt to have them fixed.

I assume you will work on Protege, so I won't volunteer to provide textual patches to the `.xml` file. For this reason, I also do not provide line numbers on each one; let me know if you need them.

1. `primaryPackagePurpose` has no range (should be `Purpose`)
2. `referenceCategory` has no range (should be `ReferenceCategory`)
3. `externalDocumentRef` has no range (should be `ExternalDocumentRef`)
4. `order` has no domain (should be `CrossRef`)
5. `hasExtractedLicensingInfo` has two domains, but `ExtractedLicensingInfo` should be its range
6. `range` (in `Snippet`) is shown as having range (ha!) both `CompoundPointer` and `StartEndPointer` at different places
7. in `SpdxDocument`, there are restrictions that `dataLicense` can be `AnyLicenseInfo` but also that it has to be `http://spdx.org/licenses/CC0-1.0`
8. there is a reference to `http://www.w3.org/2000/01/rdf-schema#member` which is confusing, given that we have our own `http://spdx.org/rdf/terms#member`
9. `licenseInfoFromFiles` in `SpdxItem`(and corresponding sublasses in packages, files, snippets) seem to be `SimpleLicenseInfo` (or NOASSERTION or NONE). Why aren't license expressions allowed?
10. `License` has `licenseText`, `standardLicenseTemplate`, `standardLicenseHeader`, and `standardLicenseHeaderTemplate`.  `ListedLicense` has `licenseTextHtml` `standardLicenseHeaderHtml`.  However, `standardLicenseHeaderTemplate` has domain `ListedLicense` instead of `License`.

Sorry about all this!

goneall:
Thanks @zvr - you really did go through a thorough review.

I'll take a look at these and see if I can get some fixes in today or tomorrow.

goneall:
I think the range in snippets (number 6 above) is OK since StartEndPointer is subclass of CompoundPointer and the more specific restriction is a subclass of the Snippet class.  However, I'm not completely sure.  We could just remove the range on the range property or change the Snippet range.  @zvr - let me know your thoughts on this.

goneall:
[...]

PR #759, opened 2022-07-21
Fix minor ontology errors

goneall:
Fixes the following items from #757:

- `1.` `primaryPackagePurpose` has no range (should be `Purpose`)
- `2.` `referenceCategory` has no range (should be `ReferenceCategory`)
- `3.` `externalDocumentRef` has no range (should be `ExternalDocumentRef`)
- `4.` `order` has no domain (should be `CrossRef`)
- `8.` there is a reference to `http://www.w3.org/2000/01/rdf-schema#member` which is confusing, given that we have our own `http://spdx.org/rdf/terms#member`
- `10.` `License` has `licenseText`, `standardLicenseTemplate`, `standardLicenseHeader`, and `standardLicenseHeaderTemplate`.  `ListedLicense` has `licenseTextHtml` `standardLicenseHeaderHtml`.  However, `standardLicenseHeaderTemplate` has domain `ListedLicense` instead of `License`.

### DIFF
--- a/ontology/spdx-ontology.owl.ttl
+++ b/ontology/spdx-ontology.owl.ttl
@@ -140,6 +140,7 @@ xsd:date rdf:type rdfs:Datatype .
 ###  http://spdx.org/rdf/terms#externalDocumentRef
 :externalDocumentRef rdf:type owl:ObjectProperty ;
                      rdfs:domain :SpdxDocument ;
+                     rdfs:range :ExternalDocumentRef ;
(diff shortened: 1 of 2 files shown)


zvr (approved):

zvr:
BTW. this also fixes

5. `hasExtractedLicensingInfo` has two domains, but `ExtractedLicensingInfo` should be its range

zvr (approved):

PR #767, opened 2022-07-27
Update SPDX license list Annex to version 3.17

kestewart:
Added new licenses since v. 3.8.   
Added FSF column
Reworked intro text to closer match current web pages
Added in new introduced execptions in the last 2 years
Added in 3 deprecated licenses since v 3.8


swinslow (approved):
Hi @kestewart, I pushed a handful of changes / fixes to a few values and added a few missing licenses. With those edits, I think this should be good to go.

kestewart:
Thanks Steve.     There's two places where I question the "-", but the rest looks good, and detailed review is appreciated. 

@goneall - if you don't spot any show stoppers, please merge, and we can tweak it a bit more later during the review period, if needed.

swinslow:
@kestewart Fully agree that the "-" is incorrect in those places -- but it matches what is (I think incorrectly) on the actual license list, so I think it's appropriate to add it here. :shrug:  Thanks!

goneall (approved):

PR #769, opened 2022-07-27
Add new diagrams for v2.3 ontology

zvr:


zvr:
@goneall I've just realized that [Annex C](https://spdx.github.io/spdx-spec/v2.3/RDF-object-model-and-identifier-syntax/) uses your diagram of the ontology, so I didn't touch it.

I'll submit a new PR for the HTML version of the ontology shortly.

goneall (approved):
LGTM - I like having the separate full and simplified

PR #772, opened 2022-07-27
Update Annex C to use ontology graphic

goneall:


chapters/RDF-object-model-and-identifier-syntax.md
```diff
@@ -6,9 +6,9 @@ SPDX ® Vocabulary Specification
 
 See: [http://spdx.org/rdf/ontology/spdx-2-2-1](http://spdx.org/rdf/ontology/spdx-2-2-1)
```
kestewart, line 7: I think this line needs to be updated as well.

kestewart (changes requested):
Spotted a 2.2.1 reference that still needs to be updated.

goneall:
> Spotted a 2.2.1 reference that still needs to be updated.

Fixed in the text, but it will be a dead link until I figure out how to update the new website.  We'll probably have the HTML live by the time the review goes out.

PR #773, opened 2022-08-03
Fix list rendering in Differences page

puerco:
This PR adds newlines to items in the bulleted lists of differences between V2.3 and V2.2.2 and V2.2.2 and V2.2.1 to fix a rendering problem when building the HTML code from the markdown source.

(Note that this rendering problem [is present in the v2.2.2 spec](https://spdx.github.io/spdx-spec/diffs-from-previous-editions/) too).

### Before: 
image

### Now: 

image

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>


goneall (approved):
Thanks @puerco - much better rendering!

puerco:
Thank you!

Issue #593, opened 2021-10-29
Contradiction in Clause 5.2.3


Jayman2000:
>[5.2.3 File information section](https://spdx.github.io/spdx-spec/composition-of-an-SPDX-document/#523-file-information-section)
>One instance of the File Information [section] shall exist for each file in the software package.
>[…]
>Cardinality: Optional, one or many.

“shall” and “optional” are basically opposites.

kestewart:
Agreement to clean up the wording here in 2.3 from call.   Pull request welcome.

rnjudge:
I can try to help with this.

kestewart:
@nishakm to review the PR and check.

nishakm:
Is the change just `s/shall/may`? I can submit a PR for this. Is it targeted to 2.3?

Jayman2000:
> @Jayman2000 can you clarify the contradiction -- does it still exist?

I can still find it [here](https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/composition-of-an-SPDX-document.md#523-file-information-section-).

Hopefully, this will clarify why I think this is a contradiction:

> One instance of the file information shall exist for each file in the software package.

I’m assuming here that a word was accidentally left out and that the author meant “one instance of the file information _section_”. To me, that quote means that every SPDX document needs to have at least one file information section instance because every SPDX document describes at least one software package.

> Cardinality: Optional, one or many.

To me, that quote means that you could create an SPDX document with zero file information section instances. So which one is it? Are you allowed to have zero or do you need to have at least one?

Now that I’m rereading it, I’m realizing that there’s another way you could interpret this clause. You could interpret it as “you need to have one file information section instance per file per package information section instance, but you can have zero file information section instances because you can have zero package information sections instances.”

---

> According to https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/file-information.md file name and file SPDX identifier, at least some of `File Information` fields are required (i.e. file field name, file SPDX identifier, etc). There are a few descriptors that are not required (i.e. `File type field`), but at least _some_ file information is required which seems appropriate for the description using `shall exist`.

Yes, but those fields are required inside file information section instances. This issue is about whether or not file information section instances are required, not about what each of them must contain

PR #775, opened 2022-08-10
Clarify sections as optional.

kestewart:
Change that "Optional, one or many" to "Optional, zero or many" so it's clear that these elements are indeed "optional".    The notation of "one or many" was being interpreted to override the "optional".    This should address remaining part of https://github.com/spdx/spdx-spec/issues/593

### DIFF
--- a/chapters/composition-of-an-SPDX-document.md
+++ b/chapters/composition-of-an-SPDX-document.md
@@ -39,7 +39,7 @@ Note that some of these could be represented in SPDX as a file as well.
 
 In an SPDX document, relationship elements can be used to indicate relationships between packages, such as dependency relationships.
 
-Cardinality: Optional, one or many.
+Cardinality: Optional, zero or many.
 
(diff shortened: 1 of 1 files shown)


chapters/composition-of-an-SPDX-document.md
```diff
@@ -75,11 +75,11 @@ When implementing file information in RDF, the `spdx:hasFile` property is used t
 
 ### 5.2.4 Snippet information section <a name="5.2.4"></a>
 
-Snippets can optionally be used when a file is known to have some content that has been included from another original source. They are useful for denoting when part of a file may have been originally created under another license.
+Snippets can optionally be used when a file is known to have some content that has been included from another original source. They are useful for denoting when part of a file may have been originally created under another license or copied from a place with a known vulnerability.
```
goneall, line 78: I like this addition :)

goneall (approved):

PR #779, opened 2022-08-11
Clarify that name is optional for extracted license info

### LINKED ISSUE #506: Other licensing information detected in JSON example is not spec compliant
https://github.com/spdx/spdx-spec/blob/3dddfc34d7e770d4c17fc19f61be8554e3934c3d/examples/SPDXJSONExample-v2.2.spdx.json#L33-L36

The JSON example does not include `name` field for all instances of other licensing information detected, even though it is ["conditional (mandatory, one) if license is not on SPDX License List"](https://spdx.github.io/spdx-spec/6-other-licensing-information-detected/#63-license-name) according to the spec.

goneall:
Fixes #506 

This PR makes the name field for the extracted license text name field to be optional rather than requiring a `NOASSERTION` value.  This is consistent with the other licensing related changes in the 2.3 release where we are making values optional rather than requiring `NOASSERTION` when the field is not known.

### DIFF
--- a/chapters/other-licensing-information-detected.md
+++ b/chapters/other-licensing-information-detected.md
@@ -113,15 +113,15 @@ If indeed full text of license present in File:
 
 Provide a common name of the license that is not on the SPDX list.
 
-Use `NOASSERTION` If there is no common name or it is not known. The metadata for the license name field is shown in Table 65.
+The metadata for the license name field is shown in Table 65.
 
 **Table 65 — Metadata for the license name field**
 
 | Attribute | Value |
 | --------- | ----- |
-| Required | Conditional |
-| Cardinality | 0..1 conditional (Mandatory, one) if license is not on SPDX License List. |
-| Format | Single line of text | `NOASSERTION` |
+| Required | No |
+| Cardinality | 0..1 |
+| Format | Single line of text |
 
 ### 10.3.2 Intent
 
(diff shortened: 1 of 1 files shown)


swinslow:
I'm on board with the idea of making this change, but I have two comments:

1) The cardinality format for other licensing section fields that are optional seems to be `0..1 conditional (optional, one) if license is not on SPDX License List.` -- see [section 10.4](https://spdx.github.io/spdx-spec/other-licensing-information-detected/#104-license-cross-reference-field) as an example.

2) I wouldn't suggest removing `NOASSERTION` altogether. Arguably that would be incompatible with 2.2 and prior. (Arguably not, because the word "NOASSERTION" is a text string, but I think we ought to keep the meaning aligned with prior uses.) I'd be more inclined to leave the prior language there, and to add what we added for the separate licensing fields -- e.g. `If the License Name field is not present for a license, it implies an equivalent meaning to NOASSERTION.`

kestewart:
I'm ok with leaving the NOASSERTION statements, to minimize migration problems.   Can you update this PR to leave the NOASSERTION part there @goneall

goneall:
@swinslow Thanks for the review and comments.

I just updated the PR - if you could give it another look.

swinslow (approved):
LGTM, thanks @goneall!

kestewart (approved):
Looks good Gary.  Thank you.

PR #784, opened 2022-08-15
Fix 7.21 violation (`PACKAGE_MANAGER`)

patriksvensson:
According to [7.21 External reference field](https://spdx.github.io/spdx-spec/package-information/#721-external-reference-field), it should be `PACKAGE-MANAGER` and not `PACKAGE_MANAGER`.

The SPDX Online Tool agrees with this conclusion.

image


goneall:
@patriksvensson Thanks for the PR to fix this - this is also fixed in PR #782 which is awaiting review before merging.

Once PR #782 is reviewed / merged we can probably close this PR.

goneall:
resolved with PR #782 - closing

PR #793, opened 2022-10-02
Update JSON schema for dash and underscore enums

### LINKED ISSUE #792: PACKAGE_MANAGER vs PACKAGE-MANAGER
I see in #618 that it was decided to use PACKAGE_MANAGER instead of PACKAGE-MANAGER as an external reference category.  However, in a [commit from July](https://github.com/spdx/spdx-spec/commit/214f23d34ee287cb1db5b31c3d571af291e836f3) this  was undone in the schema so that it now is PACKAGE-MANAGER.  There are SBOM generators out there that use both values (I was doing some testing and noticed the difference when reviewing output).  I personally don't care which one gets used, though I suspect PACKAGE_MANAGER would require fewer utilities to be updated, but either one needs to be chosen OR both need to be allowed for interoperability.

Note: This affects PERSISTENT_ID as well.

goneall:
Allow for both 'PACKAGE_MANAGER' and 'PACKAGE-MANAGER'. Also allows for both 'PERSISTENT-ID' and 'PERSISTENT_ID'.

Resolves #792

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -322,7 +322,7 @@
                 "referenceCategory" : {
                   "description" : "Category for the external reference",
                   "type" : "string",
-                  "enum" : [ "OTHER", "PERSISTENT-ID", "SECURITY", "PACKAGE-MANAGER" ]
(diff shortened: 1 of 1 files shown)


goneall:
Discussed on the SPDX tech call on 4 Oct 2022 - no objections to this PR, so I'll go ahead and merge.

PR #796, opened 2022-10-06
Add documentDescribes to required fields in JSON Schema

### LINKED ISSUE #795: JSON schema should include `documentNamespace` as required
According to the spec here: https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#65-spdx-document-namespace-field `Document Namespace` is required, but the JSON schema does not include it as a required field here: https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json#L738

goneall:
Fixes #795

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -735,6 +735,6 @@
       }
     }
   },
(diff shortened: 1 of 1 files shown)


maxhbr (approved):
LGTM ;)

PR #800, opened 2022-10-13
Update examples with correct referenceType for purl

### LINKED ISSUE #799: Erroneous referenceType in xml, rdf.xml and yaml examples
The referenceType in [the rdf.xml example](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/examples/SPDXRdfExample-v2.3.spdx.rdf.xml#L1525), [the xml example](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/examples/SPDXXMLExample-v2.3.spdx.xml#L270) and [the yaml example](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/examples/SPDXYAMLExample-2.3.spdx.yaml#L235) is given as
```
referenceType: "http://spdx.org/spdxdocs/spdx-example-444504E0-4F89-41D3-9A0C-0305E82C3301#purl"
```
According to [this discussion](https://github.com/spdx/tools-java/issues/80#issuecomment-1276511882) with @goneall , it should be
```
http://spdx.org/rdf/references/purl
```
instead (see also https://spdx.github.io/spdx-spec/package-information/#7213-examples)

goneall:
Fixes #799


goneall:
> It looks like the rdf example was renamed to SPDXJSONExample-... - that's probably by mistake?

Yes - that was an error - I'll add a commit to rename.

> Finally, there are loads of changes in the rdf. Not sure why (maybe different tool version?), but I didn't really look at those.

This is an issue with the [Jena library](https://jena.apache.org/) we use to serialize the RDF.  It tends to resort everything making it very difficult to diff/compare.  I wonder if there is a way to declare some ordering which would make the diffs easier - we probably aren't the first people to run into this.

Issue #803, opened 2022-10-28
SPDX Spec URL points to 2.2.2 spec - should be 2.3


goneall:
https://spdx.github.io/spdx-spec/ points to 2.2.2 instead of 2.3

@iamwillbar @tsteenbe Could you both take a look at this and see if it can be fixed?  We have a number of website which use this URL for the current version.

nicoweidner:
To add to this, since we also noticed recently: In the URL Gary posted, there is no version selection dropdown in the top left. When navigating to a version-specific URL such as https://spdx.github.io/spdx-spec/v2.3/, the dropdown appears (see screenshot).
I think having the dropdown on the "main" page would be good!
![Screenshot from 2022-10-31 10-42-56](https://user-images.githubusercontent.com/66428403/198978808-f69c13f1-227b-4726-a2ec-a104bb7f47b2.png)

maxhbr:
The publish step only runs, if both `build_current` and `build_v2-draft` were executed
https://github.com/spdx/spdx-spec/blob/4021a7ae502352cd79bae2979859f28072fc402f/.github/workflows/publish.yml#L46-L49

But
* `build_current` only runs on `'refs/heads/master'` and
* `build_v2-draft` only runs on `'refs/heads/development/v2.3'`

so they never run both.

As a result, the publish is never executed.

goneall:
Thanks @maxhbr for the quick analysis and PR.

You'll find some related PR's where we've tried to fix similar issues - perhaps one of those PR's caused this issue?

goneall:
From a quick check - it looks like this has been resolved

PR #806, opened 2022-11-08
This is a quick fix, to redirect https://spdx.github.io/spdx-spec/ to https://spdx.github.io/spdx-spec/v2.3

maxhbr:
This comes with the disadvantage, that deep link to files that this PR removes, no longer work. 

This is an ugly hack, which might mitigate the issue until CI and https://github.com/spdx/spdx-spec/issues/803 is solved appropriately.


goneall (approved):
Thanks @maxhbr - you're right - it's an ugly hack, but it will get us by in the short term.

goneall:
@iamwillbar @tsteenbe - Let me know if you have any concerns with this approach.  If not (or if I don't hear back in a couple of days), I'll go ahead and merge.

@kestewart FYI

goneall:
I'll go ahead and merge this as a temporary fix

Issue #720, opened 2022-06-15
Should packagePurpose have cardinality 0..1?


goneall:
[Package Purpose](https://github.com/spdx/spdx-spec/blob/development/v2.3/chapters/package-information.md#724-package-purpose-field-) currently has a cardinality of `0..*` allowing more than one purpose.

@rjb4standards raised a good [issue in the schema pull request](https://github.com/spdx/spdx-spec/pull/716#pullrequestreview-1007709766):

> I'm concerned that making PackagePurpose an array will impact fidelity when translating to other formats, plus it raises ambiguity as to the "primary purpose" of a package.

Should we change the cardinality to `0..1` to resolve this issue?

rnjudge:
> I agree with @rjb4standards issue - I would vote to change the cardinality to `0..1`

+1 from me as well. `0...1` makes sense to me.

davaya:
This was covered in the v3 discussions, and I believe the consensus was that a package could have multiple SoftwarePurposes (e.g., [APPLICATION, SOURCE]).  Since Package Purpose didn't exist in v2.2.2 and was added in 2.3, one might conclude that it was a backport of something useful from v.3.

If the decision is to allow only a single purpose in 2.3, then the same rationale should apply and v3.0 should also allow only a single purpose.

goneall:
> If the decision is to allow only a single purpose in 2.3, then the same rationale should apply and v3.0 should also allow only a single purpose. I take no position on which to choose, only that they be consistent.

Agree - my vote would be to change this for both 2.3 and 3.0

Now bringing @iamwillbar into the conversation.  William - do you recall the proponents of allowing more than one purpose?  If it was theoretical, we have a practical reason to keep it at 1.  If there was a use case which required multiple, we should probably discuss before making a decision.

rjb4standards:
+1 to Garys suggestion

seabass-labrax:
@goneall, I'm one of the proponents you mention :smiley:

I feel strongly that this should be `0..*`, which is what we came to in previous discussions (see my [reply to Nisha on the original pull request](https://github.com/spdx/spdx-spec/pull/622#issuecomment-1040577007) for context) for context). This is because some of the package purpose options overlap, limiting future use of a SBOM in which a more general statement is made.

PS. This would apply to both 2.x and 3.x.

rjb4standards:
[...]

PR #819, opened 2022-12-02
Fix cardinality of Primary Package Purpose

rnjudge:
It was brought to the attention of the group in the most recent SPDX DocFest that the Primary Package Purpose cardinality was incorrectly listed in the spec as 0..* despite the decision[1] that it should be 0..1. The cardinality was correct in the JSON schema and OWL ontology, however, so the consensus was to update the 2.3 specification text with a note noting the update.

[1] https://github.com/spdx/spdx-spec/issues/720

### DIFF
--- a/chapters/package-information.md
+++ b/chapters/package-information.md
@@ -1445,7 +1445,7 @@ The metadata for the Primary Package Purpose field is shown in Table 36.
 | Attribute | Value |
 | --------- | ----- |
 | Required | No |
-| Cardinality | 0..* |
+| Cardinality | 0..1 <br> _NOTE: The cardinality was incorrectly printed as 0..\* in the original publishing of the 2.3 specification_ |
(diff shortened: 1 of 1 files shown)


rnjudge:
@kestewart @goneall I added a note to the change in the direct text of the spec. I wasn't sure if this is where we wanted to note the typo or somewhere else (In the release notes or changelog?). I am happy to update this PR if you would like the note somewhere else. Also, I opened the PR against the 2.3 branch since it is already fixed in 2.3.1.

chapters/package-information.md
```diff
@@ -1445,7 +1445,7 @@ The metadata for the Primary Package Purpose field is shown in Table 36.
 | Attribute | Value |
 | --------- | ----- |
 | Required | No |
-| Cardinality | 0..* |
+| Cardinality | 0..1 <br> NOTE: The cardinality was incorrectly printed as 0..\* in the original publishing of the 2.3 specification |
```
goneall, line 1448: One very minor formatting suggestion is to [italicize](https://www.markdownguide.org/basic-syntax/#italic) the note (e.g. add an underscore before `Note` and after `specification`).

goneall (approved):
Thanks @rnjudge 

One minor optional suggestion for formatting, otherwise looks good

chapters/package-information.md
rnjudge: Sounds good. I'll update that.

goneall:
Looks good @rnjudge - thanks!

PR #821, opened 2022-12-05
Update conformance.md

hackrush01:
Updated conformance.md to point to the correct Annex for difference between v2.3 and v2.1.


goneall (approved):
Good catch @hackrush01 - thanks

goneall (approved):
Thanks @hackrush01

goneall (approved):

Issue #105, opened 2018-12-28
Add country or region of origin as an optional field


chandanbn:
Please add an optional field to declare the country or region(s) of origin or where the software as made. The value may be an array of ISO country codes.

Some countries may legally require vendors to add backdoors, weaknesses and vulnerabilities (see recent news from Southern Hemisphere). Identifying software originating from such countries enables closer scrutiny or taking additional precautionary steps. Though one should taking this information with a grain of salt and scrutinize every package prior to use for backdoors and vulnerabilities regardless of region of origin. Having this field may allow some automation in prioritization or policy enforcement.

kestewart:
Optional field at package level suffice?   I don't think we want this at the file level.
Also,  not sure how one counts for countries/regions when code is developed in open source and actual nationality of contributors is not tracked.   Maybe an "OPENSOURCE" type of option?    Thoughts?

swinslow:
I don't tend to think that this should be added as a separate field in the Package definition (and definitely agree that it wouldn't work at the File level).

As @kestewart noted, in an open source context, contributors may be from any number of countries, and might not share their countries of origin. Additionally, the source code could be stored, and executables could be built, on servers that are located across the globe. Given all of that, I'm not sure that a Country of Origin field would be meaningful here.

That said, if someone does want to track this information for an SPDX document they are preparing, there are several existing fields where it could be stored -- e.g., **3.12: Source Information**:

> This field provides a place for the SPDX file creator to record any relevant background information or additional comments about the origin of the package. For example, this field might include comments indicating whether the package was pulled from a source code management system or has been repackaged.

chandanbn:
While it may be difficult for large, legacy projects to make a statement, it helps software developers and software suppliers who want to be transparent about where their software was written or built. IMHO applies at file level as well as a package level, as different parts of a software project may have been developed in different places. It is also perfectly acceptable to have "**unknown**", "**various countries**" or an array of country codes as possible values. At the package level it would be simply an aggregation of values at file level. It is also perfectly ok to have source code and complied code to have different country of origins.

There are plenty of examples where organizations proudly make such an attestation about origin of their products:
https://www.bitmi.de/leistungen/software-made-in-germany
https://www.openeye.net/made-in-the-usa
https://www.apple.com/designed-by-apple

Not being able to exchange such software related metadata in SPDX could be hindering software transparency.

Much of this data is perhaps can be automatically gathered and shared as demonstrated by github:
https://octoverse.github.com/people#location

[...]

PR #822, opened 2022-12-05
Added Country of origin

riteshsonawane1372:
Added countryOfOrigin in the spdx-schema.json in creationInfo. It is important to know from where this doc was created

Signed-off-by: Ritesh Sonawane <riteshsonawane1372@gmail.com>

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -58,6 +58,10 @@
             "type" : "string"
           }
         },
+        "countryOfOrigin":{
+          "type":"string",
+          "description": "Country of origin, or where this lisence was created"
+        },
         "licenseListVersion" : {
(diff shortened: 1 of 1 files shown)


zvr:
@riteshsonawane1372 this is not the way to change the specification.

Instead of changing only one of the machine-generated schemas, the ontology has to be changed. In conjunction with that, the actual specification text also has to be changed.

In order to get consensus on the topic and the proposed solution, please fist raise an issue and be prepared to discuss it in one of the Tech Team weekly meetings. Since you are interested in the subject, I assume you are already familiar with the previous discussion (e.g. in #105, raised 4 years ago).

riteshsonawane1372:
@zvr Thank you so much

PR #828, opened 2023-01-10
Add pkg verification code + checksum clarification

rnjudge:
There had been some confusion around when to use the Package verification code field vs Package checksum. A small group of us got together and tried to provide more clarity around the two fields in a new section, Annex K.3 Verifying SPDX Packages. We also updated the `Description` and `Intent` fields for sections 7.9 and 7.10.

Signed-off-by: Jeff Schutt <jefschut@cisco.com>
Signed-off-by: Ivana Atanasova <iyovcheva@vmware.com>
Signed-off-by: Nisha Kumar <nisha@ctlfsh.tech>
Signed-off-by: Gary ONeall <garysourceauditor@gmail.com>
Signed-off-by: Adolfo Veytia <puerco@chainguard.dev>

### DIFF
--- a/chapters/package-information.md
+++ b/chapters/package-information.md
@@ -629,7 +629,7 @@ EXAMPLE 2 RDF: Property `spdx:filesAnalyzed` in class `spdx:Package`
 
 ### 7.9.1 Description
 
-This field provides an independently reproducible mechanism identifying specific contents of a package based on the actual files (except the SPDX document itself, if it is included in the package) that make up each package and that correlates to the data in this SPDX document. This identifier enables a recipient to determine if any file in the original package (that the analysis was done on) has been changed and permits inclusion of an SPDX document as part of a package. The metadata for the package verification code field is shown in Table 21.
+This field provides an independently reproducible mechanism that permits unique identification of the contents of the files contained within a package (except the SPDX document itself, if it is included in the package). The metadata for the package verification code field is shown in Table 21. For more specific use cases and examples of how to apply the package verification code field, see [Annex K.3 Verifying SPDX Packages](how-to-use.md#k3-verifying-spdx-packages).
 
 **Table 21 — Metadata for the package verification code field**
 
@@ -662,7 +662,7 @@ Required sort order: '0','1','2','3','4','5','6','7','8','9','a','b','c','d','e'
 
 ### 7.9.2 Intent
 
-Provide a unique identifier based on the files inside each package, eliminating confusion over which version or modification of a specific package the SPDX document refers to. This field also permits embedding the SPDX document within the package without altering the identifier.
+Verify, based on the files inside each package, which version or modification of a specific package the SPDX document references. This field also permits embedding the SPDX document within the package without altering the verification. This verification enables a recipient to determine which, if any, files in the original package have been changed.
 
 ### 7.9.3 Examples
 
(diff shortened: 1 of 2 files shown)


rnjudge:
You can view the markdown version of the changes here:
- https://github.com/rnjudge/spdx-spec/blob/clarify-pvc-pcs/chapters/how-to-use.md#k3-verifying-spdx-packages
- https://github.com/rnjudge/spdx-spec/blob/clarify-pvc-pcs/chapters/package-information.md#710-package-checksum-field-
- https://github.com/rnjudge/spdx-spec/blob/clarify-pvc-pcs/chapters/package-information.md#79-package-verification-code-field-

chapters/package-information.md
```diff
@@ -662,7 +662,7 @@ Required sort order: '0','1','2','3','4','5','6','7','8','9','a','b','c','d','e'
 
 ### 7.9.2 Intent
 
-Provide a unique identifier based on the files inside each package, eliminating confusion over which version or modification of a specific package the SPDX document refers to. This field also permits embedding the SPDX document within the package without altering the identifier.
+Identify, based on the files inside each package, which version or modification of a specific package the SPDX document references. This field also permits embedding the SPDX document within the package without altering the identifier. This identifier enables a recipient to determine which, if any, files in the original package have been changed.
```
goneall, line 665: Suggest replacing Identify with Verify and identifier with verification - checksums with potential collisions should not be used as identifiers, but could be used as verification.  Note that this issue is also present in the previous text.

chapters/package-information.md
```diff
@@ -708,7 +708,7 @@ Provide an independently reproducible mechanism that permits unique identificati
 
 ### 7.10.2 Intent
 
-Eliminate confusion over which version or modification of a specific package the SPDX document references by providing a unique identifier of the package.
+Identify, based on the binary blob, which version or modification of a specific package the SPDX document references. This identifier enables a recipient to determine if any bits in the original package have been changed but does not specify which bits.
```
goneall, line 711: Similar comment as above - replace "Identify" with "Verify" and "identifier" with "verification".

goneall (changes requested):
Just a couple minor comments - which admittedly apply to the previous text as much as it applies to the proposed changes.

chapters/package-information.md
```diff
@@ -629,7 +629,7 @@ EXAMPLE 2 RDF: Property `spdx:filesAnalyzed` in class `spdx:Package`
 
 ### 7.9.1 Description
 
-This field provides an independently reproducible mechanism identifying specific contents of a package based on the actual files (except the SPDX document itself, if it is included in the package) that make up each package and that correlates to the data in this SPDX document. This identifier enables a recipient to determine if any file in the original package (that the analysis was done on) has been changed and permits inclusion of an SPDX document as part of a package. The metadata for the package verification code field is shown in Table 21.
+This field provides an independently reproducible mechanism that permits unique identification of a package and the files contained within it (except the SPDX document itself, if it is included in the package). The metadata for the package verification code field is shown in Table 21. For more specific use cases and examples of how to apply the package verification code field, see [Annex K.3 Verifying SPDX Packages](how-to-use.md#k3-verifying-spdx-packages).
```
goneall, line 632: Suggest changing `This field provides an independently reproducible mechanism that permits unique identification of a package` to `This field provides an independently reproducible mechanism that permits unique identification of the contents a package`.  The reason for this change is the SPDX Package is defined not only by it's content but also the metadata associated with the content.  One example is that a supplier can choose to distribute the exact same content under different licenses to different downstream packages.  In SPDX these would be considered different packages even though the content is the same.  To uniquely identify the package, you would need to include the metadata as well as the content in the algorithm for the fingerprint.

chapters/package-information.md
```diff
@@ -708,7 +708,7 @@ Provide an independently reproducible mechanism that permits unique identificati
 
 ### 7.10.2 Intent
 
-Eliminate confusion over which version or modification of a specific package the SPDX document references by providing a unique identifier of the package.
+Verify, based on the binary blob, which version or modification of a specific package the SPDX document references. This verification enables a recipient to determine if any bits in the original package have been changed but does not specify which bits.
```
goneall, line 711: Propose changing `Verify, based on the binary blob, which version or modification of a specific package the SPDX document references.` to `Verify, based on the binary blob, which version or modification of a specific package content the SPDX document references.`

Same reasoning as above

goneall (changes requested):
3 suggested changes to clearly distinguish the Package content from the Package.

chapters/package-information.md
rnjudge: @goneall does this work? `This field provides an independently reproducible mechanism that permits unique identification of the files contained with a package (except the SPDX document file itself, if it is included in the package).` I would like to keep the clarification about the SPDX document itself not being included because that is a driver for having the two different PVC and PCS fields.

goneall: @rnjudge - Your suggestion is fine - but if we want to be completely accurate, we could say `This field provides an independently reproducible mechanism that permits unique identification of the contents of the files contained with a package (except the SPDX document file itself, if it is included in the package).`

kestewart:
Per discussion in call, no objections to it being merged.

PR #832, opened 2023-01-24
fix swid example to be a proper URI

armintaenzertng:
The example was missing the `swid:` prefix.
I also changed the formatting to use the one found in other examples like in the Package-Manager section.


goneall (approved):
LGTM

Thanks @armintaenzertng

goneall:
Note that this section has been removed in 3.0 - merging in case we do a 2.X dot release.

PR #833, opened 2023-01-24
replace soft hyphens in cpe23Type section

armintaenzertng:
The regex and example in the "cpe23Type" section used soft hyphens. I replaced them with normal dashes.

Signed-off-by: Armin Tänzer <armin.taenzer@tngtech.com>


goneall (approved):
LGTM

PR #836, opened 2023-02-22
fix: add correct referenceCategory values to the 2.2 JSON schema

### LINKED ISSUE #798: Discrepancy between docs and json schema on PACKAGE_MANGER vs. PACKAGE-MANAGER
On both the docs for SPDX 2.2 and 2.3 there are discrepancies with the json schema.

https://spdx.github.io/spdx-spec/package-information/ (2.2) and https://spdx.github.io/spdx-spec/v2.3/package-information/ under Table 33:

```
Format | <category> <type> <locator>where:<category> is SECURITY \| PACKAGE-MANAGER \| PERSISTENT-ID \| OTHER<type> is one of the types listed in Annex F.<locator> is the unique string with no spaces necessary to access the package-specific information, metadata, or content within the target location. The format of the locator is subject to constraints defined by the <type>.
```

On SPDX 2.2.2 schema - https://github.com/spdx/spdx-spec/blob/development/v2.2.2/schemas/spdx-schema.json:

```
                "referenceCategory" : {
                  "description" : "Category for the external reference",
                  "type" : "string",
                  "enum" : [ "OTHER", "PERSISTENT_ID", "SECURITY", "PACKAGE_MANAGER" ]
                },
```

on SPDX 2.3.1 schema:

```
                "referenceCategory" : {
                  "description" : "Category for the external reference",
                  "type" : "string",
                  "enum" : [ "OTHER", "PERSISTENT-ID", "PERSISTENT_ID", "SECURITY", "PACKAGE-MANAGER", "PACKAGE_MANAGER" ]
                },
```

The 2.3.1 json schema looks the most correct but the docs aren't correct. Similarly, from a cursory glance of the XML examples they also have their own format doing camel casing. If there are standard rules for SPDX format vs. JSON vs. XML it's unclear from the documentation or schemas.

kzantow:
This PR adds `PERSISTENT-ID` and `PACKAGE-MANAGER` as valid values for `referenceCategory` to the v2.2 JSON schema, which aligns more correctly with the SPDX spec and the 2.3 schema.

I'm not sure if this is the correct base branch to target -- please let me know if this should be changed!

This may fix #798

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -317,7 +317,7 @@
                 "referenceCategory" : {
                   "description" : "Category for the external reference",
                   "type" : "string",
-                  "enum" : [ "OTHER", "PERSISTENT_ID", "SECURITY", "PACKAGE_MANAGER" ]
+                  "enum" : [ "OTHER", "PERSISTENT-ID", "PERSISTENT_ID", "SECURITY", "PACKAGE-MANAGER", "PACKAGE_MANAGER" ]
                 },
                 "referenceLocator" : {
                   "description" : "The unique string with no spaces necessary to access the package-specific information, metadata, or content within the target location. The format of the locator is subject to constraints defined by the <type>.",


kzantow:
@goneall please let me know if this is the wrong base branch to target for this change

goneall (approved):
Thanks @kzantow 

Changes LGTM

goneall:
@tsteenbe @maxhbr @nikoweider @meretp @puerco @RishabhBhatnagar - as producers and consumers of the JSON format - any concerns with merging this into the 2.2.2 version of the JSON schema?

maxhbr (commented):
I am afraid that some tools that automatically generate code from the enum might now get confused in a different way. Although I have no example in mind

goneall:
>

> I am afraid that some tools that automatically generate code from the enum might now get confused in a different way. Although I have no example in mind

Agree - although this would also be true of the 2.3 version of the schema which has the same duplicate enums

MP91:
Hey @goneall I don't know your procedure on PRs but is there any timeframe when this gets merged?

We also have this problem in our tooling, which is expecting PACKAGE_MANAGER...

puerco:
Just as with `PACKAGE_MANAGER` and others before, I've found that the least disruptive way to do it is this one (have both). I share Max's concern but I it think this is the least harmful of our options. So lgtm from me :+1:

goneall:
@MP91 - It is now merged

MP91:
@goneall  Thank you for the fast response!

PR #837, opened 2023-02-28
Update reference

zvr:
- Updates reference from 2.2 to 2.3


goneall (approved):
Good catch - thanks @zvr

PR #838, opened 2023-02-28
ExceptionRef

zvr:
- Adds ExceptionRef- syntax


zvr:
Closing this, as the decision was to go with `AdditionRef-`, which is implemented in #839

PR #841, opened 2023-03-01
chore: deprecate shorthand properties

kzantow:
This PR adds deprecation to the "shorthand" properties in the schema and marks previously deprecated fields as such with the `deprecated` property.

Based on the SPDX WG meeting 2023/02/28, these fields will not be included in SPDX 3.0 and we want to indicate they are deprecated in the latest SPDX 2.3 JSON schema.

cc: @goneall @lumjjb

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -1,5 +1,5 @@
 {
-  "$schema" : "http://json-schema.org/draft-07/schema#",
+  "$schema" : "https://json-schema.org/draft/2019-09/schema#",
   "$id" : "http://spdx.org/rdf/terms/2.3",
   "title" : "SPDX 2.3",
   "type" : "object",
@@ -217,7 +217,9 @@
       "description" : "The URI provides an unambiguous mechanism for other SPDX documents to reference SPDX elements within this SPDX document."
     },
     "documentDescribes" : {
-      "description" : "Packages, files and/or Snippets described by this SPDX document",
+      "description" : "DEPRECATED: use relationships instead of this field. Packages, files and/or Snippets described by this SPDX document",
+      "deprecated": true,
+      "$comment": "This field has been deprecated as it is a duplicate of using the SPDXRef-DOCUMENT DESCRIBES relationship",
       "type" : "array",
       "items" : {
         "type" : "string",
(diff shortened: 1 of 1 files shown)


schemas/spdx-schema.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema" : "http://json-schema.org/draft-07/schema#",
+  "$schema" : "https://json-schema.org/draft/2019-09/schema#",
```
kzantow, line 2: This had to get updated because the `2019-09` schema is the first to support the `deprecated` property.

schemas/spdx-schema.json
```diff
@@ -343,7 +345,9 @@
             "type" : "boolean"
           },
           "hasFiles" : {
-            "description" : "Indicates that a particular file belongs to a package.",
+            "description" : "DEPRECATED: use relationships instead of this field. Indicates that a particular file belongs to a package.",
+            "deprecated": true,
+            "$comment": "This field has been deprecated as it is a duplicate of using SPDXRef-<package-id> CONTAINS SPDXRef-<file-id> relationships",
```
lumjjb, line 350: nit: for consistency, maybe just SPDXRef-PACKAGE CONTAINS?

lumjjb (approved):
lgtm. minor nits

schemas/spdx-schema.json
kzantow: If you're referring to the `SPDXRef-DOCUMENT` for the other deprecation text, I don't think there is a consistency problem because that is the actual value, whereas the `SPDXRef-<package-id>` needs to match the containing package, but I'm happy to change this if people feel strongly. I'll see if I can find a better way to identify this.

kzantow: I did update the text here a bit, hopefully it's clear enough.

schemas/spdx-schema.json
```diff
@@ -217,7 +218,9 @@
       "description" : "The URI provides an unambiguous mechanism for other SPDX documents to reference SPDX elements within this SPDX document."
     },
     "documentDescribes" : {
-      "description" : "Packages, files and/or Snippets described by this SPDX document",
+      "description" : "DEPRECATED: use relationships instead of this field. Packages, files and/or Snippets described by this SPDX document",
```
goneall, line 221: How about "use CONTAINS relationships from the SPDXRef-DOCUMENT to the SPDX ID's in the documentDescribes list"?

goneall: Nevermind - I see the `$comment` below which does have the specifics

goneall (approved):
LGTM - @lumjjb if it looks good to you, I'll update the tools to generate the same changes and merge in this PR.

goneall:
ping @lumjjb - does the latest changes look good?  If so, I'll merge and update the tools

lumjjb:
yea LGTM!

goneall:
I'll go ahead and merged.

I opened https://github.com/spdx/spdx-java-jackson-store/issues/54 to track updating the Java tools.

@armintaenzertng do we need to open a similar issue for the Python tools?

@lumjjb @kzantow I'll assume you have the GoLang tools covered.

armintaenzertng:
Thanks for the heads-up, Gary, but the python tools should be fine as we only support these fields during parsing, which we will most likely continue to do.

PR #861, opened 2023-08-22
update cardinality text for omitted optional fields

lumjjb:
As a follow up based on conversations on the mailing list (https://lists.spdx.org/g/Spdx-tech/topic/100823660), and on today's tech call 2023-08-22, ([minutes](https://spdx.swinslow.net/p/spdx-tech-minutes)) - . This PR is to add additional text to communicate that omission of fields with optional cardinality should be treated as signaling NOASSERTION.

### DIFF
--- a/chapters/conformance.md
+++ b/chapters/conformance.md
@@ -21,7 +21,7 @@ This standard contains more than a few *cardinality assertions*, each of which i
 * Cardinality: 1..1
 * Cardinality: 1..\*
 
-Each of these assertions can easily be understood as to whether a feature is required, and if so, how many occurrences are required; also, whether a feature is permitted, and if so, in what number. As this is the format long familiar to the SPDX community, it has been preserved in this document.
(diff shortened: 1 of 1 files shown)


bact:
Do we like to also have this proposed text:

> Unless specified otherwise, omission of an optional field should be interpretted as signaling NOASSERTION.

in the post-v3.0 version?

bact (approved):

goneall:
Discussed on the 15 Oct 2024 tech call - there were a couple concerns with documenting this for the 3.0 spec:

- NOASSERTION is declaring a known unknow and we may loose the semantics of this being an intentional declaration
- NOASSERTION in 3.0 is actually for an Object property with a range of ELEMENT

Leaving this issue open for further discussion.

bact:
- Related issue #896

goneall (approved):
LGTM

kestewart (approved):

PR #865, opened 2023-10-17
feat: allow $schema entry in SPDX 2.3 JSON

### LINKED ISSUE #864: Attribute `$schema` not allowed within JSON schema
It is currently a schema violation to reference the JSON schema from within an SPDX JSON document. Adding a field `$schema` will cause a schema violation. See screenshot of an example SPDX document in VisualStudio Code.

image

mcombuechen:
This adds an allowed field `$schema` to the JSON schema, which can be used to reference the SPDX 2.3 JSON schema.
Ideally this would be an enum with a list of allowed locators, but since the schema is not hosted anywhere other than in GitHub (not to my knowledge), I left it as a free-form string.

Closes #864.

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -4,6 +4,10 @@
   "title" : "SPDX 2.3",
   "type" : "object",
   "properties" : {
+    "$schema": {
+      "type": "string",
+      "description": "Reference the SPDX 2.3 JSON schema."
+    },
     "SPDXID" : {
       "type" : "string",
       "description" : "Uniquely identify any element in an SPDX document which may be referenced by other elements."


zvr:
Since the JSON schema is automatically generated, shouldn't the change be done in https://github.com/spdx/tools-java/blob/master/src/main/java/org/spdx/tools/schema/OwlToJsonSchema.java ?

mcombuechen:
Thanks @zvr and @goneall , I was not aware of the auto-generation. Will look into adding the change there.

mcombuechen:
However, it seems that latest changes to the file have been made without changes elsewhere; suggesting that it's not auto-generated and maintained manually? https://github.com/spdx/spdx-spec/commits/development/v2.3.1/schemas/spdx-schema.json

Or am I getting this wrong?

goneall:
> Since the JSON schema is automatically generated, shouldn't the change be done in https://github.com/spdx/tools-java/blob/master/src/main/java/org/spdx/tools/schema/OwlToJsonSchema.java ?

Thanks @zvr  for catching this.

We have generated the spec in the past, but the last few commits were merged in manually.  Once the 3.0 spec is released, we'll go back to automatically generating the schema.

At the same time, we should update the utility to generate the correct schema in case we want to do another dot release of the spec.

I added https://github.com/spdx/tools-java/issues/144 to track the tools update.

@zvr - we can probably merge this in if you're OK with the schema change.

mcombuechen:
@zvr could you have another look? 🙏

zvr (approved):

goneall:
Hmm. It looks like the CI is failing due to some dependency update issues unrelated to this PR.

I'll go head and merge this and add an issue to look into the CI failures.

PR #879, opened 2024-02-05
Temporarily remove SPDX-Lite Annex for RC2 release

goneall:
Based on the tech call discussion on Annexes, this will be updated for the next release candidate after RC2 since it depends on a Tag/Value like format
See:
https://docs.google.com/spreadsheets/d/1QHp5QwmBlsmd2iXNy-Y7K6oka_u5qc51iR6Xr0s0Qao/edit#gid=0


kestewart:
Not convinced it depends on TagValue @goneall - but we need to since I can see this being made available in JSON, etc. as well.   I updated the title to reflect it's just the ANNEX,  not the profile itself that is being removed.   I do agree that the annex needs to be rewritten, to pick up the accurate references, and that can happen after rc2.

kestewart (changes requested):
Annex needs to be rewritten,  and agree it is not critical path for rc2, so best to drop for now, and re add later. 

Rather than drop though,  since there are references to it through the text,  how about leave the header intact, and put a "will be documented in final release"?    That way the linkages won't break.   Otherwise we'll have to do a pass through the text and make sure we remove all Annex refs as well.

goneall:
> Annex needs to be rewritten, and agree it is not critical path for rc2, so best to drop for now, and re add later.
> 
> Rather than drop though, since there are references to it through the text, how about leave the header intact, and put a "will be documented in final release"? That way the linkages won't break. Otherwise we'll have to do a pass through the text and make sure we remove all Annex refs as well.

@kestewart - If there are references, I agree it would be better.  I checked and the Lite profile in the model does not reference the annex.  Is there any other possible references?  If not, we can safely delete then add back later.

goneall:
This is superseded by PR #883

PR #883, opened 2024-02-15
Remove files no longer used in the 3.0 RC2 version of the spec

### LINKED ISSUE #881: Clean up development/v3.0 file structure and README
There are some duplicate files in the chapters directory and in the docs directory.

The README doesn't indicate what folders are to be maintained.

Note that the chapters directory has been updated with some of the decisions from the tech team.

If these changes should be applied to the docs directory, we should synchronize, move over the updated files.

goneall:
Fixes #881


goneall:
@licquia @zvr - Please review - I tried to clean this up based on the information I have, but it looks like there is a lot of additional changes that need to be made to mkdocs configs etc.

goneall:
@licquia If this doesn't make the mkdocs work any worse, I'd like to merge this in - let me know.

kestewart (approved):
Thanks Gary,  lots of changes here, but from a review, I think you've caught most of what's been discussed.   @licquia - can you take a look this doesn't cause problems for the work you've been doing, and then we can merge.

zvr (approved):
I'd also delete the `examples` subdirectory (and its `README`), since they are not part of the docs.
And rename the `RDF...` annex to be about "graphical representations" or something.

But we can do all this via another PR, after also updating mkdocs configuration.

So, let's go ahead and merge this.

goneall:
I'm going to go ahead and merge this and see if I can help @licquia with the PDF / HTML generation.

PR #886, opened 2024-02-18
Sync the model diagrams with the model repo

goneall:
This updates the diagrams with the model repo after merging https://github.com/spdx/spdx-3-model/pull/643


docs/images/model-security-profile.png
kestewart, line 1: Double check,  this is looking like the core, not security.   Also, is this reducnant with model Security.png.

kestewart (changes requested):
Not sure why model-security-profile.png is included when there is a "model Secuirty.png".   Included by error?   Also file looks like Core+Software not security.

docs/images/model-security-profile.png
goneall: Removed the extra image.

I noticed this too while on the plane.

The image is present in the model repo - it is stale compared to model Security.png.

BTW - The image was not used in the website or PDF.

goneall:
Resolved @kestewart feedback, merging

PR #887, opened 2024-02-22
Add 3.0rc2 Ontology

zvr:
Adding automatically generated RDF ontology files (in a variety of formats)


goneall (approved):
I didn't do a thorough review, but overall it looks good.

Only comment is the visualization (dot file) is very dense - we may want to improve the visuals in the future.

Thanks @zvr

Issue #62, opened 2018-01-02
Simplify license expression grammar in Appendix IV


kestewart:
This was https://bugs.linuxfoundation.org/show_bug.cgi?id=1334

**Kate Stewart 2015-12-08 16:57:25 UTC**
From David Wheeler on December 4, 2015

The current Appendix IV is also overly complex and confusing:

* There's no need to have "compound-expression" as separate from "license-expression".  The "license-expression" is defined to be either simple or compound, but a simple-expression is also a legal compound-expression, so the whole indirection is unnecessary and confusing.

* In simple-expression, the "+" should just optionally follow license-id; that's how anyone would parse it, and it's easier to explain too.

So I suggest replacing simple-expression, compound-expression (to be removed), and license-expression with this simpler spec:

```
simple-expression = license-id ["+"] / license-ref

license-expression =   simple-expression [ "WITH" license-exception-id ] /

  license-expression "AND" license-expression /

  license-expression "OR" license-expression /

  "(" license-expression ")"

```
 
You could change simple-expression to be:

simple-expression = license-id ["+"][ "WITH" license-exception-id ] / license-ref and omit the ["WITH...] in the following line, but I *like* the idea of allowing a license-ref with a standard exception. Besides, that's currently allowed, no reason to *remove* this functionality.

Both this and the original description are silent about left-to-right or right-to-left; I don't think it matters, but if someone wants things to be parsed identically, perhaps that should be mentioned.

I can imagine adding suffixes like "!" (I'm *sure* it's only this particular version of the license) or "?" (I'm *not* sure that it's limited to this particular version of the license), in addition to "+".

However, that's a separate discussion.

Also: is there any reason to *FORBID* the "+" suffix after a license-ref or license-exception-id?

In particular, someone might use a license-ref while waiting for a license to be added to the SPDX license list or exception list.

A way Would change my proposal grammar above to:

```
simple-expression = license-id / license-ref

license-expression =   simple-expression ["+"] [ "WITH" license-exception-id ["+"] ] /
```

**Kate Stewart 2015-12-15 18:34:44 UTC**
From discussion on 20151215 - Mark wants to confirm that the revised version still works in his encoding.   Other than that,  simpler is better, so once its proven out, we'll look at changing this in the 2.1 spec.

**Bill Schineller 2016-05-17 17:35:11 UTC**
Mark?

kestewart:
Note:  There has been recent discussion of the grammar in 2017 that needs to be added here.

[...]

PR #889, opened 2024-02-26
Remove compound-expression from ABNF grammar

sertonix:
`compound-expression` was equivalent to `license-expression`. Also fixing a small formatting issue.

Ref #62, #456

### DIFF
--- a/chapters/SPDX-license-expressions.md
+++ b/chapters/SPDX-license-expressions.md
@@ -19,18 +19,15 @@ license-ref = ["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring)
 
 simple-expression = license-id / license-id"+" / license-ref
 
-compound-expression = (simple-expression /
+license-expression = (simple-expression /
 
+  simple-expression "WITH" license-exception-id /
 
-simple-expression "WITH" license-exception-id /
+  license-expression "AND" license-expression /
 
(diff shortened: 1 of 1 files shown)


goneall:
Thanks @sertonix for the pull request.

Can you apply this to the development/v3.0 branch and sign off so it passes the DCO?

sertonix:
> The text below follows this convention, talking about simple and compound/complex expressions.

Couldn't the text talk about simple expressions and license expressions instead?

goneall:
> @goneall I don't agree this is needed.
> 
> Yes, `license-expression` can be expanded to the same as `compound-expression`, but this was chosen to illustrate the different types of expressions. The text below follows this convention, talking about simple and compound/complex expressions.

This makes sense - I'm OK with the way it is (even though, technically, it can be removed).

@sertonix if you're OK with this, we can close this PR (and you don't need to rebase).

zvr:
> Couldn't the text talk about simple expressions and license expressions instead?

But "simple expressions" _are also_ "license expressions" -- they are not two distinct things. 

Having the grammar was confusing enough for lawyers who have to read this (and who wrote this), so we wanted to make it crystal clear and easy to follow.

goneall:
Going to go ahead and close this per @zvr comment above

PR #893, opened 2024-03-26
Generate JSON Schema using shacl2code

JPEWdev:
Generates the JSON schema for validating JSON-LD documents using `shacl2code`

### DIFF
--- a/.github/workflows/publish_v3.yml
+++ b/.github/workflows/publish_v3.yml
@@ -32,6 +32,16 @@ jobs:
         run: pip install -r spec-parser/requirements.txt
       - name: Build model
         run: python3 spec-parser/main.py spdx-3-model/model spdx-spec/docs/model
+      - name: Copy JSON annotations
+        run: cp spdx-spec/serialization/jsonld/annotations.ttl spdx-spec/docs/model/jsonld-annotations.ttl
+      - name: Generate JSON schema
+        run: |
+          shacl2code generate \
+            --input spdx-spec/docs/model/ontology.rdf.ttl \
+            --input spdx-spec/docs/model/jsonld-annotations.ttl \
+            --context-url spdx-spec/docs/model/context.jsonld https://spdx.org/rdf/3.0.0/spdx-context.jsonld \
+            jsonschema \
+            --output spdx-spec/docs/model/schema.json
       - name: Set git identity
         run: git config user.name ci-bot; git config user.email ci-bot@spdx.dev
         working-directory: spdx-spec
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ mkdocs==1.5.3
 json-schema-for-humans==0.47
 mkdocs-pdf-export-plugin==0.5.10
 mike==1.1.2
+shacl2code==0.0.9
--- a/serialization/jsonld/annotations.ttl
+++ b/serialization/jsonld/annotations.ttl
@@ -1,0 +1,10 @@
+@base <https://rdf.spdx.org/v3/> .
+@prefix sh-to-code: <https://jpewdev.github.io/shacl2code/schema#> .
+
+<Core/Element> ;
+    sh-to-code:idPropertyName "spdxId"
+    .
+
+<Extension/Extension> ;
+    sh-to-code:isExtensible true
+    .


JPEWdev:
@licquia  Here is the actions to generate the JSON schema. Note that the schema is still incorrect until the model & spec parser are fixed

goneall (approved):
Thanks @JPEWdev - This will be quite helpful for validation

.github/workflows/publish_v3.yml
```diff
@@ -9,6 +9,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      VERSION: v3.0
```
maxhbr, line 13: maybe no discussion for this PR: we will have patch releases? Shouldn't that be v3.0.0? Especially in the persistent URL for the context.

shacl2code-annotations.ttl
maxhbr, line 1: top level seems to be a rather prominent place for such a technical file. Can we move it somewhere else?

.github/workflows/publish_v3.yml
```diff
@@ -32,12 +34,20 @@ jobs:
...
+            --input spdx-spec/shacl2code-annotations.ttl \
+            --context-url spdx-spec/docs/model/context.jsonld https://spdx.github.io/spdx-spec/${{ env.VERSION }}/model/context.jsonld \
+            jsonschema \
+            --output spdx-spec/docs/model/schema.json
       - name: Set git identity
         run: git config user.name ci-bot; git config user.email ci-bot@spdx.dev
         working-directory: spdx-spec
       - name: Sync gh-pages
-        run: git checkout gh-pages && git pull && git checkout development/v3.0
+        run: git checkout gh-pages && git pull && git checkout development/${{ env.VERSION }}
```
maxhbr, line 49: This is messy code and a cleaner way of doing that might be
```suggestion
        run: |
          git fetch origin
          git branch -f gh-pages origin/gh-pages
```

.github/workflows/publish_v3.yml
```diff
@@ -32,12 +34,20 @@ jobs:
...
         working-directory: spdx-spec
       - name: Build docs
-        run: mike deploy v3.0 v3-draft -b gh-pages -p
+        run: mike deploy ${{ env.VERSION }} v3-draft -b gh-pages -p
```
maxhbr, line 52: `VERSION` is now a variable, but `v3-draft` is still hardcoded

.github/workflows/publish_v3.yml
maxhbr: `v3.0` is still hard coded in multiple places, e.g. line 18

maxhbr (changes requested):

.github/workflows/publish_v3.yml
maxhbr: as `development/v3.0` is hardcoded in line 18, making it dependent on the variable here causes inconsistencies

JPEWdev: Reverted to hardcoded since it's inconsistent

shacl2code-annotations.ttl
JPEWdev: Done

.github/workflows/publish_v3.yml
```diff
@@ -32,6 +32,16 @@ jobs:
...
       - name: Build model
         run: python3 spec-parser/main.py spdx-3-model/model spdx-spec/docs/model
+      - name: Copy JSON annotations
+        run: cp spdx-spec/serialization/jsonld/annotations.ttl spdx-spec/docs/model/jsonld-annotations.ttl
+      - name: Generate JSON schema
+        run: |
+          shacl2code generate \
+            --input spdx-spec/docs/model/ontology.rdf.ttl \
+            --input spdx-spec/docs/model/jsonld-annotations.ttl \
+            --context-url spdx-spec/docs/model/context.jsonld https://spdx.org/rdf/3.0.0/spdx-context.jsonld \
```
maxhbr, line 42: I am confused? why is there an argument before the url?
```suggestion
            --context-url https://spdx.org/rdf/3.0.0/spdx-context.jsonld \
```
Maybe I am just not aware of the API of your tool

maxhbr (approved):
One question, but looks fine

.github/workflows/publish_v3.yml
JPEWdev: The tool lets you either specify a URL (with --context) or a local file and URL (with --context-url) in which case the data is read from the local file, but the URL is used in the bindings. The later is useful to generate bindings before the URL is published

goneall:
@licquia @JPEWdev - Is this ready to be merged?

licquia (approved):

PR #903, opened 2024-04-09
Rename ISO_foreword.md to ISO-foreword.md

bact:
Use `-` (dash) instead of `_` (underscore) to keep it consistent with other docs filenames.


zvr:
This file is actually not relevant for 3.0, since it's not an ISO standard -- it can be safely deleted.
We should re-introduce it in the branch of the version that will go towards the ISO standardization process (3.0.x, probably).

bact:
Depends on how "clean" we like each branch to be, this file can also be here as well - so long as it does not get included in the mkdocs.yml.

[v2.3](https://github.com/spdx/spdx-spec/blob/development/v2.3/mkdocs.yml) and [v2.3.1](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/mkdocs.yml) branches also use this approach: keep the file in the source repo but not include in in the documentation generation recipe.

kestewart:
Have moved this to the 3.1 release as target.

PR #907, opened 2024-04-12
Adds Annex for the Lite profile

NorioKobota:
Adds the explanation for the Lite profile as Annex.

### DIFF
--- a/docs/annexes/SPDX-Lite.md
+++ b/docs/annexes/SPDX-Lite.md
@@ -1,0 +1,132 @@
+# Annex G SPDX Lite  
+
+## G.1 Definition of the Lite profile <a name="G.1"></a>  
+
+The Lite profile is designed to make it quick and easy to start a Software Bill of Materials in situations where a company may have limited capacity for introducing new items into its process.  
+The Lite profile captures the minimum set of information required for license compliance in the software supply chain. It contains information about the creation of the SBOM, package lists with licensing and other related items, and their relationships.  
+
+All elements in Lite profile are essential for complying with licenses. It is easy to use a SPDX document with the Lite profile for anyone who does not have enough knowledge about licensing information and easy to import license information from former versions of SPDX Lite format files.  
+The Lite profile offers the flexibility to be used either alone or in combination with other SPDX profiles as a SPDX document in the software supply chain.  
+
+## G.2 Table of the Lite profile elements <a name="G.2"></a>  
+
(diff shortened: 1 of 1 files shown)


kestewart (approved):
Thanks for all the hard work in getting this pulled together. 

This looks good to me, but I'd like either Gary or Alexios to review as well.

goneall (changes requested):
Overall, this looks quite good.

I do have one suggested change:

To process an SPDX Lite document, we'll need to know where to start - what package the Lite document is describing.

There are 2 options on SPDX 3.0:
- a relationship of type describes between the SPDX document and the package 
- rootElement property

The simplest option IMHO is rootElement.  For the Lite profile, I would suggest making this property required (cardinality 1..1) and have the package element be the value for the property.

NorioKobota:
@goneall, @kestewart  
Thanks for the review. I have a question to @goneall.
> The simplest option IMHO is rootElement.

The [SpdxDocument class has rootElement in this PR](https://github.com/spdx/spdx-spec/pull/907/files/c49fa3039798f3ab481ab9eae9618bf3a18c6dd5#diff-3456d2de18ba26f6cc290202bc6877e346223d775d081e67ca3ec55f373598afR27), but does the Sbom class also should have rootElement? Or does that mean rootElement in the SpdxDocument class should be 1..1?

Based on [this JSON-LD sample](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/main/jsonld/examples/package_sbom.json#L31), I think it's enough to have rootElement in SpdxDocument..

goneall:
> @goneall, @kestewart Thanks for the review. I have a question to @goneall.
> 
> > The simplest option IMHO is rootElement.
> 
> The [SpdxDocument class has rootElement in this PR](https://github.com/spdx/spdx-spec/pull/907/files/c49fa3039798f3ab481ab9eae9618bf3a18c6dd5#diff-3456d2de18ba26f6cc290202bc6877e346223d775d081e67ca3ec55f373598afR27), but does the Sbom class also should have rootElement? Or does that mean rootElement in the SpdxDocument class should be 1..1?
> 
> Based on [this JSON-LD sample](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/main/jsonld/examples/package_sbom.json#L31), I think it's enough to have rootElement in SpdxDocument..

Very good point about the SBOM.

I'm thinking that in a Lite document the document root element should point to the SBOM element collection and the SBOM root element should point to the package.  Perhaps make them both required and add some documentation on the best practice for what these fields should contain?

goneall (approved):
LGTM - Thanks @NorioKobota

PR #909, opened 2024-04-12
Add serialization info

zvr:
A short "chapter" that explains what the spec is about:
- we define the model in RDF
- it can be serialized in many ways
- we also provide a context file
- etc.

Presumably in the future we will be adding other sections about other (non-RDF) serialization formats.

### DIFF
--- a/docs/serializations.md
+++ b/docs/serializations.md
@@ -1,0 +1,36 @@
+# 4 Model and serializations
+
+## 4.1 Overview
+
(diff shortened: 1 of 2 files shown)


kestewart (approved):
Thanks Alexios,  this is good context info.

kestewart:
@rnjudge - could you take a review pass, and if you're good with it, merge please.

rnjudge:
LGTM

PR #915, opened 2024-04-14
Organize annexes

### LINKED ISSUE #478: Move Appendix G (CC License) and J (Differences From Earlier Versions) to be Appendix A and B
These two appendices will always exist so I suggest we move them to be the first two appendices to simplify future section renumbering.

rnjudge:
Updates the annex ordering.

Closes #478


goneall (approved):
LGTM

PR #924, opened 2024-04-21
Annex C: Change 3.0 Model URL to new one at spdx.org

bact:
- `https://spdx.github.io/spdx-3-model/model.ttl` -> `https://spdx.org/rdf/3.0.0/spdx-model.ttl`
  - New IRI according to [2024-04-02 Tech call](https://github.com/spdx/meetings/blob/main/tech/2024-04-02.md)

The `https://spdx.github.io/spdx-3-model/model.ttl` can probably a model in development (in 3.x.x) and not stable.

We may want to use the public-facing IRI: `https://spdx.org/rdf/3.0.0/spdx-model.ttl` (which will be redirected to `https://spdx.github.io/spdx-spec/v3.0/model/spdx-model.ttl`).

**Note: As the link is specifically to `3.0.0`, we have to update this for `3.0.1` and later**


goneall (approved):
Nice catch - Thanks @bact

bact:
Trying to do DCO, but I think I rebased this one wrongly.

I probably better close this PR and open new one. Don't want to risk messing up with history.

bact:
Close this one. Move to #936 instead.

PR #938, opened 2024-04-23
Validate documentation

JPEWdev:
Validates security documentation


goneall (approved):
Changes LGTM and it does pass the CI with the additional checking.

docs/annexes/cross-reference.md
```diff
@@ -0,0 +1,209 @@
+# Annex I: Cross referencing in SPDX 3 (Informative)
+
+This document will walk though how to refer to SPDX Elements across documents
+(e.g. cross reference).
+
+If you do would like to construct the complete example documents from this
+Markdown file, use the following command:
+
+```shell
+cat getting-started.md | awk 'BEGIN{flag=0} /^```json/, $0=="```" { if (/^---$/){flag++} else if ($0 !~ /^```.*/ ) print $0 > "doc-" flag ".spdx.json"}'
```
maxhbr, line 10:
```suggestion
cat cross-reference.md | awk 'BEGIN{flag=0} /^```json/, $0=="```" { if (/^---$/){flag++} else if ($0 !~ /^```.*/ ) print $0 > "doc-" flag ".spdx.json"}'
```

maxhbr:
I just added the functionality to load multiple documents to my tool https://github.com/maxhbr/spdx3ToGraph and this generates the following for the cross-reference.md example:

doc-0 spdx json

JPEWdev:
> I just added the functionality to load multiple documents to my tool https://github.com/maxhbr/spdx3ToGraph and this generates the following for the cross-reference.md example:
> 
> doc-0 spdx json

Very cool!

JPEWdev:
> I just added the functionality to load multiple documents to my tool https://github.com/maxhbr/spdx3ToGraph and this generates the following for the cross-reference.md example:
> 
> doc-0 spdx json

This made me realize I forgot the `serialziedInArtifact` relationship, so I added that

maxhbr:
update:

doc-0 spdx json

docs/annexes/cross-reference.md
```diff
@@ -0,0 +1,246 @@
...
+            "@id": "_:creationinfo",
+            "specVersion": "3.0.0",
+            "createdBy": [
+                "http://spdx.example.com/Person/JoshuaWatt"
+            ],
+            "created": "2024-03-06T00:00:00Z"
+        },
+        {
+            "type": "SpdxDocument",
+            "spdxId": "http://spdx.example.com/Document1",
```
maxhbr, line 51: I think we should use more realistic and potentially globally unique IDs.

JPEWdev: What classifies as "more realistic"?

maxhbr: something that could be globally unique. Maybe what I am missing is guidance on how to create IDs. I just know hand crafted non-realistic examples. Some using `urn:..`. I think this might be something for a different document but I it is a gap.

JPEWdev: Ya, I don't mind making it better, I'm just not clear what "better" would be in this case. There's not currently great guidance on this other than that it should be "unique"

goneall (approved):
Very nice explanation - Thanks @JPEWdev

kestewart (approved):
Thanks for adding this in Joshua!

PR #960, opened 2024-05-09
Fix ABNF for license expressions

### LINKED ISSUE #922: ABNF grammar of Annex D and case-sensitivity
Hey, I echo to #664 and the discussions on the ABNF grammar described as part of SPDX v3.0.

Given RFC 5234 Section 2.3 "Terminal Values":
> ABNF strings are case insensitive [...]

This imply that `"TEST"` could produce "test", "tEsT", "TEST", and so on.
As this was not clear for most of the reviewers and later on the users of the RFC, the IETF published RFC 7405.

Nevertheless, the current ABNF grammar contains the alternation group `( "WITH" / "with" )` that produces the same results.
I suggest this alternation group is either removed to enhance readability OR replaced by `( %s"WITH" / %s"with" )` as in my understanding it better fits the idea of rule `compound-expression`.
The same goes on for `AND` and `OR`.

Moreover, this should also be clarified for `"DocumentRef-"`, `"LicenseRef-"` and `"AdditionRef-"` which should be prefixed by `%s`.

zvr:
In the license expressions expressed in ABNF form, make explicit that strings (like `"AND"` and `"and"`) are case-sensitive. This was expressed in the text below, but not in the grammar.

Fixes #922


goneall (approved):
LGTM

Issue #937, opened 2024-04-23
ontology/ directory is no longer updated - suggest deletion


bact:
Content in https://github.com/spdx/spdx-spec/tree/development/v3.0.1/ontology is outdated.

One of the files (ontology/model.plantuml) was generated by spec-parser on 2024-02-22, before the release of 3.0 (2024-04-15).

It contains outdated concepts like `Software/isDirectory` and `Dataset/sensitivePersonalInformation`; also does not contain new concepts like `Dataset/DatasetPackage` and `AI/energyUnit`.

The ontology live on spdx.org is well updated. All these files are good:

- https://spdx.org/rdf/3.0.0/spdx-model.ttl
- https://spdx.org/rdf/3.0.0/spdx-context.jsonld
- https://spdx.org/schema/3.0.0/spdx-json-schema.json

Should we update this `ontology/` directory or should we just delete it to avoid confusion?

zvr:
Every run of the spec-parser generates an updated ontology in all formats, including the PlantUML diagram.

The generated files should be copied in the `ontology` directory.

bact:
How we are going to communicate about these "alternatives"? (Like, when to use which source)

- https://spdx.org/rdf/3.0.0/spdx-model.ttl (official)
- https://spdx.github.io/spdx-3-model/model.ttl
  - Same file as https://github.com/spdx/spdx-3-model/blob/gh-pages/model.ttl (`gh-pages` branch is suggested in [spdx/spdx-3-model README](https://github.com/spdx/spdx-3-model/blob/main/README.md) and the exact link is used in
[spdx-3-model/serialization/json-ld.md](https://github.com/spdx/spdx-3-model/blob/main/serialization/json-ld.md)
- https://github.com/spdx/spdx-spec/blob/development/v3.0.1/ontology/ontology.rdf.ttl

I realized that there should be at least two sources: one is stable (generated at release for public use), another is for development (generated frequently for testing/development).

The stable one should be at spdx.org and all spec/doc should point to that new IRIs.

The question is probably where should be the source for the development.
For example, in `spdx-spec` repo or in `spdx-3-model` repo.

Once decided to pick one, we should delete another to avoid confusion.
Also after that, README and etc should be updated to make it clear that this is unstable.

The current situation is a not very friendly for new comer.

The same concern goes to https://github.com/spdx/spdx-3-model/issues/726

goneall:
From the tech call - we agreed the ontology directory can be deleted.  

[...]

PR #963, opened 2024-05-21
Remove stale ontology

JPEWdev:


goneall:
Note that this fixes #937 

Based on the last comment from @zvr - there was a suggestion that we include this in the CI.

My opinion is we should merge this for now then add back a non-stale version using a CI change - so I'll approve the PR.

@zvr - let me know if you agree.

goneall (approved):
We can add these back once we have  CI that copies these over from the model.

zvr:
Sure!  I thought it would be easy to add a couple of `cp` lines, but since we don't have them yet, let's delete these for a start.

zvr (approved):

PR #964, opened 2024-05-21
annexes: getting started: Fix SPDX IDs

JPEWdev:
The SPDX IDs used in the example are fabricated, but not good examples to follow for users trying understand how to write their own documents. Fix this by using the "https://spdx.org/spdxdocs/" prefix and using a UUID at the end


goneall (approved):
LGTM

BTW if / when we support the namespacing of SPDX ID's we may be able to make these a bit more readable by having the GUID in a common prefix.

zvr:
In JSON level they are just strings.

Instead of 
```json
            "spdxId": "https://spdx.org/spdxdocs/File-8f79956e-4089-4166-9a71-457de77e4846",
```
you can already easily write
```json
            "spdxId": "mydata:File-8f79956e-4089-4166-9a71-457de77e4846",
```
by having a `@context` defining your `mydata` prefix.

Or am I missing something?

zvr:
Re: the actual PR, I'd actually prefer if the IDs were _not_ on `spdx.org` but on another domain. We don't want to give the impression that `spdx.org/spdxdocs` is something special and should always be used...

JPEWdev:
On Thu, May 23, 2024 at 5:09 AM Alexios Zavras (zvr)
***@***.***> wrote:
>
> Re: the actual PR, I'd actually prefer if the IDs were not on spdx.org but on another domain. We don't want to give the impression that spdx.org/spdxdocs is something special and should always be used...

I used that because it is explicitly called out in SPDX 2.3 that you
can use that prefix for a document namespace when don't own a domain
(https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#652-intent)

If we don't use this same prefix, SPDX 3.0 should recommend a similar
one to use in the same circumstances.

zvr (approved):

PR #967, opened 2024-05-28
Fix typo in TOC entry for Lite profile

stefan6419846:


goneall (approved):
Good catch - thanks @stefan6419846

goneall (approved):
Good catch @stefan6419846

PR #968, opened 2024-06-01
New annex on license matching

zvr:
This copies the license matching guidelines as an annex to the spec.


goneall (approved):
LGTM - One thing to consider is the references to the LicenseXML format - we may want to either remove or bring the license XML spec into this repo.

kestewart (approved):

PR #946, opened 2024-04-27
README.md: Update build instructions and add build flow diagram

bact:
Please use #971 instead

PR #971, opened 2024-06-10
Add build instruction and flow diagram

bact:
The intention of this PR is to allow more contributors to test the spec and spot issues early.
It will also allow more people to understand the build process which involved components across several spdx repos.

- Add info on repo structure (which directory is for what purpose)
- Update build instructions
  - Add a build flow diagram and instruction on using spdx-3-model, spdx-spec, and spec-parser together
  - Add info about MkDocs configuration
  - Add info about mike and GitHub workflow
- Update info about development branch and repo
  - Was pointing to https://github.com/spdx/spdx-spec/tree/master/chapters which is no longer exist

This is a rework of #946

Note: Once we are set on internationalization/localization, this will need to be updated again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,17 +190,18 @@ Files intended for display and linking in the navigation bar should be
 included in the `nav:` section. The order of filenames in this section
 determines their order on the navigation bar.
 
-## Specifications on spdx.github.io/spdx-spec/
+## Specification versions on spdx.github.io/spdx-spec/
 
 The SPDX specifications on <https://spdx.github.io/spdx-spec/> are built
 by using a workflow in
 [`.github/workflows/publish_v3.yml`](/.github/workflows/publish_v3.yml).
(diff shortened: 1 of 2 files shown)


bact:
@JPEWdev if you have time to review this again. (I have removed the `ontology/` directory from the repo structure already).

JPEWdev (approved):

zvr:
I don't think build and website deployment instructions belong in the repo top-level README file, but maybe it's just me...

goneall (approved):
LGTM

goneall:
> I don't think build and website deployment instructions belong in the repo top-level README file, but maybe it's just me...

We discussed on the tech and agreed that there is enough improvements here that we should go ahead and merge.  We can decide later to split this out into a separate README and create a separate follow-on PR.

PR #972, opened 2024-06-11
Remove RC versions from version selector

### LINKED ISSUE #952: Version selector dropdown: Removing RC versions of the versions with stable release?
Currently the spec website at https://spdx.github.io/spdx-spec/ has a spec version selector dropdown,
which contains these versions:

https://github.com/spdx/spdx-spec/blob/21d9152828de976d19cef5e4b7dd564da2d92f97/versions.json#L1

Should we consider removing RC versions of the versions with stable release, namely `v3.0-RC2` and `v2.3-RC1`, from the list?

bact:
- Remove `v3.0-RC2` and `v2.3-RC1` from version selector  on the website
- `v3-draft` alias still point to `v3.0`, no change

This will close #952

Note that this will only remove RC versions from the version selector but the actual directories that contain HTML files of these drafts/RCs are still kept intact. So if people access directly to these URLs they can still see the content of that draft version:

- https://spdx.github.io/spdx-spec/v2-draft/ (has a content that shows "v2.2.2". This is a non-mike configuration, notice that it doesn't have the version selector)
- https://spdx.github.io/spdx-spec/v2.3-RC1/
- https://spdx.github.io/spdx-spec/v3.0-RC2/

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -1,1 +1,20 @@
-[{"version": "v3.0", "title": "v3.0", "aliases": ["v3-draft"]}, {"version": "v3.0-RC2", "title": "v3.0-RC2", "aliases": []}, {"version": "v2.3", "title": "v2.3", "aliases": ["latest"]}, {"version": "v2.3-RC1", "title": "v2.3-RC1", "aliases": []}, {"version": "v2.2.2", "title": "v2.2.2", "aliases": []}]
+[
+    {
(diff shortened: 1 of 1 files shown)


goneall (approved):
Thanks @bact - we can make follow-on PR's to clean up the aliases

zvr (approved):

PR #974, opened 2024-06-13
Rename ISO_foreword.md -> ISO-foreword.md

bact:
- Rename ISO_foreword.md to ISO-foreword.md, using `-` instead of `_` to make it consistent with other filenames in `docs/`
- Break a long line into shorter lines
- No change in content
- Note that this file is not included in v3.0 but will later included after the ISO process


goneall (approved):
LGTM - Thanks @bact

kestewart (approved):

bact:
I will close this PR as #1001 for ISO documentation already covered this.

PR #983, opened 2024-07-08
[2.3] Use H2 for headings in Annex I (diff from prev editions) to fix nav bar issue

bact:
To allow readthedocs to generate proper heading items in the navigation bar.

Currently when a user clicked on the page https://spdx.github.io/spdx-spec/v2.3/diffs-from-previous-editions/ ,
the nav bar on the left will not expanding the subheadings. This is behavior is different from other chapters/annexes.

The reason is, in Markdown, H1 is reserved for the title.
mkdocs's readthedocs theme will read only H2 headings to generate expandable items in the nav bar.

This fix add one level to all headings in the doc (H1 -> H2), except the title.

There is no change in the content.

See screenshot of a similar fix (for 3.0) in #988

### DIFF
--- a/chapters/diffs-from-previous-editions.md
+++ b/chapters/diffs-from-previous-editions.md
@@ -1,6 +1,6 @@
 # Annex I Differences from previous editions (Informative)
 
-# I.1 Differences between V2.3 and V2.2.2  <a name="I.1"></a>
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

zvr (approved):

PR #985, opened 2024-07-09
Update license expression syntax references to license list

lpanni:
In SPDX v3 `license-id` and `license-exception-id` can no longer be defined by referencing Annex A (which previously included the license list).
Instead reference the SPDX License list and License Exceptions list directly.


lpanni:
@zvr your commits are missing a sign-off

zvr:
@lpanni yeah, because I clicked on this GitHub interface during the call, committing my suggestion.

PR #994, opened 2024-07-18
Republish 3.0 as 3.0.0

bact:
### What this PR do?

Allow users to get access to the original v3.0 (v3.0.0) content, after the v3.0.1 overtake the "v3.0" as an alias.

- This PR will republish `v3.0` as `v3.0.0`
  - Using `development/v3.0` of `spdx-spec`, and `3.0` tag of `spdx-model`
- `v3.0` will then be available to be use as an alias to the latest v3.0.x version.
- See the last comment block in [publish_v3.yml](https://github.com/spdx/spdx-spec/blob/9b2a262a19d4a79be326f8015f0a569ed5eec8f5/.github/workflows/publish_v3.yml) in #1015

### Changes

- in .github/workflows/publish_v3.yml:
  - Update `actions/checkout` to v4 to prevent build warning of Node.js deprecation
  - Use `3.0` tag release of `spdx-model`
  - Deploy using `v3.0.0` version

- in mkdocs.yml:
  - Set metadata to use v3.0.0 as a version, including `canonical_version: v3.0.0`

### Notes

- We may not need this PR, if we don't like to keep many versions of v3.0.x
- If we like to keep only the latest one of v3.0.x (for example, v3.0.1), we can just apply #1015 and close this PR. (1015 will delete v3.0, and make v3.0 an alias of v3.0.1)


bobmartin3000 (approved):
Looks good to me.

bact:
Add validation examples changes from commit [`389566e`](https://github.com/spdx/spdx-spec/commit/389566e9553ae71a96cd5f6ac1d77a69a47af4fa).

All checks have passed.

bact:
@bobmartin3000 Thank you for the review. Please be notified that this PR has changed its target a bit - to anticipated with the coming release of 3.0.1 and changes in #1015. Feel free to withdraw your approval.

bobmartin3000 (changes requested):
I agree to withdraw my approval but can't figure out how to do that.

bact:
> I agree to withdraw my approval but can't figure out how to do that.

Your approval has already been withdrawn when I re-request the review.

kestewart:
Not essential for 3.0.1 pdf spec publishing.

bact:
Close. Will open a new PR for this once branch renaming is done and a new publishing workflow is updated to support the publication from the "support" branch.

bact:
Will rework on this but publish it from the new `support/3.0` branch

PR #997, opened 2024-07-23
Add <del> (~~strikethrough~~) support

### LINKED ISSUE #929: Wrong display for strikethrough text
https://spdx.github.io/spdx-spec/v3.0/annexes/diffs-from-previous-editions/

File diffs-from-previous-editions.md

In the relationship table, there should be strikethrough text for removed relationships.

But it is displayed as:
```
~~DEPENDENCY_OF~~
```
instead of
~~DEPENDENCY_OF~~

bact:
Python Markdown that mkdocs use doesn't support the rendering of `~~text~~` to ~~text~~ (strikethrough). Need to call an extension to support that.

- Adding [pymdownx.tilde](https://facelessuser.github.io/pymdown-extensions/extensions/tilde/) extension to mkdocs.yml config
- Adding [pymdown-extensions](https://pypi.org/project/pymdown-extensions/) module to requirements.txt dependencies

This will fix #929

Before fix:
Screenshot 2024-07-23 at 00 54 34

After fix:
Screenshot 2024-07-23 at 00 53 36

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -12,7 +12,8 @@ extra_css:
 - css/style.css
 markdown_extensions:
 - codehilite:
-   guess_lang: false
+    guess_lang: false
+- pymdownx.tilde
 nav:
 - 'Copyright': index.md
 - 'Introduction': introduction.md
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ mkdocs-pdf-export-plugin==0.5.10
 mike==1.1.2
 shacl2code==0.0.10
 PyYAML>=6.0.1
+pymdown-extensions==10.8.1


zvr:
I don't think we should have strikeout support. I'm not sure it's allowed in standards.

goneall:
> I don't think we should have strikeout support. I'm not sure it's allowed in standards.

Is there some other annotation we can use for the changes where a relationship is removed?  I know it is in the text, but it is much more readable with the strikethrough.

bact:
I skimmed through few OMG documents and see that they use different fonts, different sizes, character colors, background colors, underline, bold, and italic.

But that's not definitive. Wonder if there's a style guide or a list of allowed typographic devices for OMG/ISO standards?

bact:
For the diff between versions, this spec use just "blank" cells for deletion.
https://www.omg.org/spec/MOF/2.5.1/PDF
Screenshot 2024-07-24 at 10 08 03

goneall:
Discussed on 30 July 2024 tech call, we'll open a new PR to remove the strikethroughs.  Agreed we didn't need the additional formatting.

Closing this PR.

bact:
Remove the strikethroughs: https://github.com/spdx/spdx-spec/pull/1017

PR #992, opened 2024-07-16
Correct Annex list

zvr:
A number of informative documents were copied from being annexes of the spec to their own repository https://github.com/spdx/using.

This PR removes them from the spec repo, and renumbers the rest of the annexes.
It also updates the mkdocs configuration to use this new list.

Signed-off-by: Alexios Zavras <github@zvr.gr>

PR #1008, opened 2024-07-25
Add Annex F: History, Motivation and Rational

bact:
- Using text and 4 figures from Bob's Word doc 2024-07-04

### Notes
- SPDX is stylized as "Software Package Data e*X*change" and "System Package Data e*X*change" in this annex (same as in the new Introduction, as the last two paragraphs of this annex are largely similar to the Intro).
  - While in other instances of the spec, they are written as "Software Package Data Exchange" and "System Package Data Exchange".
  - Do we like to change both of them to a regular spelling?
- Links. These two links at the end of the doc may need an update for 3.0.1:
  > The SPDX 3.0 ontology is available at:
  > https://github.com/spdx/spdx-spec/tree/development/v3.0/ontology
  >
  > The SPDX 3.0 specification is available as web pages at:
  > https://spdx.github.io/spdx-spec/v3.0/


bobmartin3000 (approved):
LGTM

bact:
Note the links to model, ontology, and spec at Lines 136-143. These may need to recheck before publication.

docs/histroy.md
```diff
@@ -0,0 +1,143 @@
...
+were also working within the Linux Foundation’s Software Package Data Exchange
+(SPDX®) open-source effort to evolve their previous work. Started in 2010 to
+help organizations developing software that planned to incorporate open source
+software make sure that the licenses for that open source software were
+appropriate for how the organizations planned to use them in their own
+offerings, the SPDX community developed a series of software products,
+specifications, and capabilities to address this area. The first published work
+was a version 1.0 specification in August of 2011; followed by 1.1 version a
+year later; a 1.2 version in October 2013; a 2.0 version in 2015; and 2.1
+version in 2016. The 2.1 version of SPDX was published through the Linux
```
kestewart, line 77:
```suggestion
version in 2016.   The 2.2 version of the specification was published in 2020 to address the required SBOM minimum elements.   The 2.2.1 version of SPDX specification was published through the Linux
```

goneall (changes requested):
Since this is the history from an ISO perspective, we should start with SPDX history going back to 1.0 then, more importantly, through the previous ISO submission then adding this additional information.

I wonder if this chapter is even needed for the ISO submission since our prior submission did not include this level of detail.  If it isn't required, I would recommend removing it.

goneall:
> Since this is the history from an ISO perspective, we should start with SPDX history going back to 1.0 then, more importantly, through the previous ISO submission then adding this additional information.
> 
> I wonder if this chapter is even needed for the ISO submission since our prior submission did not include this level of detail. If it isn't required, I would recommend removing it.

In reading further, it looks like the SPDX history is there, but much later in the document.  Perhaps reordering the history to start more at the beginning may fix the problem.

kestewart:
Met with @bact, @goneall and decided this needs to be redone/renumbered, and depends on #992 being applied first.   

Discussion is this isn't essential for the .pdf, and can be handled in a cover letter to OMG with the .pdf;  and if we want to include it for ISO submission; probably needs to be reworked as described above.

kestewart (changes requested):
No longer Annex F (if we end up including).  Needs discussion.

goneall:
Also note that the annex numbering needs to change and the mkdocs.yaml file needs to be updated.

kestewart:
I'm fine with this going to OMG.   This needs the Annex F references updated.

docs/history.md
```diff
@@ -0,0 +1,143 @@
+# Annex F: History, Motivation and Rational (Informative)
```
kestewart, line 1:
```suggestion
#History with OMG, Motivation and Rational (Informative)
```

kestewart (approved):
Approve this as long as added to mkdocs

goneall (approved):
LGTM

bact:
Please use https://github.com/spdx/spdx-spec/pull/1053 instead

PR #1026, opened 2024-08-06
Add Legacy Text Template format section back to the Matching Guidelines

bact:
standardLicenseTemplate and standardAdditionTemplate still need the Legacy Text Template.

See discussion in https://github.com/spdx/spdx-3-model/issues/747#issuecomment-2263453540

- Use source from <s>https://github.com/spdx/spdx-spec/blob/development/v2.3.1/chapters/license-matching-guidelines-and-templates.md</s> https://github.com/spdx/license-list-XML/blob/main/DOCS/license-matching-guidelines-and-templates.md 
- Adjusted anchor numbering, links
- Add missing anchors: C.6.4
- Add internal link to guidelines
- Use \`...\` around some special characters to differentiate

### DIFF
--- a/docs/annexes/license-matching-guidelines-and-templates.md
+++ b/docs/annexes/license-matching-guidelines-and-templates.md
@@ -1,4 +1,4 @@
-# Annex C SPDX License List Matching Guidelines and Templates (Informative)
+# Annex C License List Matching Guidelines and Templates (Informative)
 
 ## C.1 SPDX license list matching guidelines <a name="C.1"></a>
 
@@ -249,4 +249,4 @@ The [POSIX ERE](http://pubs.opengroup.org/onlinepubs/9699919799/) in the match f
 - Semicolons are escaped with `\;`
 - POSIX Bracket Extensions are not allowed
 
-EXAMPLE: `<<var;name=organizationClause3;original=the copyright holder;match=.+>>`
+For example: `<<var;name=organizationClause3;original=the copyright holder;match=.+>>`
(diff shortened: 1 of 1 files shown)


maxhbr (approved):
LGTM, easy to review in the rich diff

kestewart:
@jlovejoy, @swinslow - here's the PR that is adding it back, that was referenced yesterday.

zvr (changes requested):
The source where we copy the whole annex should be https://github.com/spdx/license-list-XML/blob/main/DOCS/license-matching-guidelines-and-templates.md and not some ancient version from SPDXv2.

bact:
@zvr Thank you for the pointer to upstream one at license-list-XML repo. I was using v2.3.1 only because it's the latest version before v3.0 (which doesn't has this file). 

I just have checked the Legacy Text Template format section from both sources (license-list-XML and v2.3.1), they are identical.

zvr:
@bact have you checked that the complete file is identical (not only the last section) ?

bact:
@zvr this PR is based on the existing file. The last section is added. Other sections are added by PR #968.

I haven't look very carefully all of them, only headings (that's how I spotted the missing "C.6.4"). I can check them all again.

bact:
Done. Every sections in the source of license-list-XML and v2.3.1 are identical.

Through the scan, found two spots in "HTTP Protocol" section that may need changes:

- Protocol names (HTTP and HTTPS) should be in UPPERCASE
- The URL schema names at the first part of URL ("http://" and "https://", which will be the actual texts to be compared) should be in lowercase
- Open a PR https://github.com/spdx/license-list-XML/pull/2537 there in license-list-XML repo for people there to decide. Text in this PR is unchanged.

Also one typo found (introduced be myself during the link format conversion), will fix that.

jlovejoy:
@zvr - now I'm confused... are we keeping the matching guidelines in the license list repo? we definitely should not have them in two places!

kestewart (approved):

goneall (approved):
LGTM

PR #1027, opened 2024-08-07
Validate examples only when examples/**/*.json and docs/annexes/*.md changes

bact:
To safe time and resource, only run a workflow in `validate_examples.yml` when there are changes in:

- `.json` file in any directory inside `examples/`
- `.md` file inside `docs/annexes/`

The file and dir names are taken from bin/check-examples.sh


goneall (approved):
LGTM

kestewart (approved):

PR #989, opened 2024-07-10
Fix diagram filenames, make them display properly

bact:
Remove spaces from filenames to make model images displayed properly.

PR #1033, opened 2024-08-09
Fix diagram filenames, make them display properly (rework)

### LINKED ISSUE #959: The diagrams from the model aren't showing up for the 3.0 spec
From the 2024-05-07 tech call (ref @kestewart )

We have useful diagrams for some of the models; can they show up in the generated spec somehow?

bact:
- Currently diagram filenames have a space in the name. Have to replace a space with `%20` when make a link.
- If possible, it may be better to rename the image files. But for now, use them as they are.
- Also sort the diagrams by name
- This will fix #959

This is a rework of #989 (with approvals there)


bact:
@goneall sorry to bother again. Please use this one instead. I can't rebase #989

Issue #987, opened 2024-07-10
Suggested naming convention for SPDX v3 files


bact:
SPDX v2 has a suggested naming convention:

Format | Extension |
-|-|
tag:value | *.spdx |
RDF | *.spdx.rdf |
JSON | *.spdx.json |
XML | *.spdx.xml |
YAML | *.spdx.yaml or *.spdx.yml

According to https://spdx.github.io/spdx-spec/v2.3/conformance/#44-standard-data-format-requirements it "should be easy to recognize in a file system without opening the file."

Will the naming convention be the same for SPDX v3?

Or do we like to have something like `*.spdx3.*` to differentiate the version?

bact:
From Outreach meeting 2024-07-22, OMG needs that a file should have only one extension (i.e. `*.spdx` or `*.json`, but not `*.spdx.json`)

bact:
Suggested name will be in 5.12 Standard data format requirements subsection of Conformance chapter
#1002

kestewart:
This information has been removed from Conformance chapter - so keeping this open until we handle serialization and documentation.

goneall:
Suggestion is to add to the https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/serializations.md page.

We can move this to 3.1 to allow for more time to discuss.

zvr:
I also vote to leave it unspecified for now, and at some point in the future (3.1?) come up with all the corresponding entries (filenames, media types, etc.)

bact:
- Related: https://github.com/spdx/spdx-spec/issues/577

bact:
FYI, OpenSSF has published this document

Best Practices for Naming and Directory Conventions for SBOMs (Software Bill of Materials) in Open Source Projects 
https://sbom-catalog.openssf.org/sbom-naming.html

Should we agree with the OpenSSF recommendation?

vargenau:
> FYI, OpenSSF has published this document
> 
> Best Practices for Naming and Directory Conventions for SBOMs (Software Bill of Materials) in Open Source Projects https://sbom-catalog.openssf.org/sbom-naming.html
> 
> Should we agree with the OpenSSF recommendation?

The OpenSSF recommendation cites as source the SPDX 2 standard with the various formats (tag:value, JSON, etc.)
So, I do not think this recommendation applies to SPDX 3.

I really think we need a different suffix to be able to automatically distinguish SPDX 3 from SPDX 2 SBOMs.

[...]

PR #1047, opened 2024-08-11
Fixes in Conformance chapter

### LINKED ISSUE #1045: Update Conformance chapter
Some things to fix in the [Conformance](https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/conformance.md) chapter.

- remove (or at least check) the numbering of profiles chapters in second bullet list
- in the same section, fix that only Core is mandatory
- add the section on trademark conformance, from SPDXv2 standard
- In Lite profile, mention that Software also needs to be compliant
- delete the last section, since this is about serializations -- and we definitely do not want to mention Tag-Value or Spreadsheets as "current supported formats"

plus a couple of other minor points.

zvr:
- adds the section on trademark conformance, from SPDXv2 standard
- remove the numbering of profiles chapters in second bullet list
- in the same section, fixes that only Core is mandatory
- delete the last section, since this is about serializations -- and we definitely do not want to mention Tag-Value or Spreadsheets as "current supported formats"

Fixes #1045


zvr:
Depending on the order of merging (the number removal and the link to json), this might have conflicts, but they should be obvious.

docs/conformance.md
```diff
@@ -200,100 +207,24 @@ of the SPDX.
...
-Conformance to the Extension Profile compliance point does not entail support
-for the Licencing, Security, Dataset, AI, Build, or profiles of the SPDX but
-is expected to be used in combination with the other profiles to extend them.
-
-This compliance point facilitates interchange of extended information that goes
-beyond the standard SPDX produced by tools supporting SPDX and is used between
-cooperating parties that understand the form of the extension and can produce
-and consume its non-standard content.
-
-## 5.12 Standard data format requirements <a name="5.12"></a>
```
kestewart, line 246: I think we should retain this section.   Why did you choose to delete it?

docs/conformance.md
```diff
@@ -200,100 +207,24 @@ of the SPDX.
 This compliance point facilitates interchange of minimal licencing information
 when produced by tools supporting SPDX.
 
-## 5.11 Extension Profile compliance point <a name="5.11"></a>
```
kestewart, line 203: When was it determined to remove this section?

kestewart (changes requested):
All of the additions make sense and am good with, however not sure why Lite & Extension Profile Conformance point information was removed?

docs/conformance.md
zvr: Ooops, that was not the intent.

bact: In this section, we still have to figure out the standard naming convention of the SPDX files (Lines 286-295) -- if multiple extensions (`.spdx.json`) are ok for OMG. 

See issue: https://github.com/spdx/spdx-spec/issues/987 We can do it in a seperate PR.

goneall (approved):
LGTM

kestewart (approved):
Per discussions, Standard Data Format will be added back in future version, after we sort through the serialization chapter.

bobmartin3000 (approved):
LGTM

bact (commented):
LGTM - the information about data format is covered enough for now in Serialization chapter

bact (approved):

PR #1054, opened 2024-08-11
Add elements array to the example for SpdxPackage and SBOM

### LINKED ISSUE #1013: Should the SPDX V3 sbom contain a list of elements?
In the [example](https://github.com/spdx/spdx-spec/blob/development/v3.0.1/examples/jsonld/package_sbom.json), the `software_Sbom` contains a `rootElement` but does not include the list `element`.

Should the sbom include the root + all referenced elements (both direct and indirect)?  Just the root?

goneall:
Per the discussion on the tech call, there was consensus that elements should have a minimum of 1 for both SpdxPackage and SBOM.

Fixes #1013


goneall:
@JPEWdev - Please review

bact (approved):

NorioKobota:
If this is the correct specification, wouldn't it also be necessary to change ```External properties restrictions``` of Sbom class?
https://github.com/spdx/spdx-3-model/blob/main/model/Software/Classes/Sbom.md

```
## External properties restrictions

- /Core/ElementCollection/element
  - minCount: 1
- /Core/ElementCollection/rootElement
  - minCount: 1
```

Also, I think we need to discuss whether Bom and SpdxDocument classes should have at least one element, rootElement. In the end, the debate will be whether element, rootElement in [ElementCollection class](https://github.com/NorioKobota/spdx-3-model/blob/main/model/Core/Classes/ElementCollection.md) should be minCount == 1.

goneall:
> If this is the correct specification, wouldn't it also be necessary to change `External properties restrictions` of Sbom class? https://github.com/spdx/spdx-3-model/blob/main/model/Software/Classes/Sbom.md
> 
> ```
> ## External properties restrictions
> 
> - /Core/ElementCollection/element
>   - minCount: 1
> - /Core/ElementCollection/rootElement
>   - minCount: 1
> ```
> 

Thanks for providing the syntax for adding this - I created an issue in the model repo to add the restrictions, but I didn't have the information on how to modify the markdown files.  See https://github.com/spdx/spdx-3-model/issues/841

> Also, I think we need to discuss whether Bom and SpdxDocument classes should have at least one element, rootElement. In the end, the discussion will be whether element, rootElement in [ElementCollection class](https://github.com/NorioKobota/spdx-3-model/blob/main/model/Core/Classes/ElementCollection.md) should be minCount == 1.

For `ElementCollection` we discussed on a tech call and agreed that an empty collection was acceptable, so having a minCount == 0 was OK.  However, in a more recent tech call we reached a consensus that Bom and SpdxDocument (both subclasses of ElementCollection) should have a minimum of 1.

zvr:
Please see my [comment](https://github.com/spdx/spdx-3-model/issues/841#issuecomment-2283413538).
What do we decide for `rootElement` ? 

This was triggered by SPDX Lite by @NorioKobota , which I am now updating to take into effect this new change.

zvr (approved):

kestewart (approved):
This aligns with decision in the model.

bact (approved):

PR #1063, opened 2024-08-13
Add RDF IRIs and Serializing SpdxDocument to Serialization section

bact:
This PR add info about SPDX namespace, RDF resources, and JSON-LD graph objects to "Model and serialization" chapter, using content from
- https://github.com/spdx/spdx-3-model/blob/main/serialization/rdf.md
- https://github.com/spdx/spdx-3-model/blob/main/serialization/jsonld.md

Do we need this serialization info in the spec, apart from the JSON-LD context info (which is already merged) ?

Feel free to point out the sections that are not necessary to be in the spec (or too late), so I can remove them from the change set. Section reordering is possible, please suggest.


zvr:
Everything has its own IRIS in the model pages. I don't think we need to spell it out how these are constructed.

We did that in the model so that people who create the model know what will be output. But we don't expect anyone else reading the spec to create anything new.

bact:
> Everything has its own IRI in the model pages. I don't think we need to spell it out how these are constructed.
> 
> We did that in the model so that people who create the model know what will be output. But we don't expect anyone else reading the spec to create anything new.

Thank you. I will remove the IRI structure part from this PR then.

goneall:
From tech team call on 27 Aug 2024: agreed that having this information in the model repo is sufficient - concern that we would have to update both repos on every release - closing issue

PR #1067, opened 2024-08-13
Add the generated RDF files

zvr:
We used to have a top-level directory `ontology` in this repo, containing the RDF data. It was removed because it was stale.

This PR adds this data back, in a new top-level directory named `rdf`.

This includes the ontology, but also the JSON-LD context file.


bobmartin3000 (approved):
LGTM

bact:
Does it possible to take this opportunity that the URL is going to change anyway (because of the patch version update), to have this file `rdf/spdx-model.json-ld` renamed to `rdf/spdx-model.jsonld` please?

The reason is, with that we can have the extension that is consistent with `rdf/spdx-context.jsonld` and authors will not have to remember when to use `.json-ld` or when to use `.jsonld`. 

I have checked in both spdx-spec and spdx-3-model repos,
- there are at least 25 instances of `spdx-context.jsonld`
- there's only one instance of `spdx-model.json-ld`, which is in the README section, and not the spec or the code:
  https://github.com/spdx/spdx-spec/blob/399bbf524e35ea0e81e8162372dacfb2ad549b5a/README.md?plain=1#L156-L162
  Instances that reference to the SPDX model use `https://spdx.org/rdf/3.0.0/spdx-model.ttl`.

This means the change to standardize `.jsonld` will not be a breaking one.

bact (changes requested):
Does it possible to take this opportunity that the URL is going to change anyway (because of the patch version update), to have this file `rdf/spdx-model.json-ld` renamed to `rdf/spdx-model.jsonld` please?

The reason is, with that we can have the extension that is consistent with `rdf/spdx-context.jsonld` and authors will not have to remember when to use `.json-ld` or when to use `.jsonld`. 

I have checked in both spdx-spec and spdx-3-model repos,
- there are at least 25 instances of `spdx-context.jsonld`
- there's only one instance of `spdx-model.json-ld`, which is in the README section, and not the spec or the code:
  https://github.com/spdx/spdx-spec/blob/399bbf524e35ea0e81e8162372dacfb2ad549b5a/README.md?plain=1#L156-L162
  Current instances that reference to the SPDX model use `spdx-model.ttl`.

This means the change to standardize `.jsonld` will not be a breaking one.

Thank you for generating and adding these files back.

zvr:
Of course the filename can easily be changed, but let's be careful here:
we do **not** want to encourage people using these files automatically from the repo.

The idea is to simply have them for people to view (most probably via the GitHub web UI). It should not be considered a distribution mechanism of generated artifacts.
(we have had this problem before)

bact:
Thanks. +1 on "It should not be considered a distribution mechanism of generated artifacts."

I think in general, some consistency/expectability will help us the developers as well, not only the "end-users".

zvr:
I've now added all the various formats that RDF data can be automatically generated.

bact (approved):

bact:
Thank you. I think we can merge this.

PR #1090, opened 2024-08-26
Add missing para of JSON-LD context section to Model and serializations chapter

bact:
Add the following paragraph to the end of "JSON-LD context file" section in "Model and serializations" chapter (docs/serializations.md) : 

> Additional namespace mappings may be defined within a separate object within the context.

Adapted from the original paragraph in https://github.com/spdx/spdx-3-model/blob/main/serialization/jsonld.md#context-file :

> Further custom namespace mapping can be included in the context in a separate object, see for example here. (link to spdx/spdx-3-model repo)

The rest of that section has already been adopted to the chapter but this line is missing.

--

Also removing HTML anchors since they are already outdated and no longer in use.


goneall:
Holding off on merging this until the tooling (shacl2code) implements the additional namespaces. 

@JPEWdev @goneall will look into this

kestewart (approved):

zvr (approved):

PR #1091, opened 2024-08-26
Fix typos in Build and Lite Profiles Conformance

bact:
> We have "Security" in the Build Profile compliance point when it should be "Build" - line 1 of para 2.
> We have "Security" in the Lite Profile compliance point when it should be "Lite" - line 1 of para 2.

Reported by @bobmartin3000


bobmartin3000 (approved):
LGTM

goneall (approved):
LGTM - Thanks @bact

PR #1095, opened 2024-08-27
Add "Profile" to Terms and defintions

### LINKED ISSUE #1088: Add terms
In the "Terms and definitions" chapter, we have to have some definitions for terms that we use in the spec.

The only one I can think of is "**Profile**".

I went through everything defined in the Glossary in the model repo but nothing else is relevant here.

What other term do we use?

bact:
Will resolve #1088

Please suggest style/format of the definition list.

--

Currently, it put terms as a H2 heading, with definition as following paragraphs.

As a result, terms will also showing in the side navigation bar.

Screenshot 2024-08-27 at 20 18 08

--

Alternatively, it is also possible in MkDocs to use a Markdown syntax to generate standard HTML tags for a definition list (\<dl\> for list, \<dt\> for term, and \<dd\> for description), which in general will work better with a screen reader:
- https://www.markdownguide.org/extended-syntax/#definition-lists
- https://python-markdown.github.io/extensions/definition_lists/

This code:
```text
Profile

:   A scope of usage for SPDX targeting support for particular use cases and scenarios (e.g., software, licensing, security, etc.).

    A profile identifies which particular SPDX namespaces, classes, and properties it leverages, along with any custom constraints unique to its use.

Test

:   Test test
```

will generate

Screenshot 2024-08-27 at 20 18 56

--

Which one we prefer?

With the compatibility with our current PDF generation in mind.

### DIFF
--- a/docs/terms-and-definitions.md
+++ b/docs/terms-and-definitions.md
@@ -7,3 +7,11 @@ at the following addresses:
   [https://www.iso.org/obp](https://www.iso.org/obp)
 - IEC Electropedia: available at
   [http://www.electropedia.org/](http://www.electropedia.org/)
+
+## Profile
+
+A scope of usage for SPDX targeting support for particular use cases and
+scenarios (e.g., software, licensing, security, etc.).
+
+A profile identifies which particular SPDX namespaces, classes, and properties
+it leverages, along with any custom constraints unique to its use.
--- a/docs/terms-and-definitions.md
+++ b/docs/terms-and-definitions.md
@@ -8,7 +8,7 @@ at the following addresses:
 - IEC Electropedia: available at
   [http://www.electropedia.org/](http://www.electropedia.org/)
 
-## Profile
+** profile **
 
 A scope of usage for SPDX targeting support for particular use cases and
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

docs/terms-and-definitions.md
```diff
@@ -7,3 +7,11 @@ at the following addresses:
   [https://www.iso.org/obp](https://www.iso.org/obp)
 - IEC Electropedia: available at
   [http://www.electropedia.org/](http://www.electropedia.org/)
+
+## Profile
```
zvr, line 11:
```suggestion
** profile **
```

docs/terms-and-definitions.md
```diff
@@ -7,3 +7,11 @@ at the following addresses:
...
+
+A scope of usage for SPDX targeting support for particular use cases and
+scenarios (e.g., software, licensing, security, etc.).
+
```
zvr, line 15: delete line; each definition has to be a single paragraph

zvr (changes requested):

zvr (approved):

PR #1096, opened 2024-08-27
Update base IRI in annotations.ttl to 3.0.1

bact:


bact:
@JPEWdev please kindly correct me if I'm wrong. Thank you.

JPEWdev (approved):

PR #1106, opened 2024-09-06
Fix bad links related to abbreviations in References chapter

### LINKED ISSUE #1103: Non-existing links in Reference chapter, need to be fixed
In https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/references.md there are links that can not be resolved.

For example,

- https://datatracker.ietf.org/doc/rfcJSON 
- https://datatracker.ietf.org/doc/rfcSHA

Has to either remove or replace them with something valid.

bact:
- Fix errors related to abbreviations in References chapter, see starting from Line 86. For example:
  > JavaScript Object Notation (https://datatracker.ietf.org/doc/rfcJSON)

  should be just

  > JavaScript Object Notation (JSON)

  The https://datatracker.ietf.org/doc/rfcJSON is also not a valid link to JSON standard doc.
- Remove unused link anchors in headings (`<a name=.. >`, etc)

Will fix #1103

### DIFF
--- a/docs/references.md
+++ b/docs/references.md
@@ -1,6 +1,6 @@
 # References
 
-## Normative References <a name="2.1"></a>
+## Normative References
 
 The following documents are referred to in the text in such a way that some or
 all of their content constitutes requirements of this document. For dated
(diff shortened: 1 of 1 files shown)


zvr (approved):

goneall (approved):
LGTM

PR #1107, opened 2024-09-06
Remove heading numbering, fix internal refs, standardized RFC links

### LINKED ISSUE #1012: What is a standard URL to reference IETF RFC?
Currently, five styles of referencing IETF RFC found across spdx-spec and spdx-3-model

1. https://tools.ietf.org/html/rfc1321 - Found in [Normative References](https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/normative-references.md). Also used by References section in Bob's Word doc 2024-07-22.
1. http://www.ietf.org/rfc/rfc3986.txt - [packageVerificationCodeExcludedFile](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Properties/packageVerificationCodeExcludedFile.md)
1. https://datatracker.ietf.org/doc/rfc1319/ - [HashAlgorithm](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/HashAlgorithm.md)
1. https://datatracker.ietf.org/doc/html/rfc7693 - [HashAlgorithm](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/HashAlgorithm.md)
1. https://www.rfc-editor.org/rfc/rfc4634 - [HashAlgorithm](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/HashAlgorithm.md)

Additionally, there are other two styles of referencing IETF RFCs:

6. https://www.rfc-editor.org/info/rfc2046 - Used by a [bibtex](https://datatracker.ietf.org/doc/rfc2046/bibtex/) produced from [IETF Datatracker](https://datatracker.ietf.org/doc/rfc2046/) (click on "Formats: bibtex" button).
7. https://doi.org/10.17487/RFC3986 - Digital object identifier (DOI) for RFC, according to [RFC 7669](https://datatracker.ietf.org/doc/html/rfc7669#section-3)

These seven URLs land on different pages, but few of them are actually redirections:

- (4)-style redirects to (1)-style
- (7)-style redirects to (6)-style

In terms of functionality, they are all pretty much the same, as all of them (except (2)) have links to different file formats and other pages of the same RFC.

--

**Which one should we stick with?**

- Having multiple ways to refer to a single resource may require the automatic production of footnotes/bibliography  to aware that all of these are the same. Or we are at risk of producing duplicated references.
- I'm thinking about (6) as it is the one preferred by IETF themselves. But the downside of (6) (and (3)) is that you cannot link directly to subsection; while (1), (4), and (5) can. For example, https://tools.ietf.org/html/rfc7693#section-4 .

bact:
Heading numbers are now generated and some existing hard-coded anchors are not updated/wrong.

- Remove unused heading anchors that based on Heading numbering (`<a name="4.1" ...>`)
- Fix document internal refs to use `#link` that are not rely on heading numbering For example:
  - `[text](#C.2.2)` -> `[text](#guideline-official-license-headers)`
- Standardized RFC links to use `https://datatracker.ietf.org/doc/rfc...` (this will fix 1/2 parts of https://github.com/spdx/spdx-spec/issues/1012 )
- Add back SLSA to References, due to https://github.com/spdx/spdx-3-model/pull/875

Will fix #1012

### DIFF
--- a/docs/annexes/spdx-license-expressions.md
+++ b/docs/annexes/spdx-license-expressions.md
@@ -90,7 +90,7 @@ For the `tag:value` format, any license expression that consists of more than on
 
 Nested parentheses can also be used to specify an order of precedence which is
 discussed in more detail in
-*[Order of precedence and parentheses](#order-of-precedence-and-parentheses)*.
+[Order of precedence and parentheses](#order-of-precedence-and-parentheses).
 
 ### Disjunctive "OR" operator
 
--- a/docs/annexes/spdx-license-expressions.md
+++ b/docs/annexes/spdx-license-expressions.md
@@ -1,6 +1,6 @@
 # SPDX license expressions (Normative)
 
-## Overview <a name="D.1"></a>
+## Overview
 
 Often a single license can be used to represent the licensing terms of a source code or binary file, but there are situations where a single license identifier is not sufficient. A common example is when software is offered under a choice of one or more licenses (e.g., GPL-2.0-only OR BSD-3-Clause). Another example is when a set of licenses is needed to represent a binary program constructed by compiling and linking two (or more) different source files each governed by different licenses (e.g., LGPL-2.1-only AND BSD-3-Clause).
 
@@ -48,7 +48,7 @@ There MUST NOT be white space between a license-id and any following `+`. This s
 
 In the `tag:value` format, a license expression MUST be on a single line, and MUST NOT include a line break in the middle of the expression.
 
-## Case sensitivity <a name="D.2"></a>
+## Case sensitivity
 
 License expression operators (`AND`, `and`, `OR`, `or`, `WITH` and `with`) should be matched in a *case-sensitive* manner, i.e., letters must be all upper case or all lower case.
 
@@ -60,7 +60,7 @@ For user defined license identifiers, only the variable part (after `LicenseRef-
 
 The same applies to `AdditionRef-` user defined identifiers.
 
-## Simple license expressions <a name="D.3"></a>
+## Simple license expressions
 
 A simple `<license-expression>` is composed one of the following:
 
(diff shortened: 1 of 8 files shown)


bact:
Validation failed because the redirection of RDFs are not up-to-date yet. See #1108

docs/annexes/spdx-license-expressions.md
```diff
@@ -80,17 +80,19 @@ DocumentRef-spdx-tool-1.2:LicenseRef-MIT-Style-2
...
+### Introduction
 
 More expressive composite license expressions can be constructed using "OR", "AND", and "WITH" operators similar to constructing mathematical expressions using arithmetic operators.
 
 For the `tag:value` format, any license expression that consists of more than one license identifier and/or LicenseRef, may optionally be encapsulated by parentheses: "( )".
 
-Nested parentheses can also be used to specify an order of precedence which is discussed in more detail in [D.4.5](#D.4.5).
+Nested parentheses can also be used to specify an order of precedence which is
+discussed in more detail in
+*[Order of precedence and parentheses](#order-of-precedence-and-parentheses)*.
```
zvr, line 93:
```suggestion
[Order of precedence and parentheses](#order-of-precedence-and-parentheses).
```

zvr (changes requested):

bact:
Suggestions merged.

goneall (approved):
LGTM

bact:
Example validation failed - Joshua acknowledged that it is a know bug in shacl2code for 3.0.1 

PR is #1117 will update the shacl2code and should resolved this.

bact:
- PR #1117 is ready. After that one got merged, we can merge this one.

bact:
I will merge this as we need the model files updated for the coming deadline. Review again and see no issues.

Issue #577, opened 2021-09-24
Expected media type for spdx json/xml documents 


samj1912:
Hello SPDX team. We are trying to integrate spdx support in buildpacks. One of the things that we had a question about was the appropriate media type for spdx documents. It looks like the IANA media type for spdx is registered as text/spdx. Looking at the entry document this seems to be specifically for the spdx tag format. What is the expected media type for json/xml spdx documents?

text/spdx+json / text/spdx+xml? Or would all of them be text/spdx? 

Wondering as we were looking at some other sbom formats on the list

CycloneDX seems to have

application/vnd.cyclonedx+json and application/vnd.cyclonedx+xml respectively.
. Swid seems to have application/swid+xml

Any guidance here would be greatly appreciated.

samj1912:
cc: @nishakm maybe you can help?

nishakm:
@rnjudge Applied on behalf of SPDX. We could also apply for something like `application/org.spdx+json`. It's something we can bring up in the next SPDX tech meeting.

rnjudge:
@samj1912 Yes, the intention of the current IANA SPDX type was intended to represent tag-value format with the intention being to re-visit adding other formats in the future so I suppose the future is upon us :) Definitely worth raising at the next meeting. Thanks for surfacing this.

rnjudge:
Hi @samj1912 - just wanted to update you that I will open an application to add SPDX JSON/XML IANA media types this week.

rnjudge:
The `application/spdx+json` media type is officially approved and recorded with IANA: https://www.iana.org/assignments/media-types/application/spdx+json.

After discussing with a few SPDX folks, it was determined that the XML schema needs further review before officially submitting to IANA. I am having conversations around this now and will try to report back with an estimated time frame for when we can expect this.

kestewart:
Rose discussed:   Only missing media type is now XML.     Need to get follow up for @zvr for review.

bact:
Media type information page on IANA also has a field about file extension.
The tag:value SPDX text file has an extension of ".spdx" in IANA.

For JSON, in SPDX 2.3 spec it suggests ".spdx.json"
https://spdx.github.io/spdx-spec/v2.3/conformance/#44-standard-data-format-requirements

For SPDX 3, there's no suggested file extension in the spec
- https://github.com/spdx/spdx-spec/issues/987

PR #1124, opened 2024-09-26
Differences between 3.0.1 and 2.2.1 for ISO submission

### LINKED ISSUE #999: Annex A (diffs from previous) should be about the previous ISO version
Annex A (as it will appear in the updated ISO standard) should contain the differences between this version (3.0.1) and the previous ISO version (2.2.1).

Content of differences between older versions (like the one currently there, in sections A.4 - A.7) is more like "release notes".

This probably would be better to be moved to spdx/using documentation repository.

@goneall @kestewart do you agree?

bact:
Draft - Based on text in https://github.com/spdx/using/blob/main/docs/diffs-from-previous-editions.md 

## Difference from the base-text

- "Properties Removed" and "Naming Differences" sections are converted to tables for easy comparison
- Fixed typos and wrong property names, as listed in https://github.com/spdx/using/pull/18
- Added new hash algorithms (from 2.3)
- Added info about the change of License matching to XML format (from 2.3) 

To resolve #999 

## To be added

- Incorporate changes between 2.2.1 and 2.3 in
- V2.3-V2.2.2:
  - <s>License matching is now using License List XML format</s>
  - <s>Added hash algorithms (SHA3-256, SHA3-384, SHA3-512, BLAKE2b-256,
  BLAKE2b-384, BLAKE2b-512, BLAKE3, ADLER32 )</s>
  - [PrimaryPackagePurpose](https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field) is new in 2.3

## Issues to be resolved

- In SPDX 2 File Type v SPDX 3 Content Type table, there's a mention of "text/spdx", it is tag:value SPDX file format which is not available in SPDX 3 yet.
  - How to deal with this?
  - We also don't have the IANA entry yet for the JSON format (see #577)
  - Easiest way is to delete it from the table, won't effect the section as it is just a given example among other examples.

## Possible style

We may like to see examples from similar "diffs" in other ISO docs.

For example, ISO 13485:2016 - Medical devices - Quality management systems - Requirements for regulatory purposes has an "Annex A Comparison of content between ISO 13485:2003 and ISO 13485:2016" in a table format.

| Clause in ISO 13485:2016 | Comment on change compared with ISO 13485:2003 |
|-|-|
| Foreword | — Clarifies ... |
| Introduction<br />0.1 General | — Includes ...<br />— Explains ...<br/>— Alerts ... |


zvr:
I think the "changes" should start with much more general view, talking about how we went from SpdxDocument to general data, leaving behind the Package/File/Snippet types but introducing dozens of different elements, etc. etc.

I cleared the milestone because I don't think it's realistic to have this ready immediately.

bact:
Close as looks like we don't want this for ISO (or nobody proposed anything/accepted anything that got proposed).

Do we still want to have a section of diffs between versions somewhere on the website just like we do have in the past? 
If so, a diff between 3.0.1 and 3.0 is needed. These can be based on change logs in spdx-spec and spdx-3-model repos.

bact:
Close as looks like we don't want this for ISO.

bact:
Do we still want to have a section of diffs between versions somewhere on the website just like we do have in the past?
If so, a diff between 3.0.1 and 3.0 is needed. These can be based on change logs in spdx-spec and spdx-3-model repos.

goneall:
> Do we still want to have a section of diffs between versions somewhere on the website just like we do have in the past? If so, a diff between 3.0.1 and 3.0 is needed. These can be based on change logs in spdx-spec and spdx-3-model repos.

We can update the [diffs-from-previous-versions](https://github.com/spdx/using/blob/main/docs/diffs-from-previous-editions.md) in the using repo

PR #1129, opened 2024-09-28
Standardize RFC links to https://datatracker.ietf.org/doc/rfc...

### LINKED ISSUE #1012: What is a standard URL to reference IETF RFC?

bact:
- In License Expression
  - Use https://datatracker.ietf.org/doc/rfcXXX for RFC links -- to resolve #1012
  - Use `-` instead of `*` for bullets (like the rest of the spec) to avoid Markdown warnings
- In Package URL spec
  - Fix casing: semver -> SemVer; nevra -> NEVRA -- according to casing found in project docs
    - https://semver.org/
    - https://docs.fedoraproject.org/en-US/modularity/core-concepts/nsvca/
  - Fix unclickable links

### DIFF
--- a/docs/annexes/pkg-url-specification.md
+++ b/docs/annexes/pkg-url-specification.md
@@ -126,7 +126,7 @@ The rules for each component are:
 - The version is prefixed by a at-sign `@` separator when not empty.
 - This at-sign `@` is not part of the version.
 - A version must be a percent-encoded string.
-- A version is a plain and opaque string. Some package types use versioning conventions such as semver for NPMs or nevra conventions for RPMS. A type may define a procedure to compare and sort versions, but there is no reliable and uniform way to do such comparison consistently.
+- A version is a plain and opaque string. Some package types use versioning conventions such as SemVer for NPMs or NEVRA conventions for RPMS. A type may define a procedure to compare and sort versions, but there is no reliable and uniform way to do such comparison consistently.
 
 ### Rules for qualifiers
 
@@ -194,7 +194,7 @@ The current list of known types is:
 The list, with definitions for each type,
 is maintained in the file named `PURL-TYPES.rst`
 in the online repository
-https://github.com/package-url/purl-spec.
+<https://github.com/package-url/purl-spec>.
 
 ## Known qualifiers key/value pairs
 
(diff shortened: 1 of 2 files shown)


zvr (approved):

goneall (approved):
LGTM

PR #1141, opened 2024-10-22
Japanese translation

koizumistr:


bact:
The addition of translated files are looking good.

The only warning, which made the check failed, is about the inclusion of those files in `nav` section of `mkdocs.yml`.
This needs more discussion later on how we like to deal with it (it is likely that we may need one MkDocs configuration per one language?)
But for now I think this PR is good to merge (and we can improve upon afterward). The translation files themselves are good.

The only question I have in mind now is which branch this PR should merge to?

bact:
This should be retarget to merge to `support/3.0` branch

bact:
Currently the workflow shows below error because it does not know how to handle translation files:

```text
WARNING -  The following pages exist in the docs directory, but are not included in the "nav" configuration:
  - ja/index.md
  - ja/scope.md
  - ja/annexes/spdx-lite.md
  - ja/front/copyright.md
  - ja/front/introduction.md
  - ja/licenses/CC-BY-3.0.md
  - ja/licenses/Community-Spec-1.0.md
 ```

These files are expected to be handled by the MkDocs internationalization plugin. (Not something that can be fixed directly in this PR alone. Need a rework in publication workflow)

goneall:
@koizumistr Could you change the target to the support/3.0.1 branch?

PR #1145, opened 2024-11-13
Remove double "the" in Intro chapter

bact:
In the Introduction chapter,

"in advance of the, the final version" -> "in advance of the final version"


goneall (approved):
Thanks @bact

PR #1148, opened 2024-11-14
Update copy command in MkDocs build instruction to prevent copying to wrong location

bact:
Adjust copy command in the instruction to make it still copy individual subdirs inside `parser_output/mkdocs` dir (but not the `mkdocs` dir itself), even if `spdx-spec/docs/model/`and `spdx-spec/docs/rdf` dirs are already exists.

### What is wrong?

If `spdx-spec/docs/model/` exists, the current command:

```shell
cp -R parser_output/mkdocs spdx-spec/docs/model 
```

will results this content inside the `spdx-spec/docs/model/` dir:

```text
mkdocs
```

which is not what we want. We want the subdirs to be copied at this level, as MkDocs expected.

### What this PR do?

This PR update the command to:

```shell
mkdir -p spdx-spec/docs/model
cp -R parser_output/mkdocs/* spdx-spec/docs/model 
```

which will give this expected result in the `spdx-spec/docs/model/` dir, no matter if the dir exists before or not:

```text
AI			Dataset			Licensing		SimpleLicensing
Build			ExpandedLicensing	Lite			Software
Core			Extension		Security
```

This is new command also to prevent the case that the user may already have model subdirs inside `spdx-spec/docs/model/` dir from a previous run of spec generation and want to try another run. When they make copy with the current copy command, it will results to the below content (note the `mkdocs` subdir at the last row at the very end):

```text
AI			Dataset			Licensing		SimpleLicensing
Build			ExpandedLicensing	Lite			Software
Core			Extension		Security		mkdocs
```

This will lead to MkDocs errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,28 +12,37 @@ the software supply chain. SPDX reduces redundant work by providing a common
 format for companies and communities to share important data about software
 licenses and copyrights, thereby streamlining and improving compliance.
 
+Current stable version:
+
+- The current stable specification is available at:
+  <https://spdx.github.io/spdx-spec/>
+
+Next version (work in progress):
+
+- A preview of the next version, currently under active development, is
+  available at:
+  <https://spdx.github.io/spdx-spec/develop/>
+  (This website is automatically updated with each commit to the `develop`
+  branch).
+
+Translations of the specification may be available.
+English remains the normative language in all cases.
+
 ## Specification development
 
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

bact:
Will merge this documentation fix with one approval.

Further updates may needed when we have a new version of spec-parser that can specify directories for different output. We can make another PR at that time.

PR #1149, opened 2024-11-14
Add branch strategy to README

goneall:
This implements the branching strategy agreed to on the [5 November 2024 tech call](https://github.com/spdx/meetings/blob/main/tech/2024/2024-11-05.md#branch-policy).

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ The SPDX spec repo follows the [Gitflow](https://www.atlassian.com/git/tutorials
 
 The branches in use are:
 - main - This will always be the latest released specification.
-- develop - This branch will be where the active development for the next major or minor version will take place.  Once relased, the develop branch will be merged into the main branch.
+- `develop` - This branch will be where the active development for the next major or minor version takes place.  Once released, the `develop` branch will be merged into the `main` branch.
 - support/x.y - These branches will be long lived and contain any updates to a minor version of the specification.  Additions such as translations can be added to the support branch.  x.y represents the major/minor version.  Once any changes are accepted and released, the support branch will be tagged and merged into both the develop and main branches.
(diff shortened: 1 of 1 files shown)


bact (changes requested):
Thank you. I have made few minor suggestions inline.

goneall:
@bact - Thanks for the suggestions - all have been accepted

bact (approved):
LGTM

README.md
```diff
@@ -60,6 +60,17 @@ This repository consists of these files and directories (partial):
 - `mkdocs.yml` - MkDocs recipe for the spec documentation generation. The
   inclusion of model files and the order of chapters are defined here.
 
+## Branch structre
+
+The SPDX spec repo follows the [Gitflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) workflow with the addition of support branches.
```
zvr, line 65: Shall we link to something less critical of Gitflow?
For example https://gist.github.com/HeratPatel/271b5d2304de2e2cd1823b9b62bf43e0 (which also incudes support branches)

zvr (changes requested):

README.md
goneall: Yes - I was looking for a better link - so thanks for providing a good alternative.   I was hoping to find something in Wikipedia - but the only references are to the URL I used above.

I'll update the PR.

goneall: PR Updated

goneall:
Since the suggestion from @zvr were accepted and we should have this published with the branches renamed, I'm going to go ahead and merge.

If there are further suggestions, we can create a separate PR.

goneall:
@bact - Thanks for catching this - the master branch has now been renamed to main

PR #1150, opened 2024-11-14
Fix indent in dependabot.yml

bact:
To fix breaking dependabot workflow.

Sorry @goneall, I made a wrong indent.


goneall (approved):
Ahh - the fun of YAML

PR #1146, opened 2024-11-13
Update publication CI for "develop" branch

bact:
- <s>Use "main" for branch name for the website publication at https://spdx.github.io/spdx-spec/.</s>
  - The update of publication CI for "main" branch will be done separately, as it has to be merged separately into "main" branch
- Use "develop" for branch name for the website publication at https://spdx.github.io/spdx-spec/v3.x.x-dev

According to Tech Team Meeting 2024-11-05:

> Main branch is for keeping released versions

Done by updating `REF_SPEC` variable in `.github/workflows/publish_v3.yml` (Line 25).

### Notes

- This work is one part for #1092
- <s>"main" branch is not exist yet.</s>
  - <s>This PR will not be merged until the branches got renamed.</s>
- `VERSION` is the version to be published when running this workflow.
- `VERSION_DEFAULT` is the default version for redirection if no version is specified in the URL.
-  Currently, our "v3.0.1" is actually "under development" but published to "stable version" URL. This should be only temporary during this transition.
  - After v3.0.1 is officially published and merged to "main", "main" should then be the real stable version and changes in "develop" branch can then be published to https://spdx.github.io/spdx-spec/v3.x.x-dev URL.

#### Publishing "support" branch

- From Tech Team Meeting 2024-11-05:
    > For example, we may be already at 3.2, but we wan to make some changes in 3.0.1.
    > We can have "support" branches (pink line) for that.
    > - "support" branches can live longer (they live until the associated release is deprecated).
- For each "support/x.y." branch (a branch that will have its version published on the website as well under `https://spdx.github.io/spdx-spec/vx.y.z/`):
  - Need to modify `REF_SPEC`, `REF_MODEL`, `VERSION` and  `VERSION_ALIASES` variables in `publish_v3.yml` of that branch.
  - `VERSION_ALIASES` in the "support" branch should not contain "latest"; it is only reserved for "main" branch
  - `VERSION_DEFAULT` should be kept the same across all branches.
    - All non-"main" branches are now prevented from setting the default version (see from Line 180).
    - We may later like to move this VERSION_DEFAULT from the workflow file to a separated text file instead and read the value from that, so we can have a single source of truth.

#### spec-parser version

- This workflow uses the `spec-parser` directly from its "main" branch (see `REF_SPEC`), with an assumption that the "main" spec-parser will work with any version of `spdx-3-model` and with a directory structure in this workflow (in `spdx-spec`).
[...]

PR #1151, opened 2024-11-15
Pause the publication workflow during the branch renaming

bact:
- A follow up from https://github.com/spdx/spdx-spec/pull/1146#issuecomment-2477602610
- This PR will pause the publication automation by commenting out events to listens (in order to trigger the workflow)
- This is a measure to make sure that the publication workflow will not automatically run during the branch renaming process
- Once branch renaming process is done, we should bring these event listening lines back.

### DIFF
--- a/.github/workflows/publish_v3.yml
+++ b/.github/workflows/publish_v3.yml
@@ -1,11 +1,11 @@
 on:
-  push:
-    branches:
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

PR #1154, opened 2024-11-19
Proposal to allow any patch version for Lite specVersion Field

goneall:
This is to resolve https://github.com/spdx/spdx-3-model/blob/main/model/Lite/Lite.md

### DIFF
--- a/docs/annexes/spdx-lite.md
+++ b/docs/annexes/spdx-lite.md
@@ -121,7 +121,7 @@ Additionally:
 - Mandatory
     1. created
     1. createdBy (may be multiple), SHOULD be objects of type /Core/Agent
-    1. specVersion, MUST be a fixed string, “3.0.*” - where * is any supported dot version of the SPDX specification.
+    1. specVersion, MUST be a fixed string, “3.0.*” - where * is any supported dot version of the SPDX specification
 - Recommended
     1. comment
 
--- a/docs/annexes/spdx-lite.md
+++ b/docs/annexes/spdx-lite.md
@@ -121,7 +121,7 @@ Additionally:
 - Mandatory
     1. created
     1. createdBy (may be multiple), SHOULD be objects of type /Core/Agent
-    1. specVersion, MUST be a fixed string, “3.0.*” - where * is any supported dot version of the SPDX specification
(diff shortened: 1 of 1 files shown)


goneall:
@ilans @yoshi-i - Please review the proposed wording changes

bact:
Note: Validation checks failed, because of the known issue of missing CreationInfo for individuals, and they are irrelevant to the changes proposed by this PR.

bact (approved):

docs/annexes/spdx-lite.md
```diff
@@ -121,7 +121,7 @@ Additionally:
 - Mandatory
     1. created
     1. createdBy (may be multiple), SHOULD be objects of type /Core/Agent
-    1. specVersion, MUST be a fixed string, “3.0.1”.
+    1. specVersion, MUST be a fixed string, “3.0.*” - where * is any supported dot version of the SPDX specification.
```
bact, line 124:
```suggestion
    1. specVersion, MUST be a fixed string, “3.0.*” - where * is any supported dot version of the SPDX specification
```
Stylistically, other list items in this page do not have fullstops. (The fullstop is not introduced by this PR, it was there before) But I will leave this as optional so it will not blocking.

NorioKobota:
LGTM

docs/annexes/spdx-lite.md
bact, line 124: I'm not sure which one is more formal between "patch release", "point release", and "dot release".

bact: I agree with the "patch version".

docs/annexes/spdx-lite.md
```diff
@@ -121,7 +121,7 @@ Additionally:
 - Mandatory
     1. created
     1. createdBy (may be multiple), SHOULD be objects of type /Core/Agent
-    1. specVersion, MUST be a fixed string, “3.0.1”.
+    1. specVersion, MUST be a fixed string, “3.0.*” - where * is any supported dot version of the SPDX specification
```
goneall, line 124:
```suggestion
    1. specVersion, MUST be a fixed string, “3.0.*” - where * is any supported patch version of the SPDX specification
```

goneall:
Looks like we have 2 approvals from comments and reviews and the CI failure is due to the known issue with individuals, I'm going to go ahead and merge.

PR #1183, opened 2025-02-15
Update "main" branch publish workflow to use source from "main" of spdx-spec & spdx-3-model

bact:
## Background

- "main" branch as its current form was not existed at the time of the latest publication workflow update.
- New spec-parser CLI
  - Position of input directory is changed and specifying the output directory requires an option `-o`
  - The current command line no longer works

## What this PR does?

- Updating the workflow in the "main" branch to use the source from spdx-spec "main" branch.
  - This goes according to the [current branch structure](https://github.com/spdx/spdx-spec/blob/develop/README.md#branch-structure)
- Updating command line for spec-parser
  - As we can specify output directories, some of the copy steps are no longer necessary and now removed
  - The same command line is also get updated in the build doc #1186

**This PR is targetting "main" branch.**

### DIFF
--- a/.github/workflows/publish_v3.yml
+++ b/.github/workflows/publish_v3.yml
@@ -1,12 +1,12 @@
 # Publish SPDX specification to https://spdx.github.io/spdx-spec/
 #
-# There will be this workflow in "main", "develop", and "support" branches.
-# Each of them publish to a different URL.
+# This workflow is configured differently for the
+# "main", "develop", and "support" branches, each publishing to a unique URL.
 #
 # For example,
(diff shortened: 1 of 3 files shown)


goneall (approved):

bact:
As this PR updates the workflow to follows the agreed branch structure here: https://github.com/spdx/spdx-spec/blob/develop/README.md#branch-structure, follows the approved+merged new command line for spec-parser from #1186, and it does not change any content, I will merge this with one approval.

bact:
The new publication workflow runs successfully.

PR #1187, opened 2025-02-25
Fix typos in SPDX license expressions annex

bact:
- Two "License Expression" -> "license expression" (matched with other instances)
- "WITH" -> `WITH` (other operators in the same paragraph use the same style: `AND` `OR` `+`)
- "backwards compatibility" -> "backward compatibility"
- "are a valid license expression values" -> "are valid license expression values" (without 'a')
- "-ButIdDontWantToException" -> "-ButIDontWantToException" (the name of the exception is `"But I Don't Want To" Exception`)


goneall (approved):
LGTM

bact:
It has been a month already. As we already have one approval, I will merge these obvious minor typo fixes to `dev` branch.

PR #1191, opened 2025-04-10
[WIP] Multilingual support prototype

bact:
> ***This PR is not meant for merge. Just only for convenient code comparison.***
>
> - Please focus on `mkdocs.yml` and overall location of content files.
>  - This is not work yet with the GitHub CI but can be used to build the site using instructions in `build.md`
> - The content changes and changes in directory locations should be in another separate PRs.

This is what we can have now for internationalisation (i18n) support of the spec website, using `mkdocs-static-i18n` plugin with `MkDocs`, see the demo site:

- https://bact.github.io/spdx-spec/ja/scope/
- https://bact.github.io/spdx-spec/ja/model/Core/Classes/Agent/ 
- https://bact.github.io/spdx-spec/ja/model/Core/Vocabularies/AnnotationType/

For content without translation, it will fallback to English one.

Note that there is no language selector yet. To switches between languages, one needs to add the language code to the URL manually:

- https://bact.github.io/spdx-spec/ (English, default)
- https://bact.github.io/spdx-spec/ja/ (Japanese)

(Japanese is chosen for testing purpose because we have some test content in both spdx-spec and spdx-3-model sides)

The demo content is from spdx-spec and spdx-3-model repos combined, based on these PRs/issues:

- Spec pages: https://github.com/spdx/spdx-spec/pull/1141
- Model: https://github.com/spdx/spdx-3-model/pull/953
- Section names: https://github.com/spdx/spdx-3-model/issues/954

## Configuration

The configuration in `mkdocs.yml` uses `folder` structure option of `mkdocs-static-i18n` (see `i18n` section in the config file).

See the documentation here: https://ultrabug.github.io/mkdocs-static-i18n/setup/

## Unresolved issues

As I still have issues having `mike` working with the i18n plugin, the multiple version builds (3.0.1, 3.1-dev, etc) is not ready.

When mike is trying to copy "version-select.css" file from a translated site, it will find that the file is already existed (from the English website), and it will stop with this error message:

```text
ERROR   -  'css/version-select.css' is already included in 'extra_css'
```

Update: a small hack made possible to build mike with i18n now, but I'm not sure if it's a right way to do it. An issue is filed here: https://github.com/jimporter/mike/issues/240


bact:
Alternatively, Materials theme for MkDocs supports multilingual site out-of-the-box. We can use that as well.

See: https://squidfunk.github.io/mkdocs-material/setup/changing-the-language/

PR #1192, opened 2025-04-15
Fix typos and inconsistencies in license expression annex

bact:
- Fix typos
  - "of a `<license-expression>` **a** is:" -> "of a `<license-expression>` is:" (remove extra 'a')
  - "either LGPL-2.1-only **and** the expression" -> "either LGPL-2.1-only **or** the expression" (change 'and' to 'or')
- Syntax highlight ABNF and XML code blocks
- `[idString]` -> `(idstring`); use the same style and casing as in the ABNF syntax
- In description of "Simple license expressions", uses the same language and simplified pattern from user defined addition ref for user defined license ref:
  - Remove `1*` from the pattern, `1*` is also not in the full syntax at the beginning of the annex 
- Adjust indent of user defined license ref examples to be under the same bullet level as the user defined license ref itself (to avoid confusion if the examples are for the license ref or for the entire simple license expression):

  Before:
  Screenshot 2568-04-15 at 01 11 38

  After:
  Screenshot 2568-04-15 at 01 07 57


goneall (approved):
LGTM

PR #1195, opened 2025-04-15
Use JSON context URL from github.io during development

bact:
Since JSON context may not be available yet at https://spdx.org/ at the time of model development, this PR will tell JSON Schema generator to use the JSON context URL at https://spdx.github.io/spdx-spec/ instead.

For released versions, the workflow will use the https://spdx.org/ one (as it current is).

For example, the context-url passed to shacl2code for version:

- 3.0.1 will be https://spdx.org/rdf/3.0.1/spdx-context.jsonld
- 3.1-dev will be https://spdx.github.io/spdx-spec/v3.1-dev/rdf/spdx-context.jsonld
- 3.1 will be https://spdx.org/rdf/3.1/spdx-context.jsonld

The selection is based on version naming convention (suffix "-dev", "-rc1", etc.)

----

Note: Apart from this PR, also need to update version number in [serialization/jsonld/annotations.ttl](https://github.com/spdx/spdx-spec/blob/develop/serialization/jsonld/annotations.ttl)

### DIFF
--- a/.github/workflows/publish_v3.yml
+++ b/.github/workflows/publish_v3.yml
@@ -173,14 +173,24 @@ jobs:
         # from https://spdx.org/schema/<version>/spdx-json-schema.json
         # to https://spdx.github.io/spdx-spec/v<version>/rdf/schema.json
         #
-        # Note: When release a new version, update URL in --context-url line to match the version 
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

PR #1196, opened 2025-04-15
Adds trademark symbol on "System Package Data Exchange"

bact:
Adds trademark symbol on instances of the phrase "System Package Data Exchange".

(similar to #1193 but for `support/3.0` branch)


goneall (approved):
LGTM

PR #1197, opened 2025-04-15
SPDX 3 JSON format is a strict subset of JSON-LD

### LINKED ISSUE #1008: Use "SPDX 3 JSON" instead of "SPDX 3 JSON-LD"
After a long discussion with @JPEWdev and @goneall on the implementers call, I understand that the "Model and serializations" document is describing a proprietary JSON format which was defined by our community as a compromise between pure JSON and JSON-LD.

This JSON may be deserialized into RDF using JSON-LD deserializers but needs to be serialized using proprietary tools that follow the specifications in this document.

The confusion arises from using the term JSON-LD in the context of serialization. As I see it, we should use the term "SPDX 3 JSON" instead of "SPDX 3 JSON-LD" and clarify that this format may be deserialized into RDF using JSON-LD libraries, but not the other way around.

bact:
A work item from 15 Apr 2025 Tech Team meeting
Will fix https://github.com/spdx/spdx-3-model/issues/1008

- Change a section heading from "Serialization in JSON-LD" to "Serialization in SPDX 3 JSON"
- Add SPDX 3 JSON format description, based on an agreed description in https://github.com/spdx/spdx-3-model/issues/1008#issuecomment-2755093934 with a revision for formality.

The original description contains the word "our". While it is good for documents like tutorials, it is not recommended in general to be used in a formal specification as it can create a sense of inclusion/exclusion, our/them.

Original description:

> SPDX 3 JSON is a strict subset of JSON-LD, meaning it must be serialized according to our strict specification and validated using the provided JSON schema, but it may be parsed — not serialized — using standard JSON-LD libraries.

Revised, removing "our":

> The SPDX 3 JSON format is a strict subset of JSON-LD. It requires data to be serialized according to the defined serialization specification and validated against the SPDX 3 JSON Schema. It may be parsed – not serialized – using standard JSON-LD libraries.

### DIFF
--- a/docs/serializations.md
+++ b/docs/serializations.md
@@ -93,9 +93,8 @@ element.
 ## Serialization in SPDX 3 JSON
 
 The SPDX 3 JSON format is a strict subset of JSON-LD.
-Instances of SPDX 3 JSON shall be serialized in strict accordance with the
-defined serialization specification and validated against the provided
-SPDX 3 JSON Schema.
+It requires data to be serialized according to the defined serialization
+specification and validated against the SPDX 3 JSON Schema.
 It may be parsed – not serialized – using standard JSON-LD libraries.
 
 ### JSON-LD context file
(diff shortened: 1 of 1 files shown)


bact:
@ilans @kzantow @JPEWdev from 15 Apr 2025 Tech Team meeting. Please kindly review.

JPEWdev (approved):

goneall (approved):
LGTM - Thanks @bact

PR #1200, opened 2025-04-17
[support/3.0]: Fix typo license expression annex

bact:
> **This PR targets `support/3.0` branch**

> For the PR that targets `develop` branch, see #1192 (merged)

- Fix typos
    - "of a `<license-expression>` **a** is:" -> "of a `<license-expression>` is:" (remove extra 'a')
    - "either LGPL-2.1-only **and** the expression" -> "either LGPL-2.1-only **or** the expression" (change 'and' to 'or')
    - "both the LGPL-2.1-only **or** MIT licenses" -> "both the LGPL-2.1-only **and** MIT licenses" (change 'or' to 'and')
- Fix pattern of user defined license ref, in description of "Simple license expressions" section:
    - Remove `1*` from the pattern, `1*` is not in the full ABNF syntax at the beginning of the annex 
    - Pattern of User defined addition ref does not have this `1*` too
- Fix `[idString]` -> `(idstring`); use the same style and casing as in the ABNF syntax
- Syntax highlight ABNF and XML code blocks
- Adjust indent of user defined license ref examples to be under the same bullet level as the user defined license ref itself (to avoid confusion if the examples are for the license ref or for the entire simple license expression):

    Before:
    Screenshot 2568-04-15 at 01 11 38

    After:
    Screenshot 2568-04-15 at 01 07 57

### DIFF
--- a/docs/annexes/spdx-license-expressions.md
+++ b/docs/annexes/spdx-license-expressions.md
@@ -46,7 +46,7 @@ A valid `<license-expression>` string consists of either:
 
 (ii) a more complex expression constructed by combining smaller valid expressions using Boolean license operators.
 
-There MUST NOT be white space `.
+There MUST NOT be white space between a license-id and any following `+`. This supports easy parsing and backwards compatibility. There MUST be white space on either side of the operator "WITH". There MUST be white space and/or parentheses on either side of the operators `AND` and `OR`.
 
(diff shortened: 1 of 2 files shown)


goneall (approved):
LGTM - Thanks @TTMaZa and @bact

PR #1201, opened 2025-04-17
Explaining the AND-operator with an less ambiguous example that contains an "and"  instead of an "or"

TTMaZa:
Hi there,

I'd like to work out the difference between Disjunctive and  Conjunctive operator a bit more.

I would go even further and add some explanation. e.g. some text about  Conjunctive is used in cases, where a piece of software that is marked as "LGPL-2.1-only and MIT" often times is a "combined work with LGPL-2.1 parts and MIT parts".

Whereas a piece of software that is marked as "LGPL-2.1-only or MIT" is considered to be Dual-Licened and the recipient typically can choose one and ignore the other.

But before I go any deeper into that I'd like to get some feedback on this.

### DIFF
--- a/docs/annexes/spdx-license-expressions.md
+++ b/docs/annexes/spdx-license-expressions.md
@@ -119,7 +119,7 @@ It is allowed to use the operator in lower case form `or`.
 
 If required to simultaneously comply with two or more licenses, use the conjunctive binary "AND" operator to construct a new license expression, where both the left and right operands are valid license expression values.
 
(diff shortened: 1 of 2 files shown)


bact:
That's a typo that we should fix.

- A PR to fix this (and other typos) in `support/3.0` branch is at #1200
- A PR to fix this (and other typos) in `develop` branch is at #1203

TTMaZa:
Thanks for the response. So I guess I can close this Pull request, since the change is integrated via #1203

bact:
Thank you @TTMaZa for this PR.
(Have put your name as co-author in another PR)

PR #1205, opened 2025-04-25
Add back publish_common workflow for 2.3 publication

bact:
- A workflow needs to be in the default branch in order to appear on Actions tab on GitHub UI (or be available via GitHub API) and can be triggered through workflow_dispatch.
- This PR adds back `publish_common.yml` workflow to `develop` branch (the default branch of this repo). The workflow [is originally from](https://github.com/spdx/spdx-spec/blob/support/2.3/.github/workflows/publish_common.yml) `support/2.3` branch.
- The `publish_common.yml` workflow is needed for the publication of v2.3 (and v2.3.1 draft) in the multiversion environment (using mike).

### DIFF
--- a/.github/workflows/publish_common.yml
+++ b/.github/workflows/publish_common.yml
@@ -1,0 +1,37 @@
+on:
+  repository_dispatch:
+    types:
+      - publish_spec
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: Branch or tag to publish (e.g. refs/heads/support/2.3).
+        required: true
+        default: refs/heads/support/2.3
+      aliases:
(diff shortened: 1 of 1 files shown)


bact:
publish_common.yml is a workflow that designed to allow us to control the publication, with ability to pass parameters to the workflow, without editing the workflow YAML file. It is designed for v2.3 and does not yet support v3.0 (as v3.0 doc generation requires additional Markdown files generation from spec-parser). 

In a short term, we will limit its use with only v2.3.x as it is intended for.
In a longer term, we can update it to support v3.0, multilingual publication, and staging.

goneall (approved):
LGTM

PR #1206, opened 2025-04-26
Make publish_common workflow to work with both v2.2.2 and v2.3.x

bact:
> This PR targets `develop` branch, which is the default branch - so the workflow can be available in GitHub web interface. The workflow itself is used for the publication of v2.x branches. See a [similar workflow](https://github.com/spdx/spdx-spec/blob/support/2.3/.github/workflows/publish_common.yml) in `support/2.3` branch.

- Specify working-directory to fix error when set Git identity (see [run](https://github.com/spdx/spdx-spec/actions/runs/14681319300/job/41204330746#step:7:6))
- Retain the schema doc generation (was previously in publish.yml for 2.x)
- Specify Python dependencies from within the workflow (not `requirements.txt`)
  - Package versions in local `requirements.txt` can be outdated and no longer works with current version of Python.
  - For example, `development/v2.2.2` branch's requirements.txt has mkdocs==1.3.1 which its dependencies no longer installable in more recent versions of Python. 
- The updated workflow should work for the publication of content in `development/v2.2.2`, `support/2.3` and `support/2.3.1` branches.

### DIFF
--- a/.github/workflows/publish_common.yml
+++ b/.github/workflows/publish_common.yml
@@ -19,15 +19,15 @@ on:
   workflow_dispatch:
     inputs:
       ref:
-        description: Branch or tag to publish (e.g. refs/heads/support/2.3.1).
+        description: Branch or tag to publish (e.g. refs/heads/support/2.3).
         required: true
-        default: refs/heads/support/2.3.1
+        default: refs/heads/support/2.3
       as_version:
-        description: Version to publish as (e.g. v2.3.1).
+        description: Version to publish as (e.g. v2.3).
         required: true
-        default: v2.3.1
+        default: v2.3
       aliases:
-        description: Space-delimited aliases to publish (e.g. v2-latest v2-draft v2.3.1-dev).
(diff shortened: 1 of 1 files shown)


bact:
JSON Schema browser:
- The updated workflow will also allow us to have v2.2 JSON schema browser publishes in a versioned URL
  - Currently v2.2.2 schema browser is published at https://spdx.github.io/spdx-spec/spdx-json-schema.html (the URL does not have a version)
  - While v2.3 and v2.3.1-dev ones are at https://spdx.github.io/spdx-spec/v2.3/spdx-json-schema.html and https://spdx.github.io/spdx-spec/v2.3.1-dev/spdx-json-schema.html
  - The updated workflow will publish v2.2.2 schema to https://spdx.github.io/spdx-spec/v2.2.2/spdx-json-schema.html
    - The existing resource at https://spdx.github.io/spdx-spec/spdx-json-schema.html will still remain, untouched.

goneall (approved):
LGTM

kestewart (approved):
Discussed with Gary, and agreed can be merged.

PR #1207, opened 2025-04-26
[2.3] publish_common workflow: Specify working dir + Fix MkDocs warning

bact:
> This PR targets `support/2.3` branch.

- Specify working-directory to fix error when set Git identity (see [run](https://github.com/spdx/spdx-spec/actions/runs/14681319300/job/41204330746#step:7:6))
- Replace the deprecated `::set-output` with environment file output
- Fix MkDocs warning that `foreword.md` is not in 'nav' in mkdocs.yml.
  - The file is intentionally not included in 'nav' because it is used for ISO-submitted version only. (v2.3 is not)
  - Suppress the warning by mark it in 'exclude_docs'


goneall (approved):
LGTM

PR #1215, opened 2025-05-20
[3.0] Fix ID of license exception example

bact:
> This PR targets `support/3.0` branch.

Remove an extra "d".

The ID `SPDXRef-ButIDontWantToException` comes from `"But I Don't Want To" Exception`.


goneall (approved):
LGTM - Good catch @bact

PR #1216, opened 2025-05-29
primaryPackagePurpose in 2.3 defines OPERATING-SYSTEM not OPERATING_SYSTEM

sil2100:
Per the specification in https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field the `primaryPackagePurpose` enum should contain `OPERATING-SYSTEM` not `OPERATING_SYSTEM`.

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -417,7 +417,7 @@
           "primaryPackagePurpose" : {
             "description" : "This field provides information about the primary purpose of the identified package. Package Purpose is intrinsic to how the package is being used rather than the content of the package.",
             "type" : "string",
-            "enum" : [ "OTHER", "INSTALL", "ARCHIVE", "FIRMWARE", "APPLICATION", "FRAMEWORK", "LIBRARY", "CONTAINER", "SOURCE", "DEVICE", "OPERATING_SYSTEM", "FILE" ]
+            "enum" : [ "OTHER", "INSTALL", "ARCHIVE", "FIRMWARE", "APPLICATION", "FRAMEWORK", "LIBRARY", "CONTAINER", "SOURCE", "DEVICE", "OPERATING-SYSTEM", "FILE" ]
           },
           "releaseDate" : {
             "description" : "This field provides a place for recording the date the package was released.",


xnox (approved):
Note that trivy spdx json generator produces and expects dash'ed enum. Thus it seems like dashed version is the one in active use.

xnox:
Likely a new minor version of the specification should be published, like 2.3.1 because it seems like dozens of projects include the schema and would need to be updated

https://github.com/search?q=%22%5C%22SOURCE%5C%22%2C+%5C%22DEVICE%5C%22%2C+%5C%22OPERATING_SYSTEM%5C%22%22&type=code

Note sure how many documents use the field however, and at least Trivy uses the dashed constant.

bact (approved):
LGTM  - per https://github.com/spdx/tools-java/pull/211

goneall (approved):
Agree that the spec states using a dash and the schema should be consistent

goneall:
@bact can you look into the CI failures?  Looks like the pip install is erroring out.

bact:
branch support/2.3.1 PR validation workflow is using Python "3" which is the latest stable 3.x version, which is 3.13.

Python 3.13 removed the "cgi" module but one of the dependencies required that.

We can specify the workflow's Python version number to <3.13.

bact:
I can open another PR to fix that and once that is merged, we can rerun CI here.

bact:
Please review and merge #1217 first and rerun validation workflow for this PR.

goneall:
@bact -  ran into a new error after rebasing

bact:
@sil2100 do you mind to update your `sil2100:2.3-os` branch with latest changes in `spdx:support/2.3.1` please? So the validate_pull_request.yml get updated. Thank you.

bact:
#1219 is logging changes from this PR

goneall:
@bact  - I'm not sure rebasing will fix the CI error.

When I compared the branches, I don't see any changes in the CI or requirements files that would change.

Reference: https://github.com/spdx/spdx-spec/compare/support/2.3.1...sil2100:spdx-spec:2.3-os

Am I missing something?  Not sure why this PR is failing where the other more recent PRs are succeeding.

bact:
The CI in sil2100:2.3-os branch is looking for build-requirements.txt. While the updated CI no longer uses that and the file is already removed.

The updated CI specifically uses Python 3.12

bact:
In other words, the previous PR validation CI is broken in itself (it used to work with Python <= 3.12, but broken in 3.13) , not related to the content of this PR

bact:
@goneall To check if the PR validation CI works properly, I have created PR #1225 to test. It has similar change as this PR and it passed the validation CI.

goneall:
Thanks @bact for the explanation of the CI failure - should I just go ahead and merge this PR?  It seems like the CI will pass after merging.

bact:
I strongly believe that it will pass

PR #1217, opened 2025-05-29
[2.3.1] Update build dependencies and GitHub Actions

bact:
Update validation and publication workflow for v2.3.1-dev

Current workflow files are outdated and break because of breaking change in Python 3.13 (see https://github.com/spdx/spdx-spec/pull/1216#issuecomment-2920204136 )

The workflow update is based on the same updated files in `support/2.3` branch.


goneall (approved):
Thanks @bact - lgtm

PR #1221, opened 2025-05-30
[2.3] Publish schema doc

bact:
> This PR targets `support/2.3` branch.

- Generate schema doc and put it in chapters directory so mkdocs/mike can copy it and make it available in generated website.
  - The schema doc for v2.2 looks like this: https://spdx.github.io/spdx-spec/spdx-json-schema.html
  - This PR will put the schema doc to https://spdx.github.io/spdx-spec/VERSION/spdx-json-schema.html - for example, https://spdx.github.io/spdx-spec/v2.3/spdx-json-schema.html
  - This generation step was previously run inside https://github.com/spdx/spdx-spec/blob/support/2.3/.github/workflows/publish.yml but that workflow is no longer in use because it is a workflow for single version publication (using mkdocs only and not using mike).

- Also remove mike installation (redundant with pip install -r requirements.txt)

### DIFF
--- a/.github/workflows/publish_common.yml
+++ b/.github/workflows/publish_common.yml
@@ -1,7 +1,8 @@
 # Publish pre-3.0 specs (spec versions that do not require spec-parser)
 # This workflow can be triggered manually or by a repository dispatch event.
-# It has two parameters:
+# It has three parameters:
 # - ref: The branch or tag to publish
+# - as_version: The version to publish as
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

bact:
The v2.3 schema doc is now available at https://spdx.github.io/spdx-spec/v2.3/spdx-json-schema.html

PR #1225, opened 2025-05-31
[2.3.1] Fix primaryPackagePurpose enum to use dash

bact:
It should be "OPERATING-SYSTEM" (with dash) instead of "OPERATING_SYSTEM" (with underscore)

Same as #1216 / created to test CI


bact:
#1216 is merged. This PR is no longer needed. Closing now

PR #1230, opened 2025-07-04
[2.3.1] Fix cardinality of externalDocumentRef to 0..*

### LINKED ISSUE #812: Cardinality on external document references is wrong
Another small one: The spec for [external document references](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#66-external-document-references-field) says that it's not required, but cardinality is `1..*`. I strongly suspect it should be `0..*` instead.

bact:
> This PR targets `support/2.3.1` branch.
> For a similar PR targets `support/2.3`, see #1229

- The external document references is not required in the spec (Required = No), but its cardinality is `1..*`
- This PR fixes the cardinality typo to `0..*`, to make it consistent with Required = No
- To fix #812

### DIFF
--- a/chapters/document-creation-information.md
+++ b/chapters/document-creation-information.md
@@ -227,7 +227,7 @@ Identify any external SPDX documents referenced within this SPDX document. The m
 | Attribute | Value |
 | --------- | ----- |
 | Required | No |
-| Cardinality | 1..* |
+| Cardinality | 0..* |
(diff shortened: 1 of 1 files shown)


goneall (approved):
Good catch

bact:
@nicoweidner Please review.

nicoweidner (approved):

kestewart (approved):

PR #1238, opened 2025-07-08
[3.0] "this standard/specification" -> "this document"

bact:
> This PR targets `support/3.0` branch.
> For a similar PR that targets `develop` branch, see #1239

- To fix https://github.com/spdx/spdx-spec/issues/1234 (ISO issue)

PR #1239, opened 2025-07-08
[develop] "this standard/specification" -> "this document"

### LINKED ISSUE #1234: Spec should be referred to as "this document"
Only "this document" is allowed.

There are case of "this standard", "this specification", ...

bact:
> This PR targets `develop` branch.
> For a similar PR that targets `support/3.0`, see #1238 

- To fix https://github.com/spdx/spdx-spec/issues/1234 (ISO issue)

### DIFF
--- a/docs/annexes/pkg-url-specification.md
+++ b/docs/annexes/pkg-url-specification.md
@@ -27,7 +27,7 @@ A _purl_ is a URL composed of seven components:
 
 Components are separated by a specific character for unambiguous parsing.
 
-The definition for each components is:
+The definition for each component is:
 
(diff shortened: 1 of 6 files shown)


zvr (changes requested):
Same as in #1238 

The "document" changes are correct, but "more than one checksums" should remain plural.

bact:
> Same as in #1238
> 
> The "document" changes are correct, but "more than one checksums" should remain plural.

It should read:

> If the key is checksum and there is more than one checksum,

with "checksum" in singular form. See rational at https://github.com/spdx/spdx-spec/pull/1238#issuecomment-3053094314

kestewart (approved):
LGTM

goneall (approved):
LGTM

PR #1242, opened 2025-07-09
[3.0] RDF URL "3.0.1" -> "3.0"

bact:
> This PR targets `support/3.0` branch.
> For a similar PR that targets `develop` branch, see #1243

- To fix https://github.com/spdx/spdx-3-model/issues/1046 (ISO issue)

See also similar fix in the spdx-3-model repo:

- https://github.com/spdx/spdx-3-model/pull/1051

PR #1243, opened 2025-07-09
[develop] RDF URL "3.0.1" -> "3.1"

### LINKED ISSUE #1046: The RDFs should only go to second-level version
For publishing the specification, RDF should remain constant, even though textual changes happen in the specification.

Therefore, instead of `https://spdx.org/rdf/3.0.1/terms/Core/Element` they should simply be `https://spdx.org/rdf/3.0/terms/Core/Element`

These are defined in the profile definition files (like `Core.md`).

See also https://github.com/spdx/spdx-spec/issues/1237

bact:
> This PR targets `develop` branch.
> For a similar PR that targets `support/3.0` branch, see #1242 

- To fix https://github.com/spdx/spdx-3-model/issues/1046 (ISO issue)

- Also need #1249 in order for these URLs to actually work:
  - https://spdx.org/rdf/3.1/spdx-model.ttl
  - https://spdx.org/rdf/3.1/spdx-context.jsonld
  - https://spdx.org/schema/3.1/spdx-json-schema.json

--

See also similar fix in the spdx-3-model repo:

- https://github.com/spdx/spdx-3-model/pull/1052


bact:
Same CI failure as in #1242 

The examples validation failed because examples/jsonld/package_sbom.json JSON-LD context URL is `https://spdx.org/rdf/3.0/spdx-context.jsonld` but the specVersion is `3.0.1`.

Would need to update the validator to allow MAJOR.MINOR URL without the PATCH version?

bact:
Note that this PR does not include the update of URL in one SBOM example (`examples/jsonld/package_sbom.json`) as it will failed the example validation CI.

- The fix proposal is put separately in #1244 (when it works, will copy the new `bin/check-examples.sh` to `support/3.0` branch)

bact:
This is also a follow-up from Tech team meeting on 19 Aug 2025
(that one https://github.com/spdx/spdx-3-model/pull/1052 (merged) is in spdx-3-model repo, this one is in spdx-spec)

goneall:
We don't have the schema's and ontologies copied to spdx.org for version 3.1 yet.

I'm a little reluctant to add them before the release.

We could:
1. Go ahead and copy over the in development versions of the schemas and ontology files then merge the PR
2. Merge this PR which would have broken links
3. Wait to merge this PR until after 3.1 is released

@bact @zvr @kestewart - thoughts?

kestewart (approved):

goneall (approved):
LGTM

PR #1253, opened 2025-07-25
Remove personal pronoun use

RexJaeschke:
ISO does not permit personal pronouns in specs.

### DIFF
--- a/docs/front/introduction.md
+++ b/docs/front/introduction.md
@@ -45,6 +45,6 @@ Exchange to System Package Data Exchange and expanding the scope of items it
 can now convey in a bill of materials from software, security, and licensing to
 many additional aspects like data sets, AI models, and build information.
 
-Since the release of 3.0.0, we have gathered feedback on the level of
+Since the release of 3.0.0, the groups have gathered feedback on the level of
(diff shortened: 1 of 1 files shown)


kestewart (approved):

bact (approved):

zvr:
@bact , @RexJaeschke please note that all such changes must also be reflected (duplicated) into our `main` branch.
It would be better if both such PRs are merged at the same time (next to each other). Otherwise I fear we're going to miss some changes, resulting into duplicate work down the line.

bact:
Agree that all changes must be bring forward to versions after 3.0.1 as well.
Our post-3.0.1 version is now in `develop` branch, which is also keep updating for the future 3.1.

To my understanding, the `main` branch will be a copy of the most recent stable version.
For example, if SPDX 3.1 is released, the `main` branch will be SPDX 3.1.
Or if ISO SPDX 3.0 is released before SPDX 3.1, the `main` branch will be the ISO SPDX 3.0 (and will eventually overwritten by SPDX 3.1 at some point).

Will track keeping with [3.0.2 milestone](https://github.com/spdx/spdx-spec/milestone/13) be enough to address the concern of missing some changes?

Issue #1265, opened 2025-08-29
Something is wrong with the gh-pages branch


goneall:
When I pull the latest gh-branches using `git switch gh-pages` and run `git status` I get the following modifications:

```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   v3.0.1/annexes/SPDX-Lite/index.html
        modified:   v3.0.1/annexes/SPDX-license-expressions/index.html
        modified:   v3.0.1/licenses/CC-BY-3.0/index.html
        modified:   v3.0.1/licenses/Community-Spec-1.0/index.html
        modified:   v3.1-dev/annexes/SPDX-Lite/index.html
        modified:   v3.1-dev/annexes/SPDX-license-expressions/index.html
        modified:   v3.1-dev/licenses/CC-BY-3.0/index.html
        modified:   v3.1-dev/licenses/Community-Spec-1.0/index.html
```

To make sure this wasn't just my local issue, I deleted the spdx-spec and re-cloned it - same issue.

The problem appears to be in the GitHub repo.

Unfortunately, I can't restore, reset, remove, clear cache etc - to switch back to main I have to delete the directory and re-clone.

Curious is anyone else is able to reproduce this behavior.

bact:
I have got similar issue:

I run:

`git checkout --track upstream/gh-pages`

Then:

`git status`

And got this:

```text
Refresh index: 100% (7335/7335), done.
On branch gh-pages
Your branch is up to date with 'upstream/gh-pages'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   3-draft/annexes/SPDX-Lite/index.html
	modified:   3-draft/annexes/SPDX-license-expressions/index.html
	modified:   3-draft/licenses/CC-BY-3.0/index.html
	modified:   3-draft/licenses/Community-Spec-1.0/index.html
	modified:   3.0-RC1/annexes/SPDX-Lite/index.html
	modified:   3.0-RC1/annexes/SPDX-license-expressions/index.html
	modified:   3.0-RC1/licenses/CC-BY-3.0/index.html
	modified:   3.0-RC1/licenses/Community-Spec-1.0/index.html
	modified:   3.0-RC2/annexes/SPDX-Lite/index.html
	modified:   3.0-RC2/annexes/SPDX-license-expressions/index.html
[...]

PR #1269, opened 2025-09-11
Update delay in redirection template to fix Git issue in gh-pages branch

bact:
- Change redirect delay from 0 to 1 second, to match the value used in the redirection template generated by Mike
- This will avoid introducing changes to the Git and possibly resolve issue #1265


goneall (commented):
LGTM - let's give this a try

goneall (approved):

Issue #1236, opened 2025-07-08
Use "shall" and "may"


zvr:
ISO has detailed rules about the verb forms one must use to indicate compliance. These involve “may,” “may not,” “shall,” “shall not,” among others. (See the [ISO Directives Part 2](https://www.iso.org/sites/directives/current/part2/index.xhtml), Clause 7, "Verbal forms for expressions of provisions.")

ISO does **not** allow the use of [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt) "MUST", "MAY", "SHALL".

In particular, “must” can only appear in informative text—such as notes, examples, and informative annexes.

[This comment was tweaked by Rex on 2025-07-25.]

RexJaeschke:
@kestewart, @zvr, @goneall, @bact 

**Assigned to SPDX team**

Here are some specific things I noticed:

- There are numerous occurrences of “shall” and none of “SHALL”. Good!
- There are numerous occurrences of “MUST” and many of “must,” both suggesting compliance requirements. This is a hybrid of the ISO and RFC 2119 approaches. However, ISO does not permit “must” in any form for a requirement; instead, “shall” must be used. 
- There are more than a few occurrences of “must NOT,” an odd combination of casing. If this is normative, it needs to "shall not"; otherwise, if informative, "must not".
- There is one occurrence of “MAY” and many of “may,” both suggesting permission. This is a hybrid of the ISO and RFC 2119 approaches. Also, it appears that some uses of “may” might really mean “might.” “may” implies permission w.r.t compliance.
- There are a lot of occurrences of “should” some of which, at a glance, might need to be changed to “shall” if they are intended to be requirements.

goneall:
@zvr - was this already fixed in a previous PR?  If so, please go ahead and close.

goneall:
From the 14 Oct 2025 tech call:

- Need to do search / replace - @bact will create PRs for both support/3.0 and development for both the spec and model repo

bact:
Will work on this using verb forms from https://www.iso.org/sites/directives/current/part2/index.xhtml#_idParaDest-29

bact:
> * There are more than a few occurrences of “must NOT,” an odd combination of casing. If this is normative, it needs to "shall not"; otherwise, if informative, "must not".

Looks like suggesting that we should keep things in normal casing and NOT IN UPPERCASE. 

[...]

PR #1283, opened 2025-10-17
Use ISO shall: pkg-url-specification

bact:
Use "shall" instead of "must", for Package URL specification.

To address issue in #1236


zvr (approved):

goneall (approved):
LGTM

PR #1284, opened 2025-10-17
Use ISO shall: serializations

bact:
Use "shall" instead of "must", for Model and serializations.

To address issue in #1236


zvr (approved):

goneall (approved):
LGTM - Thanks @bact

PR #1281, opened 2025-10-17
Update version (3.1) and year (2025) in copyright notice

bact:
Update for 3.1 in `develop` branch

PR #1289, opened 2025-10-21
[3.0] Update version (3.0) and year (2025) in copyright notice

bact:
> This PR targets `support/3.0` branch.
> For its counterpart that targeting `develop` branch, see #1281

- Update version reference to address https://github.com/spdx/spdx-spec/issues/1237
  > The spec should only refer to second-level numbering (never 3.0.1, for example), and in a consistent manner "SPDX 3.0".)
- Update year in copyright notice to cover 2025
- Use en dash between years, as suggested in https://github.com/spdx/spdx-spec/pull/1281#pullrequestreview-3349082316

### DIFF
--- a/docs/annexes/rdf-model.md
+++ b/docs/annexes/rdf-model.md
@@ -40,11 +40,11 @@ and is published in online at
 
 [![Extension Profile diagram][fig_extension]][fig_extension]
 
-[fig_ai]: ../images/model-AI.png "SPDX 3.0.1 AI Profile diagram"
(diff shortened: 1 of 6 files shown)


zvr (approved):
It also changes "Profile" to lowercase in the figure labels in the diagrams annex, but that's fine.

bact:
> It also changes "Profile" to lowercase in the figure labels in the diagrams annex, but that's fine.

Yes. That was intentional. So the PR #1292 will not touch those lines and create conflicts.

PR #1294, opened 2025-10-22
Terms add software hardware

gregshue:
Added proposed definitions for the following terms:

- computer hardware
- computer program
- firmware
- hardware
- information system
- software
- system

These are needed to clearly identify which definition we are using form these common terms, especially  due to SPDX changing scope from Software to System.

### DIFF
--- a/docs/terms-and-definitions.md
+++ b/docs/terms-and-definitions.md
@@ -8,6 +8,11 @@ at the following addresses:
 - IEC Electropedia: available at
   [http://www.electropedia.org/](http://www.electropedia.org/)
 
+**computer hardware**
+
+All or part of the physical components of an information system. 
+(Leveraged from 'hardware' definition in ISO/IEC 2382:2015 Information technology -- Vocabulary)
+
 **computer program**
 
 Combination of computer instructions and data definitions 
--- a/docs/terms-and-definitions.md
+++ b/docs/terms-and-definitions.md
@@ -8,6 +8,16 @@ at the following addresses:
 - IEC Electropedia: available at
   [http://www.electropedia.org/](http://www.electropedia.org/)
 
+**firmware**
+
+Computer programs and data stored in hardware - 
+typically in read-only memory (ROM) or programmable read-only memory (PROM) - 
+such that the programs and data cannot be dynamically written or modified 
+during execution of the programs. 
+(From https://csrc.nist.gov/glossary/term/firmware, 
+from CNSSI 4009-2015, 
+**leveraged** from IETF RFC 4949 Ver 2 at https://datatracker.ietf.org/doc/html/rfc4949)
+
 **hardware**
 
 The material physical components of a system. 
--- a/docs/terms-and-definitions.md
+++ b/docs/terms-and-definitions.md
@@ -8,6 +8,12 @@ at the following addresses:
 - IEC Electropedia: available at
   [http://www.electropedia.org/](http://www.electropedia.org/)
 
+**computer program**
+
+Combination of computer instructions and data definitions 
+that enable computer hardware to perform computational or control functions. 
+(From ISO/IEC/IEEE 24765:2017 Systems and software engineering-Vocabulary)
+
 **firmware**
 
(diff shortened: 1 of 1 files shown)


gregshue:
The ISO/IEC publishing of this spec considers the terms and definitions to be informative content and freely available. Are there any licensing or copyright issues with reusing publicly accessible definitions?

docs/terms-and-definitions.md
```diff
@@ -8,9 +8,66 @@ at the following addresses:
...
+All or part of the physical components of an information system. 
+(Leveraged from 'hardware' definition in ISO/IEC 2382:2015 Information technology -- Vocabulary)
+
+**computer program**
+
+Combination of computer instructions and data definitions 
+that enable computer hardware to perform computational or control functions. 
+(From ISO/IEC/IEEE 24765:2017 Systems and software engineering-Vocabulary)
+
+**firmware**
```
zvr, line 22: The term "firmware" is not really used in the spec.

It appears in exactly one instance, in the `Software/SoftwarePurpose` vocabulary.

> - firmware: The Element provides low level control over a device's hardware.

But if we start defining all these terms, where will we stop? We also have software application, archive, configuration data, container image, device, device driver, executable, software library, platform, source, ...

zvr (changes requested):
If we are to add these, all the referenced material has to be added to the list of references.

docs/terms-and-definitions.md
gregshue: We must define the terms that have alternate definitions across the ISO and IEC lookups to unambiguously describe the concepts to the readers. This is needed to enable portability between document producers vs consumers. It is also needed to drive toward consistent conversions of encodings between SPDX and other encodings.

When SPDX changed from "Software" to "Systems" it has to resolve ambiguity introduced by the much larger domain. For instance, ISO 22006:2009(en) defines **product** as "end result of a farm process". We must allow the list of terms to grow until we have disambiguated the spec.

> The term "firmware" is not really used in the spec.

Thank you for identifying this, but I think you missed the point. The term "firmware" **is** used in the spec, a definition is provided, and that definition does not align with the definition found in multiple IEC standard. SPDX must define this term to disambiguate it and define it in such a way to determine what SPDX profile to use for it (e.g., Hardware Profile vs Software Profile). 

BTW, a related question exists for describing FPGA bitstreams. I have not found a definition for these in either the IEC or ISO lookups. Whatever definition we come up with will eventually appear in the IEC lookup. Should we define the term "gateware" to cover these?

gregshue:
> If we are to add these, all the referenced material has to be added to the list of references.

I'll update the reference material once we align on definitions.

docs/terms-and-definitions.md
goneall: Discussed on tech call 4 Nov 2025:
- Agree the definitions need to align between the glossary and use within the spec
- Discussion on how many of the vocabulary terms need to be represented in the glossary
- Suggestion: If there is a reference to a specification in the Vocabulary and it is only used within a single profile, it doesn't need to be in the glossary

goneall:
Reviewed on the 4 Nov 2025 tech call:
- No one objected to the definitions or inclusion of the terms - with the possible exception of the comments related to firmware
  - Note that in the comment we came up with a suggestion on how to decide which terms would be defined in the glossary
- Some discussion on the Software definition - it's inclusion of "data" - agreed after discussion it was OK - no conflict with DataSET profile

kestewart (approved):
Looks reasonable to me.   They all have citations at this point, and I'm not seeing additional comments.

goneall (approved):
Discussed on 11 Nov call - approved

kestewart:
Discussed in call.   Greg to add in Normative references in separate PR.

bact:
For IETF RFC links, try to keep it in one format:

`https://datatracker.ietf.org/doc/<number>/`

According to https://github.com/spdx/spdx-3-model/blob/develop/docs/format.md#writing-style

PR #1296, opened 2025-10-30
Keep spec website fresh - rebuild periodically

bact:
- Schedule workflow to rebuild dev spec website periodically, fetching latest changes from both spdx-spec and spdx-3-model repos (`develop` branches)
  - While PR merges in the spdx-spec repo will trigger the website rebuild, PR merges in spdx-3-model repo will not. This PR will make the dev spec website as fresh as possible towards the push for the new release.
- Will rebuild every 30 minutes during meeting hours (from https://github.com/spdx/meetings/)
- Will rebuild every two hours off meeting hours
- Also add "develop" URL alias to redirect to v3.1-dev (to fix dead link https://spdx.github.io/spdx-spec/develop/ in README)
  https://github.com/spdx/spdx-spec/blob/265717697f0312058fa90db0fdd74ab32171a6e7/README.md?plain=1#L22-L26


goneall:
I wonder if there is a way to trigger the build when the model changes.  For the LicenseListXML repo, we push commits to the license-list-data repo on merges.  We could do something similar where we push some kind of update to the spec repo.  It could be as simple as a text file with the last commit hash of the model repo or it could be a complete copy of the markdown files.

bact:
@goneall That is possible. Need to setup at both side, at the Repo A that will trigger / dispatch the rebuilding signal and at the Repo B that will receive the signal. This way, once changes are made in Repo A, Repo B will be rebuilt. 

I never setting that up myself, but looks like a doable thing from here
https://www.nieknijland.nl/blog/how-to-trigger-a-github-action-workflow-in-another-repository
It probably needs a token for authentication - but there can be other way.

zvr (approved):

zvr:
Merging this, to have periodic updates, and we can see about triggers from the model repo in another PR.

PR #1299, opened 2025-10-30
Full example of an SPDX document using all classes

goneall:
This is just a start on the full example.  It only includes the full core and software classes - other classes will be added.

The relationship completeness properties will also be added in the future.

This was generated from the SPDX Java Tools example in this pull request: https://github.com/spdx/tools-java/pull/240


goneall:
@JPEWdev  - let me know if I'm on the right track and if this is usable for the HTML generator

BTW - decided to write the example by hand - I needed an example for the Java library and any generated code seemed to not make much sense for some of the properties.

goneall:
It looks like the CI is tripping on the external Agent reference - not sure if this is an issue with the source or the validator

bact:
pyshacl failed. It can't handle external IDs.

I have tried spdx3-validate the example, it is valid.

If spdx3-validate can already handle SHACL validation, we may like to remove pyshacl from validation workflow.

goneall:
> pyshacl failed. It can't handle external IDs.
> 
> I have tried spdx3-validate the example, it is valid.
> 
> If spdx3-validate can already handle SHACL validation, we may like to remove pyshacl from validation workflow.

Create PR #1300 to remove pyshacl

JPEWdev:
> @JPEWdev - let me know if I'm on the right track and if this is usable for the HTML generator
> 
> BTW - decided to write the example by hand - I needed an example for the Java library and any generated code seemed to not make much sense for some of the properties.

Yes, this looks pretty good

bact:
Wonder if eventually we have to update all the SPDX IRIs from "3.0.1" to "3.0" ?


According to issue spdx/spdx-3-model#1046

zvr:
@goneall may I suggest to move this to the model repo?

Having it in the spec repo would make it harder to keep it in sync whenever we make changes to the model (two PRs, etc.)

examples/jsonld/spdx-full-example.json
```diff
@@ -0,0 +1,234 @@
...
+    "type" : "Organization",
+    "name" : "System Package Data Exchange (SPDX)",
+    "creationInfo" : "_:creationInfo_0"
+  }, {
+    "spdxId" : "https://spdx.github.io/spdx-spec/v3.0.1/examples/full-example-eaa46bdcfa20#softwareagent",
+    "type" : "SoftwareAgent",
+    "name" : "SPDX Spec Github CI",
+    "creationInfo" : "_:creationInfo_0"
+  }, {
+    "spdxId" : "http://spdx.org/licenses/Apache-2.0",
```
zvr: In general, this would be an external reference to this object, not included verbatim here (the same with its creator).

zvr (changes requested):
Even if we include the license, for the example we might want to choose a much smaller one ;-)

goneall:
> Wonder if eventually we have to update all the SPDX IRIs from "3.0.1" to "3.0" ?
> 
> According to issue [spdx/spdx-3-model#1046](https://github.com/spdx/spdx-3-model/issues/1046)

The context file currently refers to the 3.0.1 - so we would need to update that as well.

We could fix this when we do a 3.0.2 release at which time we can switch back to just the URI's without the patch level.

examples/jsonld/spdx-full-example.json
goneall: Agree - Will change in the next update

goneall:
> @goneall may I suggest to move this to the model repo?
> 
> Having it in the spec repo would make it harder to keep it in sync whenever we make changes to the model (two PRs, etc.)

We could move it to the model repo.  I'm thinking the spdx-spec CI would publish the HTML based on this example using the code @JPEWdev created.  Since we already pull in the model code in the CI, it doesn't really matter if the examples are here or in the model repo.

If we move it, should we move the entire examples directory?  Seems like it would have the same benefit even though the example is smaller.

You raise a good point about maintaining this file - something I've been thinking about. It's actually a bit worse since I'm currently generating it from an example Java app.  This will make it hard for non-Java programmers to contribute / maintain.  I took this approach since it is much easier for me to create than to do it by hand or use another language binding I'm less familiar with.  I was originally planning to generate it from the SHACL file, but the examples turned out a bit odd.

goneall:
Note: If I add any properties to an Extension class, the resultant JSONLD file fails schema validation.

Here's a snippet that causes the failure:

```
{
    "spdxId" : "https://spdx.github.io/spdx-spec/v3.0.1/examples/full-example-eaa46bdcfa20#document",
    "type" : "SpdxDocument",
    "extension" : [ {
      "type" : "extension_Extension",
      "https://my/extension/namespace/extensionProp" : "Extension property value"
    } ],
```

@JPEWdev - does this snippet look right to you? If so, do you think the schema is fixable?

goneall:
The example should have all classes except:
- Extension - due to validation issue
- ~~ExploitCatalogVulnAssessmentRelationship - due to an inconsistency with the locator property between the Java generated code and the current JSON schema~~

goneall:
Note that this was generated with a significant amount of manual work and a lot of copy/paste - so I would be shocked if it didn't have any errors.  I only did a cursory check and ran it through schema and Java tools validation.  Please review and if any issues are identified, please create a PR or issue in the [tools-java](https://github.com/spdx/tools-java) repository which was used to generate the example.  The file was generated from [FullSpdxV3Example.java source file](https://github.com/spdx/tools-java/blob/master/examples/org/spdx/examples/FullSpdxV3Example.java)

goneall:
ExploitCatalogVulnAssessmentRelationship  is now in the example

goneall:
Validation is currently failing.   There are 3 categories of errors.  Not sure if this is an issue with the validator or the example file.

Some sample errors from each category:

### Listed License class constraint violations:

```
Violation of type sh:ClassConstraintComponent:
	Severity: sh:Violation
	Source Shape:
		@prefix sh: <http://www.w3.org/ns/shacl#> .
		@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

		[] sh:class <https://spdx.org/rdf/3.0.1/terms/Core/Element> ;
		    sh:minCount 1 ;
		    sh:nodeKind sh:IRI ;
		    sh:path <https://spdx.org/rdf/3.0.1/terms/Core/to> .
	Focus Node: <https://spdx.github.io/spdx-spec/v3.0.1/examples/full-example-eaa46bdcfa20#snippet-declared>
	Value Node: <http://spdx.org/licenses/Apache-2.0>
	Result path: <https://spdx.org/rdf/3.0.1/terms/Core/to>
	Message: Value does not have class ns1:Element
```

In this case, the "to" Element is a listed license URI where the definition is external to the JSON file.  I can change the settings to include the full license, but it will make the file much larger.

This error occurs several times.

Similar class constraints fail for class constraints `ExtendableLicense`, `LicenseAddition`

### Decimal data type validation

```
Violation of type sh:DatatypeConstraintComponent:
	Severity: sh:Violation
	Source Shape:
		@prefix sh: <http://www.w3.org/ns/shacl#> .
		@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

		[] sh:datatype xsd:decimal ;
		    sh:maxCount 1 ;
		    sh:minCount 1 ;
		    sh:nodeKind sh:Literal ;
		    sh:path <https://spdx.org/rdf/3.0.1/terms/Security/score> .
	Focus Node: <https://spdx.github.io/spdx-spec/v3.0.1/examples/full-example-eaa46bdcfa20#cvssv3vuln>
	Value Node: "5.0"^^<http://www.w3.org/2001/XMLSchema#decimal>
	Result path: <https://spdx.org/rdf/3.0.1/terms/Security/score>
	Message: Value is not Literal with datatype xsd:decimal
```

Don't quite understand this error - sure looks like `5.0` is of type `xsd:decimal` - @JPEWdev - any thoughts?

Similar issue for `energyQuantity`, `probabability`, and `percentile`

### Custom license validation failure

```
Violation of type sh:ClassConstraintComponent:
	Severity: sh:Violation
	Source Shape:
		@prefix sh: <http://www.w3.org/ns/shacl#> .
		@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

		[] sh:class <https://spdx.org/rdf/3.0.1/terms/SimpleLicensing/AnyLicenseInfo> ;
		    sh:minCount 2 ;
		    sh:nodeKind sh:IRI ;
		    sh:path <https://spdx.org/rdf/3.0.1/terms/ExpandedLicensing/member> .
	Focus Node: <https://spdx.github.io/spdx-spec/v3.0.1/examples/full-example-eaa46bdcfa20#complexlicense>
	Value Node: <https://spdx.github.io/spdx-spec/v3.0.1/examples/full-example-eaa46bdcfa20#LicenseRef-customlicense1>
	Result path: <https://spdx.org/rdf/3.0.1/terms/ExpandedLicensing/member>
	Message: Value does not have class <https://spdx.org/rdf/3.0.1/terms/SimpleLicensing/AnyLicenseInfo>
```

The value is of type `expandedlicensing_CustomLicense` which is a subclass of `AnyLicenseInfo` - so - not sure why this constraint is failing.

goneall:
> Decimal data type validation

This validation issue can be solved by enclosing the decimal values in quotes.

FYI - @JPEWdev - your suggestion worked :)

goneall:
Most validation issues have been fixed.

There is one remaining issue with a class constraint on `customlicense1`

I created a separate SPDX example file 
[expanded-license-example.json](https://github.com/user-attachments/files/23402090/expanded-license-example.json)
 with just the expanded licensing and it passed validation.

Not sure why the larger example is failing.

goneall:
Fixed validation issue - it was a duplicate IRI for customlicense1.

@JPEWdev @bact @zvr - If I could get a couple reviews, we can merge

bact:
Tested validation locally (using ajv, check-jsonschema, and spdx3-validate), looks good, all reported valid.

Minor point but wonder if we can keep the formatting the same as in
https://github.com/spdx/spdx-spec/blob/develop/examples/jsonld/package_sbom.json ?

- Line 20 is for example rather too long and it is possible to have one entry per line (like package_sbom.json does).

If it is easy to put it to some formatter (like https://jsonlint.com/ ), I think it is worth to do it for readability and consistency.

If it is too much of the hassle, the current file as it is is good for merging.

bact (approved):

goneall:
As part of the spdx-3-model CI, we could generate a new example each time based on the generated SHACL file.  This would eliminate the need to manually commit updates on every change to the model.

bact:
I like the idea, but where will we generate the example from

bact:
Ha. I got it now, the JSON can be generated from SHACL. Populate anything that will fit the restrictions. I think it's good if we want some examples for people to look at.

But if the example is to be used for validation, it will defeat the purpose. (It should always pass)

goneall:
> Ha. I got it now, the JSON can be generated from SHACL. Populate anything that will fit the restrictions. I think it's good if we want some examples for people to look at.
> 
> But if the example is to be used for validation, it will defeat the purpose. (I should always pass)

Agree - it would only be used as an example, not validation.

The goal is to publish an HTML version of the example using the Python code @JPEWdev developed and link the generated spec web pages to the examples.

bact:
Btw, we could use this for regression test if the newer version of spec (during development stage) breaks existing JSONs from older specs or not.

kestewart:
Discussed in call possible workflow,   to incorporate into CI;  and then generate HTML file. 
Decision to include it now, and remove later once the workflow is finished.

zvr (approved):

zvr:
Merging this as ready, and we can look in the future about having an automatically produced example file from the model definition.

PR #1300, opened 2025-10-31
Remove the pyshacl check for the examples

goneall:
PyShacl doesn't support external elements which will soon be added in one of the examples.

It looks like spdx3-validate should be sufficient.

Reference: https://github.com/spdx/spdx-spec/pull/1299#issuecomment-3471615629

### DIFF
--- a/.github/workflows/validate_examples.yml
+++ b/.github/workflows/validate_examples.yml
@@ -21,7 +21,7 @@ jobs:
           cache: "pip"
       - name: Install Python dependencies
         run: |
-          python3 -m pip install check-jsonschema==0.31.0 pyshacl==0.30.0 spdx3-validate==0.0.5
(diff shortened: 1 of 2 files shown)


bact (approved):

.github/workflows/validate_examples.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           cache: "pip"
       - name: Install Python dependencies
         run: |
-          python3 -m pip install check-jsonschema==0.31.0 pyshacl==0.30.0 spdx3-validate==0.0.5
+          python3 -m pip install check-jsonschema==0.31.0 spdx3-validate==0.0.5
```
bact:
```suggestion
          python3 -m pip install check-jsonschema==0.34.1 spdx3-validate==0.0.5
```

Optionally, we can also update the check-jsonschema.


Or we can do this separately.

bact (approved):

PR #1306, opened 2025-11-12
Fix links in terms and definitions

bact:
- Add links to ISO vocabularies (both are free)
- Make IETF RFC URL format consistent with other instances in the spec (for example, as in [references.md](https://github.com/spdx/spdx-spec/blob/develop/docs/references.md))
  - Use format: `https://datatracker.ietf.org/doc/rfc<number>/`, from https://github.com/spdx/spdx-3-model/blob/develop/docs/format.md#writing-style 
- Make links clickable (currently they are not, see https://spdx.github.io/spdx-spec/v3.1-dev/terms-and-definitions/)
- Remove trailing spaces


goneall (approved):
LGTM

bact:
I slightly updated the ISO standard names with an em dash (—), to make it the same as in the References https://github.com/spdx/spdx-spec/blob/develop/docs/references.md

Ping @gregshue for review

zvr (approved):

PR #1309, opened 2025-11-14
Update headers to differentiate

zvr:
This changes the same (short) header in two consecutive sections of the matching guidelines annex.

Signed-off-by: Alexios Zavras (zvr) <github@zvr.gr>


bact (approved):

goneall (approved):
LGTM

PR #1325, opened 2026-01-08
Add UUID to spdxIds in example data

### LINKED ISSUE #1323: Unrealistic spdxIds in Example
The example contains unrealistic `spdxId`s and `@id`s which are probably not globally unique. This causes a wrong expectation and might lead to wrong usage.

https://github.com/spdx/spdx-spec/blob/5c117e8b36996285ba843634f7dfcb7ebccd4810/examples/jsonld/package_sbom.json#L28

I would suggest to add UUIDs to the url.

I do not know what the best convention for the hostname of the URL are.

bact:
- <s>Use https://spdx.org/spdxdocs/ prefix, as found in many examples in https://github.com/spdx/spdx-examples/</s>
- Add UUID to the ID

Update 24 Apr 2026: Use `spdxId` as suggested in the newly updated spdxId description (see https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Properties/spdxId/ - revised by https://github.com/spdx/spdx-3-model/pull/1215)

To fix https://github.com/spdx/spdx-spec/issues/1323


maxhbr:
is the `https://spdx.org/spdxdocs/...` prefix to IDs what we expect third-parties to do? Do we have guidance on that?

bact:
> is the `https://spdx.org/spdxdocs/...` prefix to IDs what we expect third-parties to do? Do we have guidance on that?

From SPDX 2.3 spec, 
https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#652-intent :

> A best practice for creating the URI for SPDX documents available on the public internet is `https://[CreatorWebsite]/[pathToSpdx]/[DocumentName]-[UUID]` ...

> If the creator does not own their own website, a default SPDX CreatorWebsite and PathToSpdx can be used `spdx.org/spdxdocs`. Note that the SPDX documents are not currently stored or accessible on this website. The URI is only used to create a unique ID following the above conventions.

However, I don't think we have that written in 3.0 spec. The closest thing we have is probably an example in "Getting Started" guide, hosted in the `using` repo: https://spdx.github.io/using/getting-started/

I think it will benefit spec readers and implementers if we have this written somewhere in 3.0 spec. (Keith raised this spdxId schema question as well during Implementers call on 5 Jan 2026)

bact:
[Off-topic] Looking at SPDX 2.x specs, I really like the way they have the Purpose and Intent sections for every fields. They are very useful and can help spec readers, including implementers, understand the design choices. We tend to have those explanations in issues and PRs of spdx-spec and spdx-3-model repos, meeting minutes, and mailing list, but not in one single structured place like in 2.x.

zvr:
I suggest we use `https://spdx.org/example-data/...` for identifiers in our examples.

maxhbr:
But what is the guidance for a tool vendor. Should they write

* `https://spdx.org/example-data/...`
* `https://spdx.org/${tool-name}-document/...`
* `https://tool-vendor.com/spdx-document/...`
* `urn:....`

I think that us using the `example-data` here creates confusing examples which can not easily be transferred to something a tool would create. They are hand-curated and hand-build examples which makes them even more unrealistic. I saw developers struggling with the task "what to write in the spdxid field".

maxhbr:
I think a recommended "if not sure use the following way of generating spdx ids" would be great

examples/jsonld/package_sbom.json
```diff
@@ -5,14 +5,14 @@
             "type": "CreationInfo",
             "@id": "_:creationinfo",
             "createdBy": [
-                "http://spdx.example.com/Agent/JoshuaWatt"
+                "https://spdx.org/spdxdocs/package-sbom-ex-a665fd93-4eb4-4675-aa8a-6098b7ba7dc8/Agent/JoshuaWatt"
```
zvr:
```suggestion
                "https://spdx.org/example-data/package-sbom-ex-a665fd93-4eb4-4675-aa8a-6098b7ba7dc8/Agent/JoshuaWatt"
```

examples/jsonld/package_sbom.json
```diff
@@ -5,14 +5,14 @@
...
             ],
             "specVersion": "3.0.1",
             "created": "2024-03-06T00:00:00Z"
         },
         {
             "type": "Person",
-            "spdxId": "http://spdx.example.com/Agent/JoshuaWatt",
+            "spdxId": "https://spdx.org/spdxdocs/package-sbom-ex-a665fd93-4eb4-4675-aa8a-6098b7ba7dc8/Agent/JoshuaWatt",
```
zvr:
```suggestion
            "spdxId": "https://spdx.org/example-data/package-sbom-ex-a665fd93-4eb4-4675-aa8a-6098b7ba7dc8/Agent/JoshuaWatt",
```

zvr (dismissed):
... and so on.

Essentially, use `https://spdx.org/example-data/` as prefix.

zvr:
> I think a recommended "if not sure use the following way of generating spdx ids" would be great

I completely agree, but this is a separate point from the actual example data file.

kestewart:
DIscussed on call.   Agreed to move to -rc2.    Need feedback from Max.

bact:
Apparently, this issue is not new:
- https://github.com/spdx/spdx-spec/pull/964

bact:
To update changes to reflect the recently updated [`spdxId`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/spdxId.md ) (updated by PR https://github.com/spdx/spdx-3-model/pull/1215).

examples/jsonld/package_sbom.json
bact: Updated the base URL to one in https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Properties/spdxId/

examples/jsonld/package_sbom.json
bact: Updated the base URL to one in https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Properties/spdxId/

bact:
I have updated the `spdxId` in both examples to the format suggested in the newly updated [spdxId description](https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Properties/spdxId/).

Note that `#` is removed from `spdxId` of `SpdxDocument`, to follow the suggestion above.

@maxhbr @zvr @JPEWdev @goneall please review.


(if we don't want to use `https://spdx.org/spdxdocs/...`, I think we can also use the URL that based on the actual path, for example, `https://spdx.github.io/spdx-spec/v3.1/examples/jsonld/spdx-full-example.json/...` - if we decided to publish these examples on https://spdx.github.io/spdx-spec/v3.1/ -- the spdx-full-example.json already use something close to this convention)

goneall (approved):
Discussed on 21 July 2026 Tech call - we'll go ahead and approve/merge for now - if we want a shorter version, we can do a separate PR

goneall:
@zvr - any concern with just merging this PR and creating a separate PR if we want a shorter / different ID approach?

bact:
> Discussed on 21 July 2026 Tech call - we'll go ahead and approve/merge for now - if we want a shorter version, we can do a separate PR

Will merge with one approval here and as agreed in the Tech call.

PR #1247, opened 2025-07-24
[3.0] Use MAJOR.MINOR version JSON-LD context URL in example

bact:
> This PR targets `support/3.0` branch.
> A similar PR for `develop` is at #1244

To use "3.0" RDF URLs (version without patch version), as required by https://github.com/spdx/spdx-3-model/issues/1046, we need to verify if new RDF URLs are working.

We may check that with an SBOM example (`examples/jsonld/package_sbom.json`) and the example validation workflow in this repo.

This issue may not directly affect the texts in the ISO publication but without the fix the texts in ISO publication may not back with relevant working URLs.

- See https://github.com/spdx/spdx-3-model/issues/1046 (ISO issue)

PR #1331, opened 2026-01-16
[3.0] Use MAJOR.MINOR version JSON-LD context URL in example

goneall:
Replaces #1247


goneall:
This PR isn't complete, we can go with #1247 once we fix the CI failure.

PR #1345, opened 2026-01-24
Allow custom version publication for v3

bact:
This will allow user to pass parameters to the v3 publication workflow directly from the GitHub web UI.

Similar to what we already have with v2 publication workflow.


kestewart (approved):
Walked through the flow with @bact,  agree it makes sense with the mapping work he's done.   Going ahead and approving.

PR #1346, opened 2026-01-24
Remove exact VERSION from aliase

bact:


kestewart (approved):
Discussed with @bact.   Should fix last issue,  going ahead and publishing.

PR #1348, opened 2026-01-24
Exclude exact version from aliases

bact:
Updated the deduplication process for aliases to exclude the exact version from the list.


kestewart (approved):
Hopefully 3rd attempt will work now.

PR #1356, opened 2026-02-05
Terms add: stakeholder, persona

gregshue:
The terms 'stakeholder' and 'personal' have been well defined in Systems Engineering standards for quite a while. For clarity across SPDX these terms and definitions should be included in the terms_and_definitions file.


bact:
@bennetkl @rgopikrishnan91 should bring this to Raymond, Victor, and other people attention as well.

docs/terms-and-definitions.md
```diff
@@ -49,6 +49,19 @@ from CNSSI 4009-2015,
 *leveraged* from IETF RFC 4949 Ver 2 at
 <https://datatracker.ietf.org/doc/rfc4949/>)
 
+**persona**
+
+Representation of a type of user that includes 
+a concise summary of the characteristics of the user
```
nicpappler, line 55: I mainly agree, but isn't a persona more than only a user? Can a persona also be an actor/creator in another context? E.g. a persona can add requirements to a requirement document, right? Is this then simply a user of the requirements workflow (I think I could be ok with that) or if we need to add more definitions to persona than user?

nicpappler (commented):
I agree with the definition of stakeholder, but
I have questions regarding persona (see my comment there)

docs/terms-and-definitions.md
gregshue: @nicpappler This is the definition from ISO/IEC 25063:2014. The users of this standard may need to exchange 25063:2014:persona information using SPDX encodings. If SPDX uses a different definition for persona then how much meaning of the original data is lost? Can the original clarity be recovered? 

(edit) Please pay special attention to the Note 1 text below. This clarifies the persona information is fictional. ;-) A persona itself is not an actor/agent/role/stakeholder/user, so it cannot add requirements.

gregshue: @nicpappler I have added a Persona grammar, file and initial item into my SPDX-example/conformance/example1 PR. You can easily see sdoc source [here](https://github.com/gregshue/spdx-examples/blob/conformance-example1/conformance/example1/content/src/problem_space/personas/sdoc/index.sdoc).

goneall (approved):
Discussed on SPDX tech call - consensus to approve

PR #1360, opened 2026-02-22
Fix JSON-LD context URL in JSON schema to use spdx.org URL

### LINKED ISSUE #1358: Update publication workflow to use 3.1 context file URL from spdx.org
The 3.1 context file is now available: https://spdx.org/rdf/3.1/spdx-context.jsonld

Update the publication workflow to use this URL when generating JSON schema.

bact:
To fix #1358 

Currently https://spdx.github.io/spdx-spec/v3.1/rdf/schema.json contains this context document line:

```json
    "properties": {
        "@context": {
            "const": "https://spdx.github.io/spdx-spec/v3.1-dev/rdf/spdx-context.jsonld"
        }
    },
```

Since we already have https://spdx.org/rdf/3.1/spdx-context.jsonld being setup,
this PR will update the JSON schema generation script to change that line to:

```json
    "properties": {
        "@context": {
            "const": "https://spdx.org/rdf/3.1/spdx-context.jsonld"
        }
    },
```

Also remove `canonical_version` setting in MkDocs. The setting used to make sense when we had only one minor version (3.0.x), now we have two (+3.1).


goneall (approved):
LGTM

bact:
We need this PR to update tooling for 3.1, including the Python binding at https://github.com/spdx/spdx-python-model/issues/17

.github/workflows/publish_v3.yml
```diff
@@ -226,17 +226,36 @@ jobs:
...
-          if [[ "$V" == "develop" || "$V" == "dev" || "$V" =~ -dev$ || "$V" =~ -draft$ || "$V" =~ -rc[0-9]*$ ]]; then
-            CONTEXT_URL="https://spdx.github.io/spdx-spec/$VERSION/rdf/spdx-context.jsonld"
+          VER_WO_V=${VERSION#[vV]}
+          VER_CLEAN=$(echo "$VER_WO_V" | sed -E 's/-.*//')
+
+          URL_1="https://spdx.org/rdf/$VER_CLEAN/spdx-context.jsonld"
+          URL_2="https://spdx.org/rdf/$VER_WO_V/spdx-context.jsonld"
+          URL_3="https://spdx.github.io/spdx-spec/$VERSION/rdf/spdx-context.jsonld"
+
+          if curl --output /dev/null --silent --head --fail "$URL_1"; then
```
JPEWdev: Do we need this curl test? It seems like it could be flaky if there is a problem resolving the URL. Maybe we could do something based on the GH tag, or a version file or something that is determinsitic

bact: Ideally, we should use URL_1, as it is the format that we advertise officially.
But since during the development time URL_1 may not be available, the script trying to use what could be available.

But yes, if there's a case of URL_1 hiccup (the URL is already available, but inaccessible temporarily at the time of access) we may ended up in a bad situation.

We can hardcoded list of versions in the script and use that to make it more deterministic (provided that we have to update the list now and then when we're going to have a new stable/dev version, which should be fine).

Btw, if we can have a public source of truth that anyone can query all the available versions (stable and in-development), that will solve a lot of issues -- maybe including the one we discussed yesterday as well https://github.com/spdx/spdx-3-model/issues/1233 (to know "the latest stable patch version" from a string of minor version).

bact: Will implement this with git tags. Using git tags as source of truth.

bact: Btw, the URL (redirection) can be setup before the actual tag release.

Currently, https://spdx.org/rdf/3.1/spdx-context.jsonld is setup to redirect to 3.1-RC1 version of the context document. When we have 3.1-RC2, it can be updated to point to the new content -- but the URL remain unchanged (this is actually against the idea of URL, but make it more convenient to test).

With the curl approach (checking availability of context URL), we can use the same context URL in 3.1-dev, 3.1-RC1, 3.1-RC2, 3.1, etc. without actual tag release.

bact: Hybrid approach implemented. Use git tag as primary source, and network check as backup.

Will check for git tag first, if a git tag release (non-dev, non-RC) is exist, it should confirm that we have `https://spdx.org/rdf/3.x/spdx-context.jsonld` URL (if not, we should).

If the git tag release is not exist, use curl to recheck the URL. If the URL exists, use it.

This will rely less on network condition.

The check will always be deterministic for any tag release (since git tag is the first thing being checked). For development versions, it will be less deterministic (may rely on network condition).

bact:
@goneall With these items:

1) the new shacl2code 1.0.0 from @JPEWdev (https://github.com/spdx/spdx-spec/pull/1368 - merged)
2) this PR

We should have a new JSON schema and context document that support short ids.

JPEWdev (approved):

PR #1361, opened 2026-02-22
Update SPDX License List refs to v3.28.0

bact:
We may need to update this again before the 3.1 final release, if there is an update in SPDX License List.


goneall (approved):

bact:
Will merge with one approval and no objections/request for changes over 10 days waiting period. No change or no substantial change in content.

PR #1377, opened 2026-03-26
Validate all timestamps and dates in the JSON Schema.

zbleness:
The HTML spec is quite clear about the format: YYYY-MM-DDThh:mm:ssZ. Many JSON Schema descriptions suggest that full ISO8601 format is supported, which include fractional seconds and timezone offsets.

This patch aligns the HTML spec with the JSON schema.

### DIFF
--- a/schemas/spdx-schema.json
+++ b/schemas/spdx-schema.json
@@ -21,7 +21,7 @@
           "annotationDate" : {
             "description" : "Identify when the comment was made. This is to be specified according to the combined date and time in the UTC format, as specified in the ISO 8601 standard.",
             "type" : "string",
-            "pattern" : "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}Z$"
+            "pattern" : "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$"
           },
           "annotationType" : {
             "description" : "Type of the annotation.",
@@ -53,7 +53,7 @@
         "created" : {
           "description" : "Identify when the SPDX document was originally created. The date is to be specified according to combined date and time in UTC format as specified in ISO 8601 standard.",
           "type" : "string",
-          "pattern" : "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}Z$"
+          "pattern" : "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$"
         },
         "creators" : {
           "description" : "Identify who (or what, in the case of a tool) created the SPDX document. If the SPDX document was created by an individual, indicate the person's name. If the SPDX document was created on behalf of a company or organization, indicate the entity name. If the SPDX document was created using a software tool, indicate the name and version for that tool. If multiple participants or tools were involved, use multiple instances of this field. Person name or organization name may be designated as “anonymous” if appropriate.",
@@ -152,7 +152,7 @@
                 "timestamp" : {
                   "description" : "Timestamp",
                   "type" : "string",
-                  "pattern" : "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}Z$"
+                  "pattern" : "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$"
                 },
                 "url" : {
                   "description" : "URL Reference",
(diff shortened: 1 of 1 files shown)


schemas/spdx-schema.json
```diff
@@ -20,7 +20,8 @@
         "properties" : {
           "annotationDate" : {
             "description" : "Identify when the comment was made. This is to be specified according to the combined date and time in the UTC format, as specified in the ISO 8601 standard.",
```
bact, line 22: We may need to update this description in the spec itself as well (probably in another PR).

In 3.0.1, it has these requirements:

> It has resolution of seconds and is always expressed in UTC time zone.

schemas/spdx-schema.json
```diff
@@ -20,7 +20,8 @@
...
-            "type" : "string"
+            "type" : "string",
+            "pattern" : "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}Z$"
```
bact: Consider using `[0-9]` instead of `\d` as a defensive pattern to avoid over-capturing of non-Arabic numerals.

See https://github.com/spdx/spdx-3-model/pull/1213/changes

zbleness: Done

goneall (approved):
LGTM - Thanks @zbleness

schemas/spdx-schema.json
```diff
@@ -20,7 +20,8 @@
         "properties" : {
           "annotationDate" : {
             "description" : "Identify when the comment was made. This is to be specified according to the combined date and time in the UTC format, as specified in the ISO 8601 standard.",
-            "type" : "string"
+            "type" : "string",
+            "pattern" : "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$"
```
bact:
```suggestion
            "pattern" : "^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])T([0-1][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])Z$"
```

I just think about this again, and maybe we can use this opportunity to go stricter.

The suggested pattern will restrict:
- month to 01-12
- date to 01-31
- hour to 00-23
- minute to 00-59
- second to 00-59

This restrictions are aligned with ISO 8601

https://www.w3.org/TR/NOTE-datetime

zbleness: seconds would need to be 00-60 to make room for an occasional [leap second](https://www.nist.gov/pml/time-and-frequency-division/leap-seconds-faqs) (which the w3 spec does not account for), but yeah we could do that.

bact: Sounds great. Should we try this? @goneall (except leap second, the constrains are in the ISO already, not our own invention)

goneall: Sounds good to me

bact: Thank you. @zbleness would you mind to implement this in every instances please?

zbleness: sure, np.

zbleness: Done

bact (approved):
LGTM. Thank you @zbleness.

Note: we may like to similarly tighten the constraint in SPDX 3.1 [`DateTime`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Datatypes/DateTime.md) as well.

zbleness:
I would like to back-port this to a 2.2.x branch as well to align with the SPDX 2.2.2 docs:
https://spdx.github.io/spdx-spec/v2.2.2/document-creation-information/

@bact  @goneall  can you please advice what target branch I should use ?

bact:
We have https://github.com/spdx/spdx-spec/tree/development/v2.2.2. But I'm not sure about the policy if the SPDX project still continue any update on 2.2.x.

goneall:
> We have https://github.com/spdx/spdx-spec/tree/development/v2.2.2. But I'm not sure about the policy if the SPDX project still continue any update on 2.2.x.

I think the update to 2.3 should be sufficient

schemas/spdx-schema.json
bact: Btw, the leap second is only allowed at 23:59:60, per

- ISO 8601:2019 Clause 3.1.1.24 https://www.iso.org/obp/ui/#iso:std:iso:8601:-1:ed-1:v1:en ; and
- RFC 3339 Appendix D https://www.rfc-editor.org/rfc/rfc3339.html#appendix-D

So the pattern could be revised as:

`^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]|23:59:60)Z$`

zbleness: It is a good catch and a nice improvement. You want to contribute that yourself (your achievement not mine - you should get the credit)?

bact: I don't mind. It is based on your observation. I never know about this leap second arrangement in ISO format before. Please create a PR if you like.

zbleness: Done: https://github.com/spdx/spdx-spec/pull/1384

Issue #1359, opened 2026-02-22
[Tracker] IANA media type and file extension


bact:
This tracker issue is for issues related to the MIME media type registration at IANA, for both SPDX 2 and SPDX 3.

A registered media type, with IANA assignment, is required for content negotiation, and help automation in general (see https://github.com/spdx/spdx-spec/issues/1364 for discussion and status of content negotiation).

As of 21 Apr 2026, we have

these media types registered:

| Media type name | SPDX version | Format | Extension | IANA assignment |
| - | - | - | - | - |
| `text/spdx` | SPDX 2 | tag:value | .spdx | https://www.iana.org/assignments/media-types/text/spdx |
| `application/spdx+json` | SPDX 2 | JSON | .spdx.json | https://www.iana.org/assignments/media-types/application/spdx+json |

these media types under IANA review:

| Media type name | SPDX version | Format | Extension | IANA status |
| - | - | - | - | - |
| `application/spdx3+json` | SPDX 3 | JSON | .spdx3.json | https://tools.iana.org/public-view/viewticket/1450117 |

these media types not yet registered:

| Media type name (suggested) | SPDX version | Format | Extension | Notes |
| - | - | - | - | - |
| `application/spdx-rdf+xml` | SPDX 2 | RDF (RDF/XML) | .spdx.rdf | We can't use `spdx+rdf` as RDF is not one of [Structured Syntax Suffixes](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml). We can't have more than one `+`. |
| `application/spdx+xml` | SPDX 2 | XML | .spdx.xml | Inquiry in [this issue](https://github.com/spdx/spdx-spec/issues/577). This is an XML format that is not RDF. |
| `application/spdx+yaml` | SPDX 2 | YAML | .spdx.yaml or .spdx.yml | |
| `application/spdx3+xml` | SPDX 3 | RDF (RDF/XML) | .spdx3.rdf | Everything in SPDX 3 is RDF, we can omit that. The format is listed in [file naming](https://spdx.github.io/spdx-spec/v3.1-dev/serializations/#file-naming) in the 3.1-dev spec but there is no implementation yet. |
| `application/spdx3+cbor` | SPDX 3 | CBOR | .spdx3.cbor | The format is listed in [SCITT](https://datatracker.ietf.org/doc/draft-ietf-scitt-architecture/18/) draft. SPDX in CBOR is presented in 21 Apr 2026 Tech call. |

[...]

PR #1383, opened 2026-04-09
Add a paragraph about file naming

vargenau:
This solves https://github.com/spdx/spdx-spec/issues/987

### DIFF
--- a/docs/serializations.md
+++ b/docs/serializations.md
@@ -152,6 +152,7 @@ The SPDX 3 OWL ontology is available at:
 It should be easy to recognize an SPDX 3 file in a file system without opening the file.
 
 A suggested naming convention is:
+
 | Format  | Extension    |
 |---------|--------------|
 | JSON-LD | *.spdx3.json |
--- a/docs/serializations.md
+++ b/docs/serializations.md
@@ -154,6 +154,6 @@ It should be easy to recognize an SPDX 3 file in a file system without opening t
 A suggested naming convention is:
 
 | Format  | Extension    |
-|---------|--------------|
+| ------- | ------------ |
 | JSON-LD | *.spdx3.json |
 | RDF/XML | *.spdx3.rdf  |
--- a/docs/serializations.md
+++ b/docs/serializations.md
@@ -146,3 +146,13 @@ The SPDX 3 JSON Schema is available at:
 
 The SPDX 3 OWL ontology is available at:
 <https://spdx.org/rdf/3.1/spdx-model.ttl>
+
+## File naming
+
+It should be easy to recognize an SPDX 3 file in a file system without opening the file.
+
+A suggested naming convention is:
+| Format  | Extension    |
+|---------|--------------|
+| JSON-LD | *.spdx3.json |
+| RDF/XML | *.spdx3.rdf  |


docs/serializations.md
```diff
@@ -146,3 +146,13 @@ The SPDX 3 JSON Schema is available at:
 
 The SPDX 3 OWL ontology is available at:
 <https://spdx.org/rdf/3.1/spdx-model.ttl>
+
+## File naming
+
+It should be easy to recognize an SPDX 3 file in a file system without opening the file.
+
+A suggested naming convention is:
```
bact, line 154:
```suggestion
A suggested naming convention is:

```

There should be one blank line before the table.
Just to avoid MkDocs rendering issues.

docs/serializations.md
```diff
@@ -146,3 +146,13 @@ The SPDX 3 JSON Schema is available at:
...
+| Format  | Extension    |
+|---------|--------------|
```
bact:
```suggestion
| ------- | ------------ |
```

Minor Markdown format.

bact (changes requested):
Very minor Markdown formatting suggestions.

The content is good. Thank you for putting this together.

bact (approved):
LGTM. Aligned with general agreement from 31 March 2026 Tech Call, see notes https://github.com/spdx/spdx-spec/issues/1359#issuecomment-4163950985

If OMG sees it other way, we can of course update this later.

zvr:
A specification, which becomes an ISO standard is not a place to have content about wishes ("should be easy") or suggestions. This is part of a Normative chapter; what does this text mean?

Is it _required_ to use these file extensions in order to conform with the specification? If not, these do not belong here.

goneall:
> A specification, which becomes an ISO standard is not a place to have content about wishes ("should be easy") or suggestions. This is part of a Normative chapter; what does this text mean?
> 
> Is it _required_ to use these file extensions in order to conform with the specification? If not, these do not belong here.

What if we change the context of the statement to reference the IANA media type once it is approved?

Something like "SPDX JSON-LD is registered as an IANA media type XXX which is associated with the file extension .spdx3.json."

vargenau:
> A specification, which becomes an ISO standard is not a place to have content about wishes ("should be easy") or suggestions. This is part of a Normative chapter; what does this text mean?
> 
> Is it _required_ to use these file extensions in order to conform with the specification? If not, these do not belong here.

Hi Alexios,

I am sorry, but I have to disagree.

In the current SPDX ISO standard, the wording is as follows:

```
Should be easy to recognize in a file system without opening the file. A suggested naming
convention is:
Table 1 — Suggested naming convention
Format Extension
tag:value *.spdx
RDF *.spdx.rdf
JSON *.spdx.json
XML *.spdx.xml
YAML *.spdx.yaml or *.spdx.yml
```

So yes the ISO standard is "suggesting".

In my proposal, I tried to be as close as possible to the SPDX 2 ISO standard.

bact:
Checked with ISO/IEC 5962:2021 (SPDX v2.2.1) https://www.iso.org/standard/81870.html (free to download) page 5, line 2, same wording there ("Should be easy to ...").

The text is inside "4.4 Standard data format requirements" clause.

vargenau:
@bact @goneall 
Hi Arthit and Gary,
As we approved this in the tech call, could you please merge?
Thank you

goneall:
@vargenau - looks like the branch is out of date - can you rebase with the latest and push the update?

Normally, I could do this from the UI - but for some reason GitHub isn't giving me the option for this PR.

goneall (approved):

vargenau:
> @vargenau - looks like the branch is out of date - can you rebase with the latest and push the update?
> 
> Normally, I could do this from the UI - but for some reason GitHub isn't giving me the option for this PR.

@goneall 
I have rebased

PR #1389, opened 2026-04-24
Restore schemas directory for IANA compatibility

bact:
The `schemas` directory is referenced in official specifications, including IANA media type registration.
It is also referenced in `examples/README.md` (alongside the `ontology` directory).

To fix https://github.com/spdx/spdx-spec/issues/1158 (reported by @RingoDev), this PR:

- Restore `schemas` directory: Maintains directory structure as documented by IANA ([application/spdx+json](https://www.iana.org/assignments/media-types/application/spdx+json)) and the SPDX v2.3 Specification ([4.4 Standard data format requirements ](https://spdx.github.io/spdx-spec/v2.3/conformance/#44-standard-data-format-requirements)).
- Keep `spdx-schema.json` pointing to the latest SPDX 2.x version, for backward compatibility. (SPDX 3 versions can access the schema at https://spdx.github.io/spdx-spec/VERSION/rdf/schema.json).
- Rename `rdf` directory back to `ontology` directory.
  - Note: This will not break SPDX 3 publication workflow, as the RDF directory for web publication is `docs/rdf`, not the root `rdf`.

**Notes:** since IANA points to https://github.com/spdx/spdx-spec/blob/master/schemas/spdx-schema.json -- the change in this PR (`develop` branch) has to be mirrored in `main` branch too. GitHub will automatically redirected `master` branch URLs to `main`.

| File | Source |
| - | - |
| spdx-schema-2-2.json | https://github.com/spdx/spdx-spec/blob/development/v2.2/schemas/spdx-schema.json |
| spdx-schema-2-2-1.json | https://github.com/spdx/spdx-spec/blob/development/v2.2.1/schemas/spdx-schema.json |
| spdx-schema-2-2-2.json | https://github.com/spdx/spdx-spec/blob/development/v2.2.2/schemas/spdx-schema.json |
| spdx-schema-2-3.json | https://github.com/spdx/spdx-spec/blob/support/2.3/schemas/spdx-schema.json |
| spdx-schema-2-3-1-dev.json | https://github.com/spdx/spdx-spec/blob/support/2.3.1/schemas/spdx-schema.json |
| spdx-schema-3-0-1.json | https://spdx.github.io/spdx-spec/v3.0.1/rdf/schema.json |
| spdx-schema-3-1-dev.json | https://spdx.github.io/spdx-spec/v3.1-dev/rdf/schema.json |
| spdx-schema.json | Latest 2.x (symlink) |

PR #1396, opened 2026-06-06
Restore "schemas" dir for IANA reference

bact:
>[!NOTE]
> This PR is similar to #1389 but for `main` branch and does not contain schemas from "dev" versions

- https://www.iana.org/assignments/media-types/application/spdx+json make a reference to https://github.com/spdx/spdx-spec/blob/master/schemas/spdx-schema.json which is currently 404.
- The IANA record points to "master" which is redirected to "main" branch.
  So there a need to have the directory in "main" as well.


bact:
Close. Will update with new structure.

PR #1394, opened 2026-05-27
Update Package-URL spelling as ECMA

vargenau:
Package-URL is now an ECMA standard.

https://ecma-international.org/publications-and-standards/standards/ecma-427/

Let us use the standard spelling with an hyphen ("-").

PR #1397, opened 2026-06-10
Use ECMA-427 for Package-URL

### LINKED ISSUE #1144: Package URL to be standardised as ECMA standards - should we converge?
Since SPDX 3.0.1 (to be published in a near future) make use of the Package URL (using [our own copy](https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/annexes/pkg-url-specification.md) of the spec, because the need to "freeze" the version for ISO submission purpose), we may have to aware of the standardisation and maybe trying not to diverge the way forward.

https://ecma-international.org/task-groups/tc54-tg2/

> Scope:
>
> Ecma TC54-TG2 is chartered with the standardization of Package URL (PURL). The scope of this task group will include:
>
> - PURL specification: development and maintenance of the PURL standard, which provides a clear and concise method for identifying packages across various package management systems.
> - VERS specification: establishment and standardization of VERS, a specification designed to consistently represent version ranges, providing a uniform approach to managing version constraints across different programming environments and package managers.
> - PURL types: standardization of PURL types, which categorize packages by their respective package management systems, ensuring that PURL remains flexible and comprehensive in covering various types of packages.
>
> Community interaction and review: Engage with the software development community to gather feedback, review new PURL type requests, and ensure the PURL specifications remain relevant and effective in addressing the needs of package management systems and software developers.

bact:
Use ECMA version of Package-URL spec instead of SPDX's own annex.

## Background

It was necessary to include the Package-URL as an annex in SPDX 3.0 as at the time of 3.0 release (Apr 2024), no official or stable Package-URL spec release is available.

In December 2025, the Package-URL becomes ECMA standard. SPDX can make reference directly to the ECMA standard.

## What this PR does?

This PR removes the Package-URL annex content and remove it from the annex list.

Note that Package-URL is also under process of ISO standardization. From https://github.com/spdx/spdx-spec/issues/1144#issuecomment-4664282228, it will take approx 11 months (May 2027) before the Package-URL will become ISO standard. Based on SPDX 3.1 timeline that targets the release at the end of 2026, we may like to reference to the ECMA version instead of waiting for the ISO one.

To resolve #1144

> [!NOTE]
> This PR is removing the Package URL Annex.
> The PR in spdx-3-model repo at https://github.com/spdx/spdx-3-model/pull/1287 is removing references to the Annex.

> [!NOTE]
> This PR will make #1394 unnecessary.


zvr:
Thanks for this, @bact.

You should also add the ECMA spec to our https://github.com/spdx/spdx-spec/blob/develop/docs/references.md list.

bact:
> You should also add the ECMA spec to our https://github.com/spdx/spdx-spec/blob/develop/docs/references.md list.

Added.

goneall (approved):
LGTM

zvr (approved):
We're losing something, since ECMA-427 is even more general than what we had, but this is the correct way to go...

bact:
> We're losing something, since ECMA-427 is even more general than what we had, but this is the correct way to go...

If there is something we think it's worth to have it somewhere in the spec, these two locations in the spdx-3-model repo could be a place we can add something:

- model/Core/Vocabularies/ExternalIdentifierType.md
- model/Software/Properties/packageUrl.md

see amendments here https://github.com/spdx/spdx-3-model/pull/1287

PR #1402, opened 2026-06-22
Remove unused symbol

zvr:
**3T-SBOM** is not used anywhere in the specification, so it should not be in the table of symbols.

It *is* used in the explanatory text for the OMG submission, so it should appear there.


bact (approved):

bact:
Will merge this minor fix with one approval, after two weeks of no objection/request for changes.

PR #1403, opened 2026-06-22
Drop minor version in base URI in serialization annotations

bact:
Follow up of
- https://github.com/spdx/spdx-3-model/pull/1277 and
- https://github.com/spdx/spec-parser/pull/212


bact:
@JPEWdev @stevenc-stb  - please help review. Thank you.

JPEWdev (approved):

goneall (approved):
LGTM - note the `@base` doesn't resolve to any documentation at this point, but that is not required (per discussion on the 23 June 2026 tech call)

PR #1406, opened 2026-06-23
[3.0] Add JSON-LD serialization annotations link

bact:
As agreed in 23 June 2026 Tech call,
add https://spdx.org/rdf/3.0/spdx-json-serialize-annotations.ttl to "Serialization in SPDX 3 JSON" section.

> [!Important]
> The URL https://spdx.org/rdf/3.0/spdx-json-serialize-annotations.ttl is currently redirected to https://spdx.github.io/spdx-spec/v3.0.1/rdf/jsonld-annotations.ttl -- which contains `@base <https://spdx.org/rdf/3.0.1/terms/> .` -- need to update that.
>
> The ISO version uses `https://spdx.org/rdf/3.0/terms/` as base IRI.
> The most recent dev version uses `https://spdx.org/rdf/3/terms/` as base IRI.

### DIFF
--- a/docs/serializations.md
+++ b/docs/serializations.md
@@ -99,6 +99,10 @@ It requires data to be serialized according to the defined serialization
 specification and validated against the SPDX 3 JSON Schema.
 It may be parsed – not serialized – using standard JSON-LD libraries.
 
+To assist implementers in developing software bindings and serialization
+tooling, a JSON-LD serialization annotations file is provided at:
+<https://spdx.org/rdf/3.0/spdx-json-serialize-annotations.ttl>
+
 ### JSON-LD context file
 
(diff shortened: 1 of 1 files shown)


zvr:
I'm not really in favor of adding a new section to the ISO document that has already been reviewed and approved...
It will also change the Table of Contents and numbering of existing sections.

My proposal would be to not do it.

goneall:
Discussed on the 21 July 2026 tech call - we could change the 3.0 to a shorter mention rather than a new section.

@bact will update this PR to move it back to a simple 1 line

goneall (approved):
Consistent with discussion on the 22 July 2026 tech call - just adding a sentence rather than a new section

bact:
Will merged this after the fix, following 21 July 2026 Tech call, with one approval and 10+ days period

PR #1409, opened 2026-06-26
Add term: interface

gregshue:
Add definition for term 'interface'. This definition is proposed by the spdx-hardware special interest group on 6/26/2026. 

Note: https://csrc.nist.gov/glossary/term/interface leverages their definition from [ISO/IEC/IEEE 15288:2015](https://www.iso.org/standard/63711.html), which has been superseded by the standards referenced in this PR.

### DIFF
--- a/docs/terms-and-definitions.md
+++ b/docs/terms-and-definitions.md
@@ -49,6 +49,13 @@ from CNSSI 4009-2015,
 *leveraged* from IETF RFC 4949 Ver 2 at
 <https://datatracker.ietf.org/doc/rfc4949/>)
 
+**interface**
+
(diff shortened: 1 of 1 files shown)


stevenc-stb (approved):

bobmartin3000 (approved):
LGTM

bact (approved):
LGTM.

We may have to later revise the whole chapter to meet ISO terminology style.

In ISO terminology management, a definition will have a "replaceability", meaning that if we see a term in sentence, we can replace the term with the definition and the sentence should still readable.

This means the first word in a definition will use lowercase (unless it is a proper noun) and the definition will not end with fullstop. -- we will do this later in later PRs.

PR #1411, opened 2026-06-30
Better table render when window is narrow

bact:
Use CSS `overflow-wrap: break-word` for text in table, to help the content in the table flow and fit the window size.

Current display of https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Classes/Action/ in a narrow window, the `maxCount` is hidden:

With changes in this PR, all columns are in the view (not very nice but at least they are there):

When window is wider, text will not breaks.

Test the dynamic flow at https://bact.github.io/spdx-spec/v3.1-dev/model/Core/Classes/Action/


bact:
Note: a JavaScript solution to insert `<wbr>` between long word like `additionalInformation` or `geographicPointLocation` has been tried, but it impacts browser performance on some older machines. Current proposal rely on browser's native CSS support. Zero JavaScript or additional dependency.

goneall (approved):

bact:
To merge a non-content improvement (CSS web stylesheet) with one approval after 3 weeks.

PR #1412, opened 2026-07-01
README: Add model URLs

bact:
Put all the canonical URLs in one place to assist developers and to promote the use of spdx.org canonical URLs of the model artifacts (discourage the use of github.io URLs).

As agreed on 1 July 2026 Implementers call.

> [!Important]
> <s>The URL https://spdx.org/rdf/3.0/spdx-json-serialize-annotations.ttl is currently 404. A redirection setup is needed.</s>
>
> Per the major.minor policy, we use "3.0" canonical URLs of model artifacts, not the "3.0.1". These are already updated in the 3.0 ISO version - see https://github.com/spdx/spdx-spec/blob/support/3.0/docs/serializations.md
> The term IRIs remain "3" (major only).


bact:
@goneall you can review this. But I will put this to draft first because we still need the redirection.

I will create another PR to update the filenames on github.io side to match filenames on spdx.org, as agreed on the 1 July 2026 Implementers call.

goneall:
I just setup a redirect of the entire spdx.org/rdf/3.0 folder to spdx.org/rdf/3.0.1 - so the link should now work and it should be current.

goneall (approved):
Changes look good and the URLs should now work.

bact:
@goneall that's quick! 

Mind that https://spdx.org/schema/3.0/spdx-json-schema.json is under `schema` dir. (but it works)

bact:
Will add URLs to development artifacts in another PR, using the matching filenames from this PR https://github.com/spdx/spdx-spec/pull/1414

bact:
Will go ahead with one approval after 3 weeks. This is a README and not part of actual spec. The changes are also agreed during implementers call.

Issue #1420, opened 2026-07-14
[3.1-RC1] Editorial cleanup: license-expression and PURL annexes (7.2)


henrylyons:
[...]

PR #1442, opened 2026-07-21
Fix typos in SPDX License Expression annex (ABNF grammar)

bact:
Fix first three issues in https://github.com/spdx/spdx-spec/issues/1420


zvr (approved):
~These should also be done in 3.0~

Forget it, we don't have the complete grammar in 3.0

goneall (approved):

Issue #1419, opened 2026-07-14
[3.1-RC1] Editorial cleanup: documentation and annexes (7.1)


henrylyons:
[...]

PR #1446, opened 2026-07-21
Model annex: typo - remove extra "in"

bact:
Fix typo as reported in https://github.com/spdx/spdx-spec/issues/1419


zvr (approved):

zvr:
This should be also done in 3.0

goneall (approved):

bact:
> This should be also done in 3.0

Fix for 3.0 is included in https://github.com/spdx/spdx-spec/pull/1451

PR #1448, opened 2026-07-21
Revise Strings in Canonical serialization

bact:
Revise for clarity to address "Says 'two characters' but enumerates three" / "Correct to 'the following characters'" point in https://github.com/spdx/spdx-spec/issues/1419


zvr:
No, that's not correct.
We want to have:
- `"` and `\` have to be "escaped" (preceded by `\`); and
- 0x00 to 0x1F can never appear.

bact:
> No, that's not correct. We want to have:
> 
>     * `"` and `\` have to be "escaped" (preceded by `\`); and
> 
>     * 0x00 to 0x1F can never appear.

We can remove 0x00 to 0x1F, if we consider that as a correction of 3.0.

Current text in `support/3.0` branch (almost identical to version on https://spdx.github.io/spdx-spec/v3.0.1/serializations/, except the "MUST"/"shall" verb):

> - Strings: UTF-8 representation without specific normalization.
  A string begins and ends with quotation marks (%x22).
  Any Unicode characters may be placed within the quotation marks,
  except for the two characters that shall be escaped by a reverse solidus:
  quotation mark, reverse solidus, and the control characters (U+0000 through U+001F).

If we are going to change that, there's another code point that we may like to change too, which is the "DEL" character (0x7F). Currently it is part of object's name:

> - Objects: An object structure is represented as a pair of curly brackets
  surrounding zero or more name/value pairs (or members).
  A name is a string containing only ASCII characters (0x21-0x7F). ...

Should be 0x21-0x7E ?

bact:
Close this issue. Use https://github.com/spdx/spdx-spec/pull/1488 instead.

PR #1449, opened 2026-07-21
Fix POSIX ERE in SPDX License List matching guidelines

bact:
To address two points in https://github.com/spdx/spdx-spec/issues/1419

- "'and is composed of 4 fields' - only 2 required"
- "Example (?i:copyright.{0,200}). is not POSIX ERE"

The **original** and **match** rule fields are only required for a rule type: `<<var; . . . >>`.

`(?i)` is not a POSIX ERE expression, use `[Cc]` instead.
Note that the new `match="[Cc]opyright.{0,200}."` is not equivalent to `match="(?i:copyright.{0,200})."`, but sufficient enough to demonstrate the point that the `match` field can use a regular expression.


goneall (approved):

goneall (approved):

bact:
Will merge this editorial fix with one approval after 10 days period.

PR #1453, opened 2026-07-22
Serialization/Parsing: Rewrite to remove "you"

bact:
Rewrite sections that migrated from notes in spdx-3-model repo.
Use ISO verbs and remove personal "you".

For example,
"You can ignore the string." -> "The string may be ignored."

(Use "may" for permission; "can" is for capability/possibility)

### DIFF
--- a/docs/serializations.md
+++ b/docs/serializations.md
@@ -147,36 +147,38 @@ It may be parsed – not serialized – using standard JSON-LD libraries.
 ### SpdxDocument
 
 The following SpdxDocument properties are mapped to native JSON-LD mechanisms
-defined within the JSON-LD syntax specifications. Any properties not listed
-below should be serialized as part of the SpdxDocument element itself within
-the JSON-LD serialized data.
+in accordance with JSON-LD syntax specifications. Any property not explicitly
+listed below shall be serialized directly within the SpdxDocument element of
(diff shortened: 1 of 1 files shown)


docs/serializations.md
```diff
@@ -253,44 +255,54 @@ Informational JSON-LD serialization examples can be found at:
...
+
+At the top level, the JSON-LD shall contain two keys: `@context` and `@graph`.
 
 #### Parsing "@context"
 
-The context is a list of a string and an object. You can ignore the string.
-The object consists of key-value pairs that allow the shortening of IDs and which we will call "namespace map" in the following.
+The `@context` key contains a list comprising a string and an object.
+The string may be ignored.
+The object contains key-value pairs used for identifier abbreviation,
```
goneall:
```suggestion
The object contains key-value pairs used for substituting identifiers with shorter abbreviations,
```

Not sure of the above suggestion is much better, but I'm think we want to add a bit more context on how the namespace map is used.

goneall (changes requested):
Just one minor suggestion to provide more context on the purpose of the namespace map

goneall (approved):
LGTM

bact:
Grammar fix. Will merge with one approval and 10+ days passes

Issue #1435, opened 2026-07-14
[3.1-RC1] Single, authoritative ABNF for license expressions


henrylyons:
**Affected Sections**: annexes/spdx-license-expressions.md<br><br>**Feedback**<br>The SPDX license-expression ABNF is defined twice and the two definitions are inconsistent (e.g., `idstring` definition, whether `NONE WITH ...` is permitted).<br><br>**Suggested Change:**<br>Provide a single, authoritative ABNF grammar definition. (annexes/spdx-license-expressions.md)<br><br>**Notes:**<br>This feedback was provided by Microsoft at the SPDX working group meeting on June 30, 2026.

bact:
Annex - should move to spdx-spec repo

zvr:
The general syntax (the first one) was kept to give a general overview of the expressions. It's not complete, and has to be clarified by wording in the text below — and still we do not explain things like when whitespace is allowed.

The second one is complete, but it was deemed too complicated to be the only one, as it's oriented towards parsing expressions.

bernhardreiter:
The first one in https://github.com/spdx/spdx-spec/blob/8bdd7c4d46695db39d6e6e0324bfe92d4fc5f3be/docs/annexes/spdx-license-expressions.md is also not in ABNF form, it violates RFC 5234.

The second ABNF can be simplified.

bernhardreiter:
#1465 makes the full ABNF shorter, from 48 to 38 lines.

This is not that much longer than the 25 lines of the first ABNF, so I suggest to delete the first oen.


And it looks to me that the can be shortened further.

PR #1465, opened 2026-08-12
simplify full ABNF in spdx-license-expressions.md

bernhardreiter:
* remove re-definitions of RFC5234 Core Rules
 * replace some hex definitions with literal strings that shorten the file and are easier to read.
 * dereference `or-later-expression` as it was only used once

improves https://github.com/spdx/spdx-spec/issues/1435

### DIFF
--- a/docs/annexes/spdx-license-expressions.md
+++ b/docs/annexes/spdx-license-expressions.md
@@ -206,11 +206,9 @@ with-expression = identifier required-ws "WITH" required-ws addition-identifier
 
 addition-identifier = license-exception-id / addition-ref
 
-identifier = license-id / or-later-expression / license-ref
+identifier = license-id / license-id "+" / license-ref
 
-or-later-expression = license-id PLUS
-
-parenthesized-expression = LPAREN optional-ws SPDX-license-expression optional-ws RPAREN
+parenthesized-expression = "(" optional-ws SPDX-license-expression optional-ws ")"
 
 special-identifier = "NONE" / "NOASSERTION"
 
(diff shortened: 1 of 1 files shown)


bact:
@bernhardreiter do you mind to update the branch with recent changes in the base branch please. Thank you.

docs/annexes/spdx-license-expressions.md
```diff
@@ -224,22 +222,14 @@ license-exception-id = <short form license exception identifier from SPDX Licens
...
 alnum    = ALPHA / DIGIT
 
 ; --- Whitespace and characters ---
 
-optional-ws = *SPACE       ; Optional whitespace (zero or more spaces)
-required-ws = 1*SPACE      ; Required whitespace (one or more spaces)
-
-SPACE  = %x20              ; Space character
-LPAREN = %x28              ; ( - Left parenthesis
-RPAREN = %x29              ; ) - Right parenthesis
```
bact, line 238: I think there is a merit to explicitly spelled out the exact code points here, as there are various parenthesis-lookalike characters in Unicode.

At the same time, the current grammar is not that strict with the `DASH` in `DocumentRef-`, `LicenseRef-`, `AdditionRef-`.

So this is totally depends on how far we like to go.

bernhardreiter: > I think there is a merit to explicitly spelled out the exact code points here, as there are various parenthesis-lookalike characters in Unicode.

ABNF is US-ASCII only, so I think the merit is outweighted by the higher readability. We could place `%x28` in the place above as well, but it would be less readable

bernhardreiter:
> do you mind to update the branch with recent changes in the base branch please.

Done

docs/annexes/spdx-license-expressions.md
bact: Agreed. Your compact version has lower cognitive complexity and may provide better readability.

We may discuss this more in the Tech call for final decision.

There is also a pending question on `NONE` and `NOASSERTION` that need @pmonks

pmonks: @bact would you mind pointing me to where that question is?  I seem to have missed it.

bact: @pmonks  I mean in general, also from this week Tech call.
I think you have mentioned several times about whether `NONE`/`NOASSERTION` should be a token in grammar or a special kind of "License ID".

pmonks: @bact ah ok. Yeah that conversation happened in this PR: https://github.com/spdx/spdx-spec/pull/1262

bernhardreiter: > There is also a pending question on NONE and NOASSERTION

Which I see as unrelated to this merge request. This merge request keeps the syntax the same.

pmonks: @bernhardreiter the addition of `NONE` and `NOASSERTION` as shown here hasn't been published yet (and yes I realise you didn't modify that part, but it's possible they will be removed before the next version is published, which may make merging this PR more complicated - I assume that's why @bact brought it up)

bact (approved):

goneall (approved):
LGTM

goneall:
@zvr - I don't know if you had a chance to review, but I'll go ahead and merge since it has 2 reviews.  If you disagree, we can create another PR to fix.

zvr:
Well, @goneall I disagreed about all of these, and I was expecting to discuss them in a call.

- the `or-later-expression` was to help understanding. Almost any grammar-based code generator would not produce more complicated code if you have a single production used once.
- the characters constants were useful to be in the grammar and not have people refer to an external RFC. In that RFC, the definitions are "provided as a convenience" and are not part of the ABNF definition.

But of course both forms (previous and current) are correct, so it makes no sense to change things back and forth.

goneall:
@zvr - I'll create a PR to reverse the changes so we can discuss

BTW - I did the merge just to avoid having constant rebases and I didn't see any comments suggesting there were any issues

bernhardreiter:
> I disagreed about all of these, and I was expecting to discuss them in a call.

Note that the length of the second ABNF was cited as cause why the first one (which isn't in ABNF form) was kept.

And an indirection and second definition means more parsing effort for people knowing the ABN standard.

PR #1468, opened 2026-08-19
Add sections for signing and attestation

JPEWdev:
Adds sections to the SPDX 3 spec that describe how documents can be signed or used for attestation

### DIFF
--- a/docs/attestation.md
+++ b/docs/attestation.md
@@ -1,0 +1,8 @@
+# Attestation with SPDX 3 JSON documents
+
+SPDX 3 JSON documents can be used as an [in-toto][in-toto] attestation
+predicate (version 1.2.0 or later) by using the predicate type
+`https://spdx.dev/Document/v3`. The predicate data is the JSON content of the
+SPDX document.
+
+[in-toto]: https://in-toto.io/
--- a/docs/signing.md
+++ b/docs/signing.md
@@ -1,0 +1,12 @@
+# Signing SPDX 3 JSON documents
+
+SPDX 3 JSON documents may be signed using [ITU-T X.590 JSON Signature Schema (JSS)][JSS].
+The signature shall be placed in a `signatures` property at the top level of
+the document. For the structure of the signature, see the [JSS][JSS]
+specification.
+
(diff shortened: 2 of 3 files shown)


docs/attestation.md
```diff
@@ -0,0 +1,8 @@
+# Attestation with SPDX 3 JSON Documents
```
bact:
```suggestion
# Attestation with SPDX 3 JSON documents
```

Not sure if "Document" with uppercased D has a special meaning here.
If not, uses "document".
(I guess not, because in line 6 it uses just "document").

Then follows ISO guidelines on using sentence casing for headings.

docs/attestation.md
```diff
@@ -0,0 +1,8 @@
...
+
+SPDX JSON Documents can be used as an [in-toto][in-toto] attestation predicate
```
bact:
```suggestion
SPDX 3 JSON documents can be used as an [in-toto][in-toto] attestation predicate
```

docs/signing.md
```diff
@@ -0,0 +1,11 @@
+# Signing SPDX 3 JSON Documents
```
bact:
```suggestion
# Signing SPDX 3 JSON documents
```

docs/signing.md
```diff
@@ -0,0 +1,11 @@
...
+
+SPDX JSON Documents may be signed using [ITU-T X.590 JSON Signature Schema (JSS)][JSS].
```
bact:
```suggestion
SPDX 3 JSON documents may be signed using [ITU-T X.590 JSON Signature Schema (JSS)][JSS].
```

Separately, for "may":
In ISO verbs, "may" means permission.
Is that what we like to say here?

docs/signing.md
```diff
@@ -0,0 +1,11 @@
...
+The signature must be placed in a `signatures` property at the top level of the
```
bact:
```suggestion
The signature shall be placed in a `signatures` property at the top level of the
```

must -> shall

bact (approved):

goneall (approved):

PR #1473, opened 2026-09-08
Add cisa minimum elements annex

vargenau:


bact:
may need to apply the CISA annex changes to `support/2.3.1` branch instead

vargenau:
> may need to apply the CISA annex changes to `support/2.3.1` branch instead

Hi @bact 
You are right.
Please use https://github.com/spdx/spdx-spec/pull/1474 instead.
I will close this one.

